PR #4, opened 2018-09-18
Support for RAK833 SPI LoRa Gateway Concentrator

kaizimmerm:
Added support for rak833 based concentrators. The chip needs apparently one more reset call which is now added to the reset script. Other chipset I tested with doe not seem to mind.

In addition:
* I switched to smaler size docker images and 
* Made the reset pin configurable by env


LoRaEngine/deployment.template.json
```diff
@@ -65,6 +65,11 @@
               "image": "${MODULES.LoRaWanPktFwdModule.arm32v7}",
               "createOptions": "  {\r\n  \"HostConfig\": {\r\n    \"NetworkMode\": \"host\", \"Privileged\": true \r\n\r\n  },\r\n  \"NetworkingConfig\": {\r\n    \"EndpointsConfig\": {\r\n      \"host\": {}\r\n    }\r\n  }\r\n}\r\n  \r\n  \r\n  \r\n"
             },
+            "env": {
+              "IOT_SK_SX1301_RESET_PIN": {
+                "value": "5"
```
Mandur: Should we give such a specific name tag or rather a more generic one to be used across device ? (like RESET_PIN)

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile
```diff
@@ -6,16 +6,15 @@ RUN apt-get install -y --no-install-recommends apt-utils build-essential
 RUN git clone https://github.com/Lora-net/packet_forwarder.git
 RUN git clone https://github.com/Lora-net/lora_gateway.git
 RUN sed -i "s|/dev/spidev0.0|/dev/spidev2.0|g" ./lora_gateway/libloragw/src/loragw_spi.native.c
-RUN sed -i "s|IOT_SK_SX1301_RESET_PIN=7|IOT_SK_SX1301_RESET_PIN=5|g" ./lora_gateway/reset_lgw.sh
 WORKDIR /packet_forwarder
 RUN ./compile.sh
 WORKDIR /build
 
-FROM microsoft/dotnet:2.0-runtime AS exec
+FROM debian:stretch-slim AS exec
```
Mandur: Not sure about this but Based on https://hub.docker.com/r/microsoft/dotnet/~/dockerfile/ it looks like the dotnet is currently based on Jessie. If we want to do image sharing we might change to Jessie rather than Stretch? (TBD)

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile.arm32v7
```diff
@@ -3,16 +3,18 @@ FROM sdthirlwall/raspberry-pi-cross-compiler as builder
...
 
-FROM microsoft/dotnet:2.0-runtime-stretch-arm32v7 as exec
+FROM arm32v7/debian:stretch-slim as exec
 WORKDIR /LoRa
 COPY --from=builder /build/packet_forwarder/lora_pkt_fwd/lora_pkt_fwd .
 COPY --from=builder /build/packet_forwarder/lora_pkt_fwd/global_conf.json .
 COPY --from=builder /build/lora_gateway/reset_lgw.sh .
 COPY local_conf.json .
 COPY start_pktfwd.sh .
-ENTRYPOINT ["./start_pktfwd.sh","7"]
```
Mandur: Is 7 the default value here?

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile
ronniesa: I agree either we stay with the dotnet one or better we test and use the Jessie that dotnet uses. Saving space and download time.

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile.arm32v7
ronniesa: there is no more default value we need to pass it as env variable and we need to update the arm template to reflect this change too. The intel and RAK833 uses 5, seeedstudio 7, iC880A 13 (still need to be tested)

LoRaEngine/deployment.template.json
ronniesa: RESET_PIN sound more easy to understand

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile
```diff
@@ -6,16 +6,15 @@ RUN apt-get install -y --no-install-recommends apt-utils build-essential
 RUN git clone https://github.com/Lora-net/packet_forwarder.git
 RUN git clone https://github.com/Lora-net/lora_gateway.git
 RUN sed -i "s|/dev/spidev0.0|/dev/spidev2.0|g" ./lora_gateway/libloragw/src/loragw_spi.native.c
-RUN sed -i "s|IOT_SK_SX1301_RESET_PIN=7|IOT_SK_SX1301_RESET_PIN=5|g" ./lora_gateway/reset_lgw.sh
```
ronniesa: If we change this dockerfile that is used by the intel gateway we need to change the corrisponding deployment.template.amd64.json to have the env variable I would also need to test it

LoRaEngine/modules/LoRaWanPktFwdModule/module.json
```diff
@@ -4,7 +4,7 @@
   "image": {
     "repository": "$CONTAINER_REGISTRY_ADDRESS/lorawanpktfwdmodule",
     "tag": {
-      "version": "0.1.0-preview",
+      "version": "0.1.1-preview",
```
Mandur: Should we switch "preview" to "dev" like 0.2.1-dev to harmonize with GitHub and avoid errors in case we patch preview branch containers.

LoRaEngine/deployment.template.json
kaizimmerm: That is why I kept the name. At the end the script is so specific. But I am open to rename it.

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile
kaizimmerm: Thats why I switched away from .NET. Stretch is the one that is below the 2.0 .NET image we used. I can upgrade if you want, no problem.

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile
kaizimmerm: Done.

LoRaEngine/modules/LoRaWanPktFwdModule/module.json
kaizimmerm: You are right, I changed it to 0.2.0-preview

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile.arm32v7
kaizimmerm: My RAK833 actually used 12. I change the defaults in the template to 7 for ARM (seedstudio) and 5 for X86_64 .

LoRaEngine/modules/LoRaWanPktFwdModule/Dockerfile
kaizimmerm: I think its good to have the same based linux image for both modules so I would stay with debian:stretch-slim here.

PR #40, opened 2018-11-23
 Add typed configuration to network server

fbeltrao:
Since I introduced a few changes to the core logic, we need to discuss them. It is a small step into decoupling the system.

Changes:
* Passing configuration as constructor to IoTHubConnector, MessageProcessor and UdpServer
* MessageProcessor.ProcessMessageAsync now returns a byte[] which is forwarded to pkt forwarder by the UdpServer. Previously MessageProcessor was calling UdpServer.UdpSendMessage directly. This changes decoupled message processing from the udp server. It won't work if processing message sends more than one message back to udp
* LoraDeviceInfoManager is now passed to dependent classes in the constructor (UdpServer, MessageProcessor). It was previously available as static methods which are hard to mock in unit tests.
* Environment variables resolution is case insensitive: LOG_TO_HUB == log_to_hub
* Debugging integration tests in Visual Studio now prints logs to debug window

Integration test succeeded with exception of C2D and device twin verification (EdgeHub having a bad day)

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoraDeviceInfoManager.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoraDeviceInfoManager.cs
@@ -18,14 +18,16 @@ namespace LoRaWan.NetworkServer
 
     public class LoraDeviceInfoManager
     {
-        public static string FacadeServerUrl;
-        public static string FacadeAuthCode;
-        public LoraDeviceInfoManager()
-        {
+        public string FacadeServerUrl;
+        public string FacadeAuthCode;
+        private readonly NetworkServerConfiguration configuration;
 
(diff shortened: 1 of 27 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoraDeviceInfoManager.cs
```diff
@@ -242,7 +244,7 @@ public static async Task<LoraDeviceInfo> PerformOTAAAsync(string GatewayID, stri
             //}
 
 
-            joinLoraDeviceInfo.HubSender = new IoTHubConnector(joinLoraDeviceInfo.DevEUI, joinLoraDeviceInfo.PrimaryKey);
+            joinLoraDeviceInfo.HubSender = new IoTHubConnector(joinLoraDeviceInfo.DevEUI, joinLoraDeviceInfo.PrimaryKey, this.configuration);
```
Mandur: Hello

PR #41, opened 2018-11-26
Lo ren dev

Ellerbach:
Adding support for RAK833 modules connected thru USB for arm7 platforms
- New module with full container creation
- Added environement variable into the example.env file
- Added documentation
Updated main documentation


Mandur:
PR redone in #45

PR #43, opened 2018-11-28
Implementation of Join Req Mic Check, Fport down, Conf. Down

Mandur:
This **PR changes Core processing** of LoRa messages with the following features

* Added MIC check on join request message
* Added possibility to send Confirmed cloud to device messages
* Added Option to set the Fport of cloud to device messages 
* Changed Fctrl flow to an Enum flag for better clarity
* Write tests for all above


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -340,24 +347,36 @@ private async Task<byte[]> ProcessLoraMessage(LoRaMessageWrapper loraMessage)
...
+                                    if (c2dMsg.Properties.Keys.Contains("CidType"))
                                     {
-                                        MacCommandHolder macCommandHolder = new MacCommandHolder(Convert.ToByte(macCmd.First().Value));
-                                        macbytes = macCommandHolder.macCommand[0].ToBytes();
+                                            MacCommandHolder macCommandHolder = new MacCommandHolder(Convert.ToByte(c2dMsg.Properties["CidType"]));
+                                            macbytes = macCommandHolder.macCommand[0].ToBytes();
                                     }
-                                    var confirmCmd = c2dMsg.Properties.Where(o => o.Key == "Confirmed");
-                                    if (confirmCmd.Count() != 0)
+                                    if (c2dMsg.Properties.Keys.Contains("Confirmed")&&c2dMsg.Properties["Confirmed"] == "true")
```
fbeltrao: Suggestion: use case insensitive comparison (true, TRUE, True) == true

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -340,24 +347,36 @@ private async Task<byte[]> ProcessLoraMessage(LoRaMessageWrapper loraMessage)
...
                                     {
                                         requestForConfirmedResponse = true;
+                                        Logger.Log(loraDeviceInfo.DevEUI, $"Cloud to device message requesting a confirmation", Logger.LoggingLevel.Info);
+
                                     }
+                                    if (c2dMsg.Properties.Keys.Contains("Fport"))
+                                    {
+                                        int fportint= int.Parse(c2dMsg.Properties["Fport"]);
```
fbeltrao: Suggestion: if FPort is not numeric will thrown an exception. You can use 
```
if (!int.TryParse(c2dMsg.Properties["Fport"], out var fportint))
  fportint = 1; // or whatever is the default value
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -11,6 +11,18 @@
 
 namespace LoRaTools.LoRaMessage
 {
+
+    public enum FctrlEnum : short
```
fbeltrao: Does it make to add the [Flags] attribute?

LoRaEngine/test/LoRaWan.IntegrationTest/C2DMessageTest.cs
```diff
@@ -245,5 +245,222 @@ public async Task Test_OTAA_Unconfirmed_Receives_C2D_Message()
             Assert.True(foundReceivePacket, $"Could not find lora receiving message '{expectedRxSerial}'");
                 
         }
+
+        // Ensures that C2D messages are received when working with unconfirmed messages
+        // Uses Device10_OTAA
```
fbeltrao: Correct comment would be Uses Device_15_OTAA

LoRaEngine/test/LoRaWan.IntegrationTest/C2DMessageTest.cs
```diff
@@ -245,5 +245,222 @@ public async Task Test_OTAA_Unconfirmed_Receives_C2D_Message()
...
+
+            await this.ArduinoDevice.setDeviceModeAsync(LoRaArduinoSerial._device_mode_t.LWOTAA);
+            await this.ArduinoDevice.setIdAsync(device.DevAddr, device.DeviceID, device.AppEUI);
+            await this.ArduinoDevice.setKeyAsync(device.NwkSKey, device.AppSKey, device.AppKey);
+
+            await this.ArduinoDevice.SetupLora(this.TestFixture.Configuration.LoraRegion);
+
+            var joinSucceeded = await this.ArduinoDevice.setOTAAJoinAsyncWithRetry(LoRaArduinoSerial._otaa_join_cmd_t.JOIN, 20000, 5);
+
+            if (!joinSucceeded)
```
fbeltrao: The code you copied from me was sub-optimal. We don't need an if here. We can just do the Assert.True(...)

LoRaEngine/test/LoRaWan.IntegrationTest/IntegrationTestFixture.cs
```diff
@@ -242,7 +248,29 @@ void SetupTestDevices()
...
+                AppKey = "8AFE71A145B253E49C3031AD068277A3",
+                GatewayID = gatewayID,
+                IsIoTHubDevice = true,
+                SensorDecoder = "DecoderValueSensor",
+            };
+
+            // Device15_OTAA: used for the Fport test
+            this.Device15_OTAA = new TestDeviceInfo()
+            {
+                DeviceID = "0000000000000013",
```
fbeltrao: DeviceID should end with 15

LoRaEngine/test/LoRaWan.IntegrationTest/IntegrationTestFixture.cs
```diff
@@ -242,7 +248,29 @@ void SetupTestDevices()
                 GatewayID = gatewayID,
                 IsIoTHubDevice = true,                                      
                 SensorDecoder = "DecoderValueSensor",                           
-            };  
+            };
+
+            // Device14_OTAA: used for Confirmed C2D message
+            this.Device14_OTAA = new TestDeviceInfo()
+            {
+                DeviceID = "0000000000000013",
```
fbeltrao: DeviceID should end with 14

LoRaEngine/test/LoRaWan.IntegrationTest/IntegrationTestFixture.cs
```diff
@@ -296,9 +324,16 @@ internal async Task<Twin> GetTwinAsync(string deviceId)
             return await GetRegistryManager().GetTwinAsync(deviceId);            
         }
 
-        internal async Task SendCloudToDeviceMessage(string deviceId, string messageText)
+        internal async Task SendCloudToDeviceMessage(string deviceId, string messageText, Dictionary<String,String> messageProperties)
```
fbeltrao: Tip 1: when adding an optional parameter to an existing method you can set a default value. In that case you don't have to fix all the other callers that will pass the default parameter. 

In this example:
```
internal async Task SendCloudToDeviceMessage(string deviceId, string messageText, Dictionary<String,String> messageProperties = null)
```

Tip 2: When you don't need an concrete collection implementation it is a good practice to use interfaces:
Dictionary<> &rarr; IDictionary<>
List<> &rarr; ICollection<> or IEnumerable<> (this way you can use with arrays, lists, hashsets, etc)

LoRaEngine/test/LoRaWan.IntegrationTest/TestLogger.cs
```diff
@@ -17,15 +17,13 @@ static class TestLogger
         /// <param name="text"></param>
         internal static void Log(string text)
         {
-            // Log to diagnostics if a debbuger is attached
-            if (Debugger.IsAttached)
+            Console.WriteLine(text);
+
```
fbeltrao: Any reason why you reverted this change?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoraMessageWrapper.cs
```diff
@@ -179,7 +179,7 @@ public LoRaMessageWrapper(LoRaPayload payload, LoRaMessageType type, byte[] phys
             LoRaPayloadMessage = payload;
             PktFwdPayload = new DownlinkPktFwdMessage(Convert.ToBase64String(LoRaPayloadMessage.GetByteMessage()), datr, rfch, freq, tmst);
             var jsonMsg = JsonConvert.SerializeObject(PktFwdPayload);
-            Logger.Log(ConversionHelper.ByteArrayToString(payload.GetLoRaMessage().DevEUI.ToArray()),$"{((MType)(payload.Mhdr.Span[0])).ToString()} {jsonMsg}", Logger.LoggingLevel.Full);
+            Logger.Log(ConversionHelper.ByteArrayToString(payload.DevAddr.Span.ToArray()),$"{((MType)(payload.Mhdr.Span[0])).ToString()} {jsonMsg}", Logger.LoggingLevel.Full);
```
ronniesa: We log the devaddr because we don't have yet the deveui right?

LoRaEngine/test/LoRaWan.IntegrationTest/TestLogger.cs
Mandur: No, I feel it might be due to the multiple rebase i had to do, should I revert it to the old version?

fbeltrao: My version looks like this:
```
// Log to diagnostics if a debbuger is attached
            if (Debugger.IsAttached)
            {
                System.Diagnostics.Debug.WriteLine(text);
            }
            else
            {
                Console.WriteLine(text);
            }
```

Up to you. I will fix with my CI PR later this week.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoraMessageWrapper.cs
Mandur: Yes, Having the DevEUI here would be possible, but need a substantial refactor of the current downstream pipe. I will add a Workitem and we could prioritize this accordingly

ronniesa: Not a priority but it is not great to follow the log.

Mandur:
@fbeltrao @ronniesa  I modified my code according to your comments, let me know if there are any additional issues. Thank you for the review.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -336,28 +342,42 @@ private async Task<byte[]> ProcessLoraMessage(LoRaMessageWrapper loraMessage)
...
+                   
+                                    if (c2dMsg.Properties.TryGetValueCaseInsensitive("confirmed", out var confirmedValue) && confirmedValue.Equals("true", StringComparison.OrdinalIgnoreCase))
                                     {
                                         requestForConfirmedResponse = true;
+                                        Logger.Log(loraDeviceInfo.DevEUI, $"Cloud to device message requesting a confirmation", Logger.LoggingLevel.Info);
+
+                                    }
+                                    if (c2dMsg.Properties.TryGetValueCaseInsensitive( "fport", out var fPortValue))
+                                    {
+                                        int fportint = int.Parse(fPortValue);
```
fbeltrao: We should check what we do if the value coming is not as expected:
- Let exception be raised
- Use a default and log a warning
- Ignore C2D and log a warning/error

ronniesa: In general there can be many different issues like fport to big, message to big for the DR and country etc... view that we cannot control the UI we need an easy way to inform the "user" of the issue. Changing the fport number to 1 is not a good idea and probably cutting the message as we do now also not. My feeling says refuse and log as error.

Mandur: As the fport could direct this data to very different logical endpoints, I feel we should ignore the C2D and

ronniesa: We can offer a CLI utility that sends the c2d where we can do some checks but we still need to manage the case that they use the UI or other tools

Mandur:
Fix AB#657

PR #47, opened 2018-11-30
Refactor UDP Server

Mandur:
* Added support and ACKs for packet forwarder
* Added log receipt for transmission to pktfwd (to be improved)
* Simplified MessageProcessor

This PR change the Core LoRa Engine, please review carefully

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
@@ -35,7 +35,6 @@ namespace LoRaWan.NetworkServer
         {
             startTimeProcessing = DateTime.UtcNow;
             LoRaMessageWrapper loraMessage = new LoRaMessageWrapper(message);
-            byte[] udpMsgForPktForwarder = new Byte[0];
             if (loraMessage.IsLoRaMessage)
             {
                 if (RegionFactory.CurrentRegion == null)
@@ -43,17 +42,15 @@ namespace LoRaWan.NetworkServer
                 //join message
                 if (loraMessage.LoRaMessageType == LoRaMessageType.JoinRequest)
                 {
-                    udpMsgForPktForwarder = await ProcessJoinRequest(loraMessage);
+                    return await ProcessJoinRequest(loraMessage);
                 }
                 //normal message
                 else if (loraMessage.LoRaMessageType == LoRaMessageType.UnconfirmedDataUp || loraMessage.LoRaMessageType == LoRaMessageType.ConfirmedDataUp)
                 {
-                    udpMsgForPktForwarder = await ProcessLoraMessage(loraMessage);
+                    return await ProcessLoraMessage(loraMessage);
                 }
             }
-
-            //send reply to pktforwarder
-            return udpMsgForPktForwarder;
+            return null;
 
         }
         private async Task<byte[]> ProcessLoraMessage(LoRaMessageWrapper loraMessage)
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
@@ -350,7 +350,7 @@ namespace LoRaWan.NetworkServer
                                         Logger.Log(loraDeviceInfo.DevEUI, $"Cloud to device message with a Fport of " + fPortValue, Logger.LoggingLevel.Info);
 
                                     }
-                                    Logger.Log(loraDeviceInfo.DevEUI,String.Format("Sending a C2D message with ID {0}",
+                                    Logger.Log(loraDeviceInfo.DevEUI,String.Format("Sending a downstream message with ID {0}",
                                         ConversionHelper.ByteArrayToString(rndToken)),
                                         Logger.LoggingLevel.Full);
                                 }
@@ -605,11 +605,14 @@ namespace LoRaWan.NetworkServer
                     }
                 }
                 LoRaMessageWrapper joinAcceptMessage = new LoRaMessageWrapper(loRaPayloadJoinAccept, LoRaMessageType.JoinAccept, loraMessage.PhysicalPayload.token, datr, 0, freq, tmst);
+
                 udpMsgForPktForwarder = joinAcceptMessage.PhysicalPayload.GetMessage();
 
                 //add to cache for processing normal messages. This awoids one additional call to the server.
                 Cache.AddToCache(joinLoraDeviceInfo.DevAddr, joinLoraDeviceInfo);
-                Logger.Log(devEui, $"join accept sent", Logger.LoggingLevel.Info);
+                Logger.Log(devEui, String.Format("join accept sent with ID {0}", 
+                    ConversionHelper.ByteArrayToString(loraMessage.PhysicalPayload.token)), 
+                    Logger.LoggingLevel.Full);
             }
             else
             {
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
@@ -36,11 +36,7 @@ namespace LoRaWan.NetworkServer
             startTimeProcessing = DateTime.UtcNow;
             LoRaMessageWrapper loraMessage = new LoRaMessageWrapper(message);
             byte[] udpMsgForPktForwarder = new Byte[0];
-            if (!loraMessage.IsLoRaMessage)
-            {
-                udpMsgForPktForwarder = ProcessNonLoraMessage(loraMessage);
-            }
-            else
+            if (loraMessage.IsLoRaMessage)
             {
                 if (RegionFactory.CurrentRegion == null)
                     RegionFactory.Create(loraMessage.PktFwdPayload.GetPktFwdMessage().Rxpks[0]);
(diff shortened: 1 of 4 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -76,36 +78,86 @@ async Task RunUdpListener()
...
-                    }
-                });                    
-                #pragma warning restore CS4014 
+                }
+#pragma warning restore CS4014
+
             }
         }
 
+        private void sendAcknowledgementMessage(UdpReceiveResult receivedResults, byte messageType, int remotePort)
```
fbeltrao: C# convention is to have method names in Pascal case.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -76,36 +78,86 @@ async Task RunUdpListener()
...
                 UdpReceiveResult receivedResults = await udpClient.ReceiveAsync();
 
-                //Logger.Log($"UDP message received ({receivedResults.Buffer.Length} bytes) from port: {receivedResults.RemoteEndPoint.Port}");
+               //Logger.Log($"UDP message received ({receivedResults.Buffer[3]}) from port: {receivedResults.RemoteEndPoint.Port} and IP: {receivedResults.RemoteEndPoint.Address.ToString()}",LoggingLevel.Always);
+                 
 
-
-                //Todo check that is an ack only, we could do a better check in a future verstion
-                if (receivedResults.Buffer.Length == 12)
+                switch (receivedResults.Buffer[3])
```
fbeltrao: I have a few suggestions to, maybe, improve code readability and chance of running into unhandled exceptions.

Move all the byte manipulation to classes that already do that, namely PhysicalPayload. Moreover, accessing receivedResults[3] without checks will kill the process if it is null or has length < 4.

As suggestion, to improving the readability, I would make a few changes to the code to not rely on magic numbers (`switch buffer[3]`, `case 0x02`), specially that you already have an enum defining these values.

* Move low level manipulation to PhysicalPayload class. 

```C#

public enum PhysicalIdentifier : byte // <- use byte here since you know it has to fit a into a byte
{
        PUSH_DATA, PUSH_ACK, PULL_DATA, PULL_RESP, PULL_ACK, TX_ACK, 
        UNKOWN = byte.MaxValue // <-- can become a problem in future if they ever use 255 to define a message type
}

public class PhysicalPayload
{
        // Position in payload where the identifier is located (3)
        const int PHYSICAL_IDENTIFIER_INDEX = 3;

        // Gets the identifier of the payload
        public static PhysicalIdentifier GetIdentifierFromPayload(byte[] packet)
        {
            // Unknown if: packet is null or does not have the physical identifier byte
            if (packet == null || packet.Length < (PHYSICAL_IDENTIFIER_INDEX + 1))
                return PhysicalIdentifier.UNKOWN;

            return (PhysicalIdentifier)packet[PHYSICAL_IDENTIFIER_INDEX];
        }
  
         ...
}
```

* Clear UdpServer.cs a bit. Don't forget to check the case where buffer.Length < 12 on the ProcessTxAck method.

```C#
 async Task RunUdpListener()
 {


            IPEndPoint endPoint = new IPEndPoint(IPAddress.Any, PORT);
            udpClient = new UdpClient(endPoint);

            Logger.Log($"LoRaWAN server started on port {PORT}", Logger.LoggingLevel.Always);


            while (true)
            {
                UdpReceiveResult receivedResults = await udpClient.ReceiveAsync();

               //Logger.Log($"UDP message received ({receivedResults.Buffer[3]}) from port: {receivedResults.RemoteEndPoint.Port} and IP: {receivedResults.RemoteEndPoint.Address.ToString()}",LoggingLevel.Always);
                
                switch (PhysicalPayload.GetIdentifierFromPayload(receivedResults.Buffer))
                {
                    case PhysicalIdentifier.PULL_DATA:
                    {
                        //In this case we have a keep-alive PULL_DATA packet we don't need to start the engine and can return immediately a response to the challenge                                            
                        if (pullAckRemoteLoRaAggregatorPort == 0)
                        {
                            remoteLoRaAggregatorIp = receivedResults.RemoteEndPoint.Address;
                            pullAckRemoteLoRaAggregatorPort = receivedResults.RemoteEndPoint.Port;
                        }
                        SendAcknowledgementMessage(receivedResults, PhysicalIdentifier.PUSH_DATA, pullAckRemoteLoRaAggregatorPort);
                        break;
                    }


                    
                    case PhysicalIdentifier.PUSH_DATA:
                    {
                        //This is a PUSH_DATA (upstream message).
                        if (pushDataRemoteLoRaAggregatorPort == 0)
                        {
                            remoteLoRaAggregatorIp = receivedResults.RemoteEndPoint.Address;
                            pushDataRemoteLoRaAggregatorPort = receivedResults.RemoteEndPoint.Port;
                        }
                        SendAcknowledgementMessage(receivedResults, PhysicalIdentifier.PUSH_ACK, pushDataRemoteLoRaAggregatorPort);

                        // Message processing runs in the background
                        MessageProcessor messageProcessor = new MessageProcessor(this.configuration, this.loraDeviceInfoManager);

                        _ = Task.Run(async () => {
                            try
                            {
                                var resultMessage = await messageProcessor.ProcessMessageAsync(receivedResults.Buffer);
                                if (resultMessage != null && resultMessage.Length > 0)
                                    await this.UdpSendMessage(resultMessage, remoteLoRaAggregatorIp.ToString(), pullAckRemoteLoRaAggregatorPort);
                            }
                            catch (Exception ex)
                            {
                                Logger.Log($"Error processing the message {ex.Message}, {ex.StackTrace}", Logger.LoggingLevel.Error);
                            }
                        });

                        break;
                    }

                    
                    case PhysicalIdentifier.TX_ACK:
                    {
                        //This is a ack to a transmission we did previously
                        ProcessTxAck(receivedResults.Buffer);
                        break;
                    }

                    default:
                    {
                        Logger.Log("Unknown packet type being received", LoggingLevel.Error);
                        break;
                     }
                }

            }
        }

        private void ProcessTxAck(byte[] data)
        {            
            if (data == null || data.Length != 12)
            {
                Logger.Log(String.Format("Packet with id {0} had a problem to be transmitted over the air :{1}",
                ConversionHelper.ByteArrayToString(data.RangeSubset(1, 2)),

                // TODO: MIK: this will throw an exception if data.Length < 12
                Encoding.UTF8.GetString(data.RangeSubset(12, data.Length - 12)))
                , LoggingLevel.Error);
            }
            else
            {
                    Logger.Log(String.Format("Packet with id {0} successfully transmitted by the aggregator",
                ConversionHelper.ByteArrayToString(data.RangeSubset(1, 2)))
                , LoggingLevel.Full);
            }
        }

        private void SendAcknowledgementMessage(UdpReceiveResult receivedResults, PhysicalIdentifier messageType, int remotePort)
        {
            // OPTIONAL: Use PhysicalPayload as the single logical place where low level messages are built:
            // var response = PhysicalPayload.CreateAck(receivedResults.Buffer, messageType);

            byte[] response = new byte[4]{
                            receivedResults.Buffer[0],
                            receivedResults.Buffer[1],
                            receivedResults.Buffer[2],
                            (byte)messageType
                            };
            _ = UdpSendMessage(response, remoteLoRaAggregatorIp.ToString(), remotePort);
        }
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -22,12 +24,12 @@ public class UdpServer : IDisposable
         private readonly NetworkServerConfiguration configuration;
         private readonly LoraDeviceInfoManager loraDeviceInfoManager;
         ModuleClient ioTHubModuleClient;
+        private IPAddress remoteLoRaAggregatorIp;
+        private int pullAckRemoteLoRaAggregatorPort=0;
```
fbeltrao: I understand what you mean with keeping the port numbers.
Push and pull have different ports and you need to use the correct one based on the payload we are sending.

Mandur: I need to keep at least the variable pullAckRemoteLoRaAggregatorPort as a variable, as it is transmitted by a PULL_ACK. If I want to send PULL_RESP (downstream message) after receiving a PUSH DATA (upstream message), I would need to send it on this port, that i can only have if I save it. 
I could save the variable pushDataRemoteLoRaAggregatorPort as I only need it to answer to Push_Data answer.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -36,11 +36,7 @@ public async Task<byte[]> ProcessMessageAsync(byte[] message)
             startTimeProcessing = DateTime.UtcNow;
             LoRaMessageWrapper loraMessage = new LoRaMessageWrapper(message);
             byte[] udpMsgForPktForwarder = new Byte[0];
```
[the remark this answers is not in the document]
Mandur: The goal of the ProcessNonLoRaMessage Method was to deal with the PULL_DATA messages. They are now managed at the UDP server level and acked directly from there (so we never generate a new thread)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
fbeltrao: At least in the code there is a chance that you call UdpSendMessage using pullAckRemoteLoRaAggregatorPort having value 0

Scenario:

1. Module starts
2. PUSH_DATA request arrives
    The Task.Run could call UdpSendMessage with pullAckRemoteLoRaAggregatorPort being zero.

Could this happen?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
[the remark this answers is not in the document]
fbeltrao: Cool. This method should never return null then?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
Mandur: I don't think so, the goal of the PULL_DATA is to check that network route is open. If there are no reply, the packet forwarder would think the connection is closed. Here are some quotes from the specs that makes me think like this: 
This data exchange is initialized by the gateway because it might be 
impossible for the server to send packets to the gateway if the gateway is 
behind a NAT.

When the gateway initialize the exchange, the network route towards the 
server will open and will allow for packets to flow both directions.
The gateway must periodically send PULL_DATA packets to be sure the network 
route stays open for the server to be used at any time.

Mandur:
I included the suggested change. Let me know what you think

PR #52, opened 2018-12-06
Update MessageProcessor.cs to fix bug 798

TechPreacher:
Now correctly accepts all of the following loraDeviceInfo.SensorDecoder:
- empty / null: Just return base64 decoded payload data
- local decoder name
- external decoder container url

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
@@ -149,7 +149,7 @@ namespace LoRaWan.NetworkServer
                             // Decoder is set in LoRa Device Twin. Send decrpytedMessage (payload) and fportUp (fPort) to decoder.
                             else
                             {
-                                Logger.Log(loraDeviceInfo.DevEUI, $"decoding with: {loraDeviceInfo.SensorDecoder} port: {fportUp}", Logger.LoggingLevel.Info);
+                                Logger.Log(loraDeviceInfo.DevEUI, $"decoding with: {loraDeviceInfo.SensorDecoder} port: {fportUp}", Logger.LoggingLevel.Full);
                                 fullPayload.data = await LoraDecoders.DecodeMessage(decryptedMessage, fportUp, loraDeviceInfo.SensorDecoder);
                             }
 
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
@@ -135,11 +135,18 @@ namespace LoRaWan.NetworkServer
                             fullPayload.port = fportUp;
                             fullPayload.fcnt = fcntup;
 
-                            if (isAckFromDevice)
+                           // ACK from device or no decoder set in LoRa Device Twin. Simply return decryptedMessage
+                            if (isAckFromDevice || String.IsNullOrEmpty(loraDeviceInfo.SensorDecoder))
                             {
+                                if (String.IsNullOrEmpty(loraDeviceInfo.SensorDecoder))
+                                {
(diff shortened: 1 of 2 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -135,11 +135,18 @@ private async Task<byte[]> ProcessLoraMessage(LoRaMessageWrapper loraMessage)
                             fullPayload.port = fportUp;
                             fullPayload.fcnt = fcntup;
 
-                            if (isAckFromDevice)
+                           // ACK from device or no decoder set in LoRa Device Twin. Simply return decryptedMessage
+                            if (isAckFromDevice || String.IsNullOrEmpty(loraDeviceInfo.SensorDecoder))
                             {
+                                if (String.IsNullOrEmpty(loraDeviceInfo.SensorDecoder))
+                                {
+                                    Logger.Log(loraDeviceInfo.DevEUI, $"no decoder set in device twin. port: {fportUp}", Logger.LoggingLevel.Info);
```
ronniesa: If you want to log it you should log it at full not info if not we are logging too much and this is not a really needed info because you will see in the upstream message that is not decoded and is a common usage that you have no decoders because the end application will decode it eventualy

fbeltrao: Shouldn't we then change the log level if we find a decoder too? Logging the decoder used (with and without one) should be in the same log level, right?

Moreover, if we change the logging level we need to fix the integration test that is using this log information for assertion

TechPreacher: Changed based on Ronnie's request but holding until we reach consensus on Francisco's input.

ronniesa: Xico you are right we should log only and full the decoder logic for the happy path because you will see the decoded json or the original payload so we save some logging/verbosity

TechPreacher: How about line 194 on sending the message to the Hub? Keep this on Info?
```
Logger.Log(loraDeviceInfo.DevEUI, $"no decoder set in device twin. port: {fportUp}", Logger.LoggingLevel.Info);
```

fbeltrao: If we go ahead with the change of only logging in max level the decoder information it will break the integration test:
```c#
                // 0000000000000004: decoding with: DecoderValueSensor port: 8
                await this.TestFixture.AssertNetworkServerModuleLogStartsWithAsync($"{device.DeviceID}: decoding with: {device.SensorDecoder} port:");

```

We either remove this assertion or increase logging level when running integration test.

ronniesa: now with the UDP we can increase the level, I really want to have a minmal yet useful logging for in production or in pilot end users

fbeltrao: In that case, can you please set NET_SRV_LOG_LEVEL to 1 in azure-pipelines.yaml @TechPreacher ?

fbeltrao: Regarding the other message?
```c#
Logger.Log(loraDeviceInfo.DevEUI, $"message '{fullPayloadAsString}' sent to hub", Logger.LoggingLevel.Info);
```
 

I think we still log in info level, right?

ronniesa: yes because we can see that will be sent to iot hub so you don't need to monitor the devices to see the message I think is quite useful to the end users

PR #54, opened 2018-12-07
Fail join request in case we can't save device twins

Mandur:
Changed join processing flow to fail if we can't save the device twin


ronniesa:
Yes if we fail saving the twins we should never send a join accept. Because it means that the keys are only in memory so at the restart of the module the device cannot send msg anymore until is rejoinig that this could be after 2 months. Alternative would be to do it asyn without waiting with a really long exp retry policy but if the module crashes we are tosted

fbeltrao:
Thanks for the clarification. That goes back to the dilemma we discussed previously wether or not we should be able to deal with edgeHub hiccups.

fbeltrao:
Sorry, clicked wrong button.

PR #62, opened 2018-12-14
Added support for random interval plus group rxpk

Ellerbach:
Added support for:
* Sending array with multiple request of devices in the rxpk packet
* Having a more random time when sending the data that can be customzed. Initial time +- a random one
Fixed:
* Removed random data sending for an incremental data of +1 every time a joined or ABP device send data
* Always log the DevEUI of the device rather than DevAddr
Updated documentation

### DIFF
--- a/LoRaEngine/modules/LoRaSimulator/LoRaSimulator/LoRaSimulator.csproj
+++ b/LoRaEngine/modules/LoRaSimulator/LoRaSimulator/LoRaSimulator.csproj
@@ -9,6 +9,9 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Update="testconfig - Copy.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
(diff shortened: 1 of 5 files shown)


LoRaEngine/modules/LoRaSimulator/LoRaSimulator/LoRaSimulator.csproj
```diff
@@ -9,6 +9,9 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Update="testconfig - Copy.json">
```
Mandur: Is the copy of the Copy file normal?

Ellerbach: hum, that should not be present. I guess commit went before the project file has been saved in VS. Please remove this part. thank you

Mandur:
Thank you @Ellerbach for your contributions!

PR #64, opened 2018-12-17
Validate function version from network server

fbeltrao:
**Changes**
 
* Added Network Server unit test to validate function versioning (see *Versioning implementation* for details)
* Moved DeviceGetter and FCntCacheCheck classes to own files
* Added ApiVersion class in LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApiVersion.cs. File is shared between network server module and Service Facade function
* LoraDeviceInfoManager interaction with HttpClient has been abtracted with `IServiceFacadeHttpClientProvider`. It creates a singleton HttpClient where proxy and versioning are handled.
* Sensor decoder uses a single HttpClient instance
* Small change to UdpServer, if a port is not found in `pullAckRemoteLoRaAggregatorPort` it uses the one available from the udp request.
* Initial tests with simulated device have been added. They are not yet active. Thanks to @Ellerbach contribution!
 
**Versioning implementation**
 
*Function changes*
 
* Version will be returned in response header (api-version)
* Starting from 2018-12-16-preview a bad request will be returned if an unknown or newer version is requested. Return message is: `Incompatible versions (requested: '{requestedVersion.Name ?? string.Empty}', current: '{currentApiVersion.Name}')"`
* Newer versions will handle older version requests normally, keeping backward compatibility. Keep in mind that the current version and not the requested version will be sent in response header
 
*Network Server Module changes*
 
* Versioning of Service Facade API is handled in a HttpClientHandler. Current version (defined in `ApiVersion.LatestVersion`) is added to query string parameter.
* If the response header "api-version" does not exists or returns an older version the middleware will return a BadRequest. More information is available in HttpResponseMessage.ReasonPhrase and in the response body
 
*Examples*
 
|Facade|NtwSrv|Result|
|-|-|-|
|0.2|2018-12-16-preview|network server call will report failure|
|2018-12-16-preview|0.2|it works|
|2018-12-16-preview|2018-12-16-preview|it works|
|2019-01-30-preview|2018-12-16-preview|it works|
|2018-12-16-preview|2019-01-30-preview|service facade will return BadRequestObjectResult| 

**NOTE**: Version 2019-01-30-preview is hipothetical new version.

Fixes AB#871, AB#770


fbeltrao:
Cancelling this, #65 replaces it

PR #66, opened 2018-12-18
Add sample code for decoding string and binary payloads

TechPreacher:
- Add ByteArrayToString helper method
- Add example decoders for string and binary payloads
- Remove unnecessary using statements

### DIFF
--- a/Samples/DecoderSample/Classes/LoraDecoders.cs
+++ b/Samples/DecoderSample/Classes/LoraDecoders.cs
@@ -1,19 +1,23 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿using System.Text;
 using Newtonsoft.Json;
 
 namespace SensorDecoderModule.Classes
 {
     internal static class LoraDecoders
     {   
-        private static string DecoderValueSensor(byte[] payload, uint fport)
+        private static string DecoderStringValueSensor(byte[] payload, uint fport)
         {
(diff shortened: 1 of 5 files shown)


Samples/DecoderSample/Classes/LoraDecoders.cs
```diff
@@ -1,19 +1,23 @@
...
-using System.Threading.Tasks;
+﻿using System.Text;
 using Newtonsoft.Json;
 
 namespace SensorDecoderModule.Classes
 {
     internal static class LoraDecoders
     {   
-        private static string DecoderValueSensor(byte[] payload, uint fport)
+        private static string DecoderStringValueSensor(byte[] payload, uint fport)
```
ronniesa: It is a bit tricky that we have in the NtwSrv the name DecoderValueSensor and here the same with another name. Also if you change it here the doc should reflect it. I would keep it DecoderValueSensor if not we need to change docs and the NtwSrv

TechPreacher: Good input. I keep one DecoderValueSensor and use code & comments inside to show how to decode both type of payloads...!

TechPreacher: Updated!

TechPreacher:
@ronniesa what do you think? Leave all in DecoderValueSensor or follow @fbeltrao 's input to make DecoderValueSensor and DecoderBinaryValueSensor? Lots of discussion on a small change. Sorry for that!

ronniesa:
I would say keep it like this if not we need to document more etc..,..

PR #76, opened 2019-01-09
Instantiate device client using MQTT or AMQP based on env. var.

TechPreacher:
Based on LoRaWanNetworkSrvModule environment variable LEAF_DEVICE_PROTOCOL which can be Mqtt_Tcp_Only or Amqp_Tcp_Only.
Defaults to Mqtt_Tcp_Only if variable is not found


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubConnector.cs
```diff
@@ -151,10 +151,17 @@ private void CreateDeviceClient()
...
-
-                    deviceClient = DeviceClient.CreateFromConnectionString(deviceConnectionStr, TransportType.Amqp_Tcp_Only);
-
-                   
+                    
+                    if (configuration.LeafDeviceProtocol.ToUpper() == "AMQP_TCP_ONLY")
+                    {
+                        deviceClient = DeviceClient.CreateFromConnectionString(deviceConnectionStr, TransportType.Amqp_Tcp_Only);
+                        Logger.Log(DevEUI, $"device client created using Amqp_Tcp_Only", Logger.LoggingLevel.Info);
+                    }
```
ronniesa: I would log as full not as info because is not useful for non developers

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubConnector.cs
```diff
@@ -151,10 +151,17 @@ private void CreateDeviceClient()
...
+                    else
+                    {
+                        deviceClient = DeviceClient.CreateFromConnectionString(deviceConnectionStr, TransportType.Mqtt_Tcp_Only);
+                        Logger.Log(DevEUI, $"device client created using Mqtt_Tcp_Only", Logger.LoggingLevel.Info);
+                    }
```
ronniesa: same as above

TechPreacher: Done!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubConnector.cs
fbeltrao: Instead of making upper and comparing use string.Equals(configuration.LeafDeviceProtocol, "amqp_tcp_only", StringComparison.InvariantCultureIgnoreCase).
It avoids creating a new instance of the string.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -260,7 +260,13 @@ private async Task<DownlinkPktFwdMessage> ProcessLoraMessage(LoRaMessageWrapper
                         {
                             //put it back to the queue for the next pickup
                             //todo ronnie check abbandon logic especially in case of mqtt
-                            _ = await loraDeviceInfo.HubSender.AbandonAsync(secondC2dMsg);
+
+                            // Don't call AbandonAsync when Mqtt_Tcp_Only is used.
+                            if(configuration.LeafDeviceProtocol.ToUpper() != "MQTT_TCP_ONLY")
```
fbeltrao: Same as before regarding string comparison

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -439,7 +445,11 @@ private async Task<DownlinkPktFwdMessage> ProcessLoraMessage(LoRaMessageWrapper
                             //todo ronnie check abbandon logic especially in case of mqtt
                             if (c2dMsg != null)
                             {
-                                _ = await loraDeviceInfo.HubSender.AbandonAsync(c2dMsg);
+                                // Don't call AbandonAsync when Mqtt_Tcp_Only is used.
+                                if(configuration.LeafDeviceProtocol.ToUpper() != "MQTT_TCP_ONLY")
```
fbeltrao: Same as before regarding string comparison

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubConnector.cs
TechPreacher: - Changed logging levels of the DeviceClient creation to Full.
- Added NET_SRV_LEAF_DEVICE_PROTOCOL environment variable to deployment.*.template.json files and to example.env
- Updated azure-pipeline.yaml to EDGEHUB_MQTTSETTINGS_ENABLED: true

LoRaEngine/deployment.test.template.json
```diff
@@ -98,6 +98,9 @@
               },
               "LOG_TO_UDP_ADDRESS": {
                 "value": "$NET_SRV_LOG_TO_UDP_ADDRESS"
+              },
+              "LEAF_DEVICE_PROTOCOL": {
```
fbeltrao: This change requires you to modify the azure-pipelines.yaml file, defining the value

TechPreacher: Ok. I'm on it. Sorry, I'm pretty new to the CI pipeline workings

ronniesa: @fbeltrao I'm not sure I want to document the MQTT settings is something internal to the engine that should not be changed. We want it as parameter so we can easily test what works best but once we decide we stick with it

ronniesa: @fbeltrao from my early test with MQTT if you don't do abandon but also no complete next pass you get the message so it is not an issue but we need to test. for a while we where using MQTT. The issue we had with MQTT is that sometime (1 msg only) will show up on the next pass even after complete is called.....

TechPreacher:
Closed as discussed. We will keep the branch in case we need to re-visit MQTT

PR #91, opened 2019-01-22
Fix UDP crash when message arrive while starting

Mandur:
Fix AB#955

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
@@ -215,7 +215,7 @@ namespace LoRaWan.NetworkServer
                     {
                         Logger.Log("UDP", 
                             "Waiting for first pull_ack message from the packet forwarder. The received message was discarded as the network server is still starting.",
-                            Logger.LoggingLevel.Error);
+                            Logger.LoggingLevel.Info);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -204,10 +204,19 @@ private async Task ProcessRxpkAsync(String remoteIp, Rxpk rxpk, DateTime startTi
...
+                    if (pullAckRemoteLoRaAggregatorPort != 0)
+                    {
+                        await this.UdpSendMessage(pyld.GetMessage(), remoteIp, pullAckRemoteLoRaAggregatorPort);
+                        Logger.Log("UDP", String.Format("message sent with ID {0}",
+                            ConversionHelper.ByteArrayToString(token)),
+                            Logger.LoggingLevel.Info);
+                    }
+                    else
+                    {
+                        Logger.Log("UDP",
```
ronniesa: Are we sure that we want to log as error? I think this can happen so should not be an error

Mandur: I felt this could lead to some confusion, so decided to always log it. Happy to revert to Info if you think it's more appropriate.

ronniesa: Info is better, if not it would alarm the user for a startup inconvenient that we can do nothing about

Mandur:
@ronniesa done

Mandur:
@ronniesa done to full

PR #99, opened 2019-01-25
Update Logger to log using Microsoft.Extensions.Logging.LogLevel Enum

TechPreacher:
- Logger: Log using Microsoft.Extensions.Logging.LogLevel Enum
- Logger: new LogAlways method
- LoggerConfiguration: handle various LOG_LEVEL env var values: 
-- 1 or Debug
-- 2 or Information or Info
-- 3 or Error
- UdpServer: Output Log Level on startup
- Other classes: Adapt for new Logger.Log and Logger.LogAlways methods, include using Microsoft.Extensions.Logging;
- ReadMe: Update with new LOG_LEVEL env var settings and point out to use LOG_LEVEL=Error in production and not combine LOG_LEVEL=Debug and LOG_TO_HUB=true

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
@@ -66,7 +66,7 @@ namespace LoRaWan
 
         public static void Log(string deviceId, string message, LogLevel logLevel)
         {
-            if ((int)logLevel >= configuration.LogLevel)
+            if ((int)logLevel >= (int)configuration.LogLevel)
             {
                 var msg = message;
 
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
@@ -66,7 +66,7 @@ namespace LoRaWan
 
         public static void Log(string deviceId, string message, LogLevel logLevel)
         {
-            if ((int)logLevel >= (int)configuration.LogLevel)
+            if (logLevel >= configuration.LogLevel)
             {
                 var msg = message;
 
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
@@ -9,26 +9,18 @@ namespace LoRaWan
     using System.Text;
     using System.Threading;
     using Microsoft.Azure.Devices.Client;
+    using Microsoft.Extensions.Logging;
 
     public class Logger
     {
         // Interval where we try to estabilish connection to udp logger
         const int RETRY_UDP_LOG_CONNECTION_INTERVAL_IN_MS = 1000 * 10;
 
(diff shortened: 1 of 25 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerConfiguration.cs
```diff
@@ -16,8 +16,8 @@ public class LoggerConfiguration
         public bool LogToConsole { get; set; } = true;
 
         // Gets/sets the logging level
-        // Default: 0 (Always logging)
-        public int LogLevel { get; set; } = 0;
+        // Default: 4 (Error)
```
fbeltrao: I would change this from int to LogLevel, avoiding casts down the road and making the code cleaner.
The downside is it will break all checks for logLevel > (int)value. To address it I would implement Logger.IsEnabled(LogLevel) to keep a similar behavior with the ILogger interface.

It will enable a future change to a 3rd-party logging library smoothly.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerConfiguration.cs
```diff
@@ -32,5 +32,26 @@ public class LoggerConfiguration
 
         // Gets/sets udp port to send logs
         public int LogToUdpPort { get; set; } = 6000;
+
+        public static int InitLogLevel(string logLevel)
+        {
+            int logLevelInt = 4;
+            logLevel = logLevel.ToUpper();
```
fbeltrao: You can avoid the allocation of the string .ToUpper by comparing strings ignoring case:
`string.Equals(str1, str2, StringComparison. InvariantCultureIgnoreCase)`

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerConfiguration.cs
TechPreacher: Done! No problem with the checks as Microsoft.Extensions.Logging.LogLevel can be cast to int.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerConfiguration.cs
TechPreacher: Done! More elegant code indeed.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
```diff
@@ -66,7 +66,7 @@ public static void Log(string message, LogLevel logLevel)
 
         public static void Log(string deviceId, string message, LogLevel logLevel)
         {
-            if ((int)logLevel >= configuration.LogLevel)
+            if ((int)logLevel >= (int)configuration.LogLevel)
```
fbeltrao: My bad, for some reason I thought the Enum values were not in correct order.
nit: you don't need to cast to int before comparing

TechPreacher: Updated!

PR #106, opened 2019-02-01
Update CI to edge runtime version 1.0.6

fbeltrao:
Update edgeHub and edgeAgent in CI environment from 1.0.6-rc to 1.0.6

Fixes AB#1028


fbeltrao:
Ran CI as test only and succeeded.

Issue #100, opened 2019-01-28
Join failures


gerfen:
Using the code form the current master branch (0.3.0)  I am receiving `Error processing the message The given key '6' was not present in the dictionary. ` for devices from one particular device manufacturer.  I asked the manufacturer to set up the devices with this frequency plan[1].  I'm trying to determine if there is something incorrectly configured on the devices or if there is a bug in the gateway code?  Here is a sampling of the logs[2].

Any insight as to where the issue lies would be much appreciated.

Thanks!

--Michael

[1] US frequency plan
````
 {
    "SX1301_conf": {
             "lorawan_public": true,
        "clksrc": 1, /* radio_1 provides clock to concentrator */
        "antenna_gain": 0, /* antenna gain, in dBi */
        "radio_0": {
            "enable": true,
            "type": "SX1257",
            "freq": 902700000,
            "rssi_offset": -166.0,
            "tx_enable": true,
            "tx_freq_min": 902000000,
            "tx_freq_max": 928000000
        },
        "radio_1": {
            "enable": true,
            "type": "SX1257",
            "freq": 903400000,
            "rssi_offset": -166.0,
            "tx_enable": false
        },
        "chan_multiSF_0": {
            /* Lora MAC channel, 125kHz, all SF, 902.3 MHz */
            "enable": true,
            "radio": 0,
            "if": -400000
        },
        "chan_multiSF_1": {
            /* Lora MAC channel, 125kHz, all SF, 902.5 MHz */
            "enable": true,
            "radio": 0,
            "if": -200000
        },
        "chan_multiSF_2": {
            /* Lora MAC channel, 125kHz, all SF, 902.7 MHz */
            "enable": true,
            "radio": 0,
            "if": 0
        },
        "chan_multiSF_3": {
            /* Lora MAC channel, 125kHz, all SF, 902.9 MHz */
            "enable": true,
            "radio": 0,
            "if": 200000
        },
        "chan_multiSF_4": {
            /* Lora MAC channel, 125kHz, all SF, 903.1 MHz */
            "enable": true,
            "radio": 1,
            "if": -300000
        },
        "chan_multiSF_5": {
            /* Lora MAC channel, 125kHz, all SF, 903.3 MHz */
            "enable": true,
            "radio": 1,
            "if": -100000
        },
        "chan_multiSF_6": {
            /* Lora MAC channel, 125kHz, all SF, 903.5 MHz */
            "enable": true,
            "radio": 1,
[...]

PR #111, opened 2019-02-05
Correct region implementation and add tests

Mandur:
* Correction of #100 
* Add tests on region mapping


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/RegionImplementationTest.cs
```diff
@@ -0,0 +1,215 @@
...
+    using LoRaTools.LoRaPhysical;
+    using LoRaTools.Regions;
+    using Xunit;
+
+    public class RegionImplementationTest
+    {
+        [Theory]
+        [CombinatorialData]
+        public void TestPhysicalMappingEU(
+      [CombinatorialValues("SF12BW125", "SF11BW125", "SF10BW125", "SF9BW125", "SF8BW125", "SF7BW125", "SF7BW250")]string datr,
```
fbeltrao: Good one, I was not aware of the combinatorial library!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
```diff
@@ -509,9 +509,17 @@ async Task<DownlinkPktFwdMessage> ProcessDataMessageAsync(LoRaTools.LoRaPhysical
...
-                tmst = rxpk.Tmst + timeWatcher.GetReceiveWindow1Delay(loRaDevice) * 1000000;
+                try
+                {
+                    datr = this.loraRegion.GetDownstreamDR(rxpk);
+                    freq = this.loraRegion.GetDownstreamChannelFrequency(rxpk);
+                    tmst = rxpk.Tmst + this.loraRegion.Receive_delay1 * 1000000;
+                }
+                catch (Exception ex)
+                {
+                    Logger.Log(loRaDevice.DevEUI, ex.Message, LogLevel.Error);
```
fbeltrao: Wouldn't it make sense to throw a `InvalidDataRateException` inside the Region.GetDownstreamDR() method? Then change the catch to only catch this kind of exception.

Additional, you could add more contextual information when logging:
`Logger.Log(devEUI, $"Error resolving downstream data rate. {ex.Message}", LogLevel.Error);`

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -241,9 +243,15 @@ public string GetDownstreamDR(Rxpk upstreamChannel)
...
-                return configuration;
-                // return this.DRtoConfiguration[10-dr].configuration;
+                if (dr >= 0 && dr < 5)
+                {
+                    var (configuration, maxPyldSize) = dr != 4 ? this.DRtoConfiguration[10 + dr] : this.DRtoConfiguration[13];
+                    return configuration;
+                }
+                else
+                {
+                    throw new Exception("Datarate in {0} region was not within the acceptable range of upstream datarates.");
```
fbeltrao: You forgot to supply the value for {0}. I think $"" are more readable than string.Format("{0}"). Perhaps add the input data rate too so that the exception message will give us all information needed to understand the problem (region, data rate).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageProcessor.cs
Mandur: I am working on this, I created a RegionMappingException to handle DR, but also frequency and other issues related to regions.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
Mandur: Good one, thank you!

Mandur:
Sorry, I forgot to squash the commits before pushing my changes 😢

Mandur:
@fbeltrao comments should now be incuded.
Waiting from US tests before removing the [WIP] tag and ask for a merge. Otherwise the PR is considered as ready.

fbeltrao:
Will approve once CI passes

PR #112, opened 2019-02-06
Change NetId check log to be less verbose

Mandur:
Fix AB#1052

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ProcessLogger.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ProcessLogger.cs
@@ -18,14 +18,12 @@ namespace LoRaWan.NetworkServer
 
         internal ProcessLogger(LoRaOperationTimeWatcher timeWatcher)
         {
-            this.LogLevel = LogLevel.Information;
             this.timeWatcher = timeWatcher;
             this.LogLevel = LogLevel.Information;
         }
 
         internal ProcessLogger(LoRaOperationTimeWatcher timeWatcher, ReadOnlyMemory<byte> devAddr)
         {
-            this.LogLevel = LogLevel.Information;
             this.timeWatcher = timeWatcher;
             this.devAddr = devAddr;
             this.LogLevel = LogLevel.Information;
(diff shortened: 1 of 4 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ProcessLogger.cs
```diff
@@ -18,12 +18,14 @@ internal sealed class ProcessLogger : IDisposable
 
         internal ProcessLogger(LoRaOperationTimeWatcher timeWatcher)
         {
+            this.LogLevel = LogLevel.Information;
```
ronniesa: This is good so we are sure that this variable is set twice :-)

fbeltrao: I think it was messed up by the merge/rebase, since we both added this change.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ProcessLogger.cs
```diff
@@ -18,12 +18,14 @@ internal sealed class ProcessLogger : IDisposable
...
             this.timeWatcher = timeWatcher;
             this.LogLevel = LogLevel.Information;
         }
 
         internal ProcessLogger(LoRaOperationTimeWatcher timeWatcher, ReadOnlyMemory<byte> devAddr)
         {
+            this.LogLevel = LogLevel.Information;
```
fbeltrao: Same as above

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ProcessLogger.cs
Mandur: This code was meant to be quantum resistant, hence the double assignation.

Mandur: (or Francisco is right, you can pick your favorite explanation ;) )

PR #115, opened 2019-02-06
Add support for string values in the sample decoder. We allow the sup…

debugerr:
AB#1063
Adding support for string values in the sample decoder

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
@@ -4,6 +4,7 @@
 namespace LoRaWan.NetworkServer
 {
     using System;
+    using System.Globalization;
     using System.Net.Http;
     using System.Reflection;
     using System.Text;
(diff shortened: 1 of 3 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
```diff
@@ -131,13 +132,17 @@ async Task<string> CallSensorDecoderModule(string sensorDecoderModuleUrl, byte[]
...
-        /// <returns></returns>
+        /// <param name="payload">The payload to decode</param>
+        /// <param name="fport">The received frame port</param>
+        /// <returns>The decoded value as a JSON string</returns>
         public static string DecoderValueSensor(byte[] payload, uint fport)
         {
             var result = Encoding.UTF8.GetString(payload);
-            return $"{{ \"value\":{result} }}";
+
+            var isSupportedNumericValue = double.TryParse(result, NumberStyles.Float, CultureInfo.InvariantCulture, out double ignored);
```
fbeltrao: Out of curiosity: why are you specifying a number style?

LoRaEngine/test/LoRaWanNetworkServer.Test/LoRaPayloadDecoderTest.cs
```diff
@@ -16,6 +16,10 @@ public class LoRaPayloadDecoderTest
         [InlineData("DecoderValueSensor", "1234", 2)]
         [InlineData("DECODERVALUESENSOR", "1234", 1)]
         [InlineData("decodervaluesensor", "1234", 2)]
+        [InlineData("decodervaluesensor", "hello world", 2)]
+        [InlineData("decodervaluesensor", " 1 ", 2)]
+        [InlineData("decodervaluesensor", "$1", 2)]
+        [InlineData("decodervaluesensor", "100'000", 2)]
         public async Task When_Decoder_Is_DecoderValueSensor_Return_In_Value(string decoder, string payloadString, byte fport)
```
fbeltrao: Nice one. Could you please modify the test (including name) ABP_Unconfirmed_With_Value_Decoder_Sends_Decoded_Numeric_Payload to handle other payloads (strings, doubles, etc).

This is a e2e test without external dependencies. It checks that the data we are sending to IoT Hub is correct, after the value has been decoded. If you need help please let me know.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
debugerr: To be specific about what characters are supported in the JSON spec. If you have en-US culture and you pass in 1,643.57 without specifying a NumberStyle it will parse fine, but the JSON parser would throw.

LoRaEngine/test/LoRaWanNetworkServer.Test/LoRaPayloadDecoderTest.cs
debugerr: Sure, I added string payloads to that test as well.

PR #116, opened 2019-02-06
Fix LinkADR Mac Command parsing

Mandur:
* Fix bug in which a LinkADRMacAns would cause an infinite loop AB#1008

NB. A full refactor of the MAC command is underway but this won't be included in 0.4. This PR is only intended a a fix.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRCmd.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRCmd.cs
@@ -13,6 +13,11 @@ namespace LoRaTools
         // private readonly byte dataRateTXPower;
         private readonly byte[] chMask = new byte[2];
 
+        public LinkADRCmd()
+        {
+            this.Length = 2;
+        }
+
         // private readonly byte redondancy;
         public override byte[] ToBytes()
         {
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MacCommandHolder.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MacCommandHolder.cs
@@ -79,6 +79,9 @@ namespace LoRaTools
                         this.MacCommand.Add(linkCheck);
                         break;
                     case CidEnum.LinkADRCmd:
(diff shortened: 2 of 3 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRCmd.cs
```diff
@@ -13,6 +13,11 @@ public class LinkADRCmd : GenericMACCommand
         // private readonly byte dataRateTXPower;
         private readonly byte[] chMask = new byte[2];
 
+        public LinkADRCmd()
```
fbeltrao: Just to understand: the virtual Length will come for next release right?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRCmd.cs
```diff
@@ -13,6 +13,11 @@ public class LinkADRCmd : GenericMACCommand
...
+        {
+            this.Length = 2;
+        }
+
         // private readonly byte redondancy;
         public override byte[] ToBytes()
         {
```
fbeltrao: I would add a `default:` handler. If the cid is not of the type you have cases for this will loop forever. Next question is what to do?
 - throw exception telling that the cid is unknown or not supported?
- skip the loop, log and remain silently?

I prefer the first option.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRCmd.cs
Mandur: Yes

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRCmd.cs
Mandur: This is correct, I added it to the refactor but this would also make sense here. Thanks for the input

Mandur:
@fbeltrao Added changes as discussed

Mandur:
https://dev.azure.com/epicstuff/Azure%20IoT%20Edge%20LoRaWAN%20Starter%20Kit/_build/results?buildId=2132 <-- Manual test result

PR #117, opened 2019-02-06
Creating a Cayenne Decoder including Encoder and tests.

Ellerbach:
Adding support for Cayenne LoRa payload message.
- Created the core Cayenne decoder
- Created a Cayenne encoder
- Created all the devices classes
- Created unit test for the encoder/decoder

### DIFF
--- a/Samples/CayenneDecoder/Classes/CayenneDecoder.cs
+++ b/Samples/CayenneDecoder/Classes/CayenneDecoder.cs
@@ -1,0 +1,227 @@
+﻿using Microsoft.AspNetCore.Mvc.RazorPages;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace CayenneDecoderModule.Classes
+{
+    public class CayenneDecoder
+    {
+
+        byte[] buffer;
+
+        public CayenneDecoder(byte[] payload)
+        {
+            buffer = payload;
+        }
+
+        public bool TryGetDigitalInput(out DigitalInput digitalInput)
+        {
+            var ret = GetOutValues(CayenneTypes.DigitalInput);
+            if (ret != null)
+            {
+                digitalInput = new DigitalInput() { Channel = ret[0], Value = ret[1] };
+                return true;
+            }
+            digitalInput = null;
+            return false;
+        }
+
+        public bool TryGetDigitalOutput(out DigitalOutput digitalOutput)
+        {
+            var ret = GetOutValues(CayenneTypes.DigitalOutput);
+            if (ret != null)
+            {
+                digitalOutput = new DigitalOutput() { Channel = ret[0], Value = ret[1] };
+                return true;
+            }
+            digitalOutput = null;
+            return false;
+        }
+
+        public bool TryGetAnalogInput(out AnalogInput analogInput)
+        {
+            var ret = GetOutValues(CayenneTypes.AnalogInput);
+            if (ret != null)
+            {
+                analogInput = new AnalogInput() { Channel = ret[0], Value = ((Int16)(ret[1] << 8) + ret[2]) / 100.0 };
+                return true;
+            }
+            analogInput = null;
+            return false;
+        }
+
+        public bool TryGetAnalogOutput(out AnalogOutput analogOutput)
+        {
+            var ret = GetOutValues(CayenneTypes.AnalogOutput);
+            if (ret != null)
+            {
+                analogOutput = new AnalogOutput() { Channel = ret[0], Value = ((Int16)(ret[1] << 8) + ret[2]) / 100.0 };
+                return true;
+            }
+            analogOutput = null;
+            return false;
+        }
+
+        public bool TryGetIlluminanceSensor(out IlluminanceSensor illuminanceSensor)
+        {
+            var ret = GetOutValues(CayenneTypes.Luminisity);
+            if (ret != null)
+            {
+                illuminanceSensor = new IlluminanceSensor() { Channel = ret[0], Value = (UInt16)((ret[1] << 8) + ret[2]) };
+                return true;
+            }
+            illuminanceSensor = null;
+            return false;
+        }
+
+        public bool TryGetPresenceSensor(out PresenceSensor presenceSensor)
+        {
+            var ret = GetOutValues(CayenneTypes.Presence);
+            if (ret != null)
+            {
+                presenceSensor = new PresenceSensor() { Channel = ret[0], Value = ret[1] };
+                return true;
+            }
+            presenceSensor = null;
+            return false;
+        }
+
+        public bool TryGetTemperatureSensor(out TemperatureSensor temperatureSensor)
+        {
+            var ret = GetOutValues(CayenneTypes.Temperature);
+            if (ret != null)
+            {
+                temperatureSensor = new TemperatureSensor() { Channel = ret[0], Value = ((Int16)(ret[1] << 8) + ret[2]) / 10.0 };
+                return true;
+            }
+            temperatureSensor = null;
+            return false;
+        }
+
+        public bool TryGetHumiditySensor(out HumiditySensor humiditySensor)
+        {
+            var ret = GetOutValues(CayenneTypes.RelativeHumidity);
+            if (ret != null)
+            {
+                humiditySensor = new HumiditySensor() { Channel = ret[0], Value = ret[1] / 2.0 };
+                return true;
+            }
+            humiditySensor = null;
+            return false;
+        }
+
+        public bool TryGetAccelerator(out Accelerator accelerator)
+        {
+            var ret = GetOutValues(CayenneTypes.Accelerator);
+            if (ret != null)
+            {
+                accelerator = new Accelerator()
+                {
+                    Channel = ret[0],
+                    X = ((Int16)(ret[1] << 8) + ret[2]) / 1000.0,
+                    Y = ((Int16)(ret[3] << 8) + ret[4]) / 1000.0,
+                    Z = ((Int16)(ret[5] << 8) + ret[6]) / 1000.0
+                };
+                return true;
+            }
+            accelerator = null;
+            return false;
+        }
+
+        public bool TryGetBarometer(out Barometer barometer)
+        {
+            var ret = GetOutValues(CayenneTypes.Barometer);
+            if (ret != null)
+            {
+                barometer = new Barometer() { Channel = ret[0], Value = ((ret[1] << 8) + ret[2]) / 10.0 };
+                return true;
+            }
+            barometer = null;
+            return false;
+        }
+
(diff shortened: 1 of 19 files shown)


Samples/CayenneDecoder/Classes/CayenneDecoder.cs
```diff
@@ -0,0 +1,227 @@
...
+                        outArray = new byte[size - 1];
+                        outArray[0] = channel;
+                        Array.Copy(buffer, cursor, outArray, 1, outArray.Length - 1);
+                        return outArray;
+                    }
+                    cursor += size - 2;
+                }
+
+            }
+            catch (Exception)
```
Mandur: I am not sure this is a good way of tackling exception

Samples/CayenneDecoder/Classes/CayenneTypes.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace CayenneDecoderModule.Classes
+{
+    public enum CayenneTypes
+    {
+        DigitalInput = 0,       // 1 byte
+        DigitalOutput = 1,      // 1 byte
+        AnalogInput = 2,      // 2 bytes, 0.01 signed
+        AnalogOutput = 3,      // 2 bytes, 0.01 signed
+        Luminisity = 101,    // 2 bytes, 1 lux unsigned
```
Mandur: Typo

Samples/CayenneDecoder/Classes/CayenneDecoder.cs
Ellerbach: it's just to make sure I catch Exceptions. In this case, I jsut return a null byte array (after the loop). It's an efficient way to trap them. The function is then call by higher level functions which are TryGet pattent

Samples/CayenneDecoder/Classes/CayenneTypes.cs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public enum CayenneTypeSize
+    {
+        // Data ID + Data Type + Data Size
+        DigitalInput = 3,      // 1 byte
+        DigitalOutput = 3,    // 1 byte
+        AnalogInput = 4,     // 2 bytes, 0.01 signed
+        AnalogOutput = 4,    // 2 bytes, 0.01 signed
+        Luminisity = 4,     // 2 bytes, 1 lux unsigned
```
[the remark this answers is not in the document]
fbeltrao: I think we are miss-using enums for this. Here `CayenneTypeSize.Accelerator == CayenneTypeSize. Gyrometer`

Samples/CayenneDecoder/Classes/CayenneDecoder.cs
```diff
@@ -0,0 +1,227 @@
...
+                    sign = -1.0;
+                }
+                gPSLocation.Altitude = sign * ((ret[7] << 16) + (ret[8] << 8) + ret[9]) / 100.0;
+                return true;
+            }
+            gPSLocation = null;
+            return false;
+        }
+
+        private byte[] GetOutValues(CayenneTypes cayenneTypes)
```
fbeltrao: Could return a ReadOnlySpan<byte> here, not need to make a copy of byte[]

Samples/CayenneDecoder/Classes/LoraDecoders.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace CayenneDecoderModule.Classes
+{
+    internal static class LoraDecoders
+    {
+        private static string CayenneDecoder(byte[] payload, uint fport)
+        {
+            CayenneDecoder cayenneDecoder = new CayenneDecoder(payload);
+            CayenneDevice cayenneDevice = new CayenneDevice();
+
+            if (cayenneDecoder.TryGetAnalogInput(out AnalogInput analogInput))
```
fbeltrao: Is the `next byte` to read telling us the type of the payload? If that is the case can't we make it faster and create the correct type instead of trying all types?

Samples/CayenneDecoderTest/CayenneTest.cs
```diff
@@ -0,0 +1,189 @@
...
+        public void Temeperature()
+        {
+            CayenneEncoder cayenneEncoder = new CayenneEncoder();
+            double temp = 33.7;
+            byte channel = 1;
+            cayenneEncoder.AddTemperature(channel, temp);
+            var buff = cayenneEncoder.GetBuffer();
+            CayenneDecoder cayenneDecoder = new CayenneDecoder(buff);            
+            cayenneDecoder.TryGetTemperatureSensor(out TemperatureSensor temperatureSensor);
+            Assert.True(temperatureSensor.Value == temp);
```
fbeltrao: For comparison use Assert.Equal(temp, temperatureSensor.Value);

Samples/CayenneDecoder/Classes/LoraDecoders.cs
Ellerbach: if there was only 1 devices to decode, then yes. But most payload contains multiple devices. And they is no specific order to put them one after the other. And as the size is different every time, then you can't use this method. Yes, if there were thousands of devices, there maybe few other ways to optimize and cache some decoded information. But in this case, there are max like 4 or 5. So it goes actually largely fast enough.

Samples/CayenneDecoderTest/CayenneTest.cs
Ellerbach: ok, will correct. thanks

Samples/CayenneDecoder/Classes/CayenneDecoder.cs
Ellerbach: I'm not keeping the original version of the array. I'm removing data in the original one and extract part of it. So I have to make a copy of it.

Samples/CayenneDecoderTest/CayenneTest.cs
```diff
@@ -0,0 +1,189 @@
+using System;
```
fbeltrao: Love the tests. I am missing a e2e test that will verify that (string payload, byte fport) => returns a correct json that will be picked by the network server.

Samples/CayenneDecoder/Classes/LoraDecoders.cs
fbeltrao: Just trying to see if I understand the logic. Would the snippet below work to decode multiple values?

```c#
byte[] input = GetInputFromSomewhere();

var index = 0;
while (index < input.Length)
{
   switch (input[index])
   {
      case (int)CayenneTypes.DigitalInput:
          // create Digital input starting from input[index]
          index += (int)CayenneTypeSize.DigitalInput;
          break;

     // add here all other types && default handling to stop
   }   
}

```

Ellerbach: The problem with the way the packet is formed is that the it is a sum of chuncks which have various sizes. And the first element of the table is unfortunately not the type of input. It is the second element. This second element gives you the type of sensor which gives you the full size (including the first element and the type of sensor) of the chunck. 
A code like you propose can work but is much longer than the one with the matching enums and extracting the real data. 
Keep in mind that once you have the real data, they all need to be transformed differently.
The way I can kind of improve the performance if we really want Something with more performance if to create a list of devices and parse only once the full buffer.

fbeltrao: Thanks for the explanation. I am not worried about performance, I am thinking about a cleaner way to accomplish it without the "ifs".

Samples/CayenneDecoder/Classes/CayenneDecoder.cs
```diff
@@ -10,218 +10,149 @@ public class CayenneDecoder
...
-                    var size = (byte)Enum.Parse<CayenneTypeSize>(Enum.GetName(typeof(CayenneTypes), type));
-                    if (type == (byte)cayenneTypes)
+                    while (cursor != buffer.Length)
                     {
-                        // found what we want to decode
-                        outArray = new byte[size - 1];
-                        outArray[0] = channel;
-                        Array.Copy(buffer, cursor, outArray, 1, outArray.Length - 1);
-                        return outArray;
+                        channel = buffer[cursor++];
```
fbeltrao: Suggest: you could move the logic of creating the concrete Cayenne type to each constructor, so this method is cleaner:
```c#
new DigitalInput(channel, value.Span(cursor, size));
or
new Accelerator(channel, value.Span(cursor, size));
```

And have constructors like:
```c#
public Accelerator(byte channel, ReadOnlySpan<byte> buffer)
{
    this.X = ((Int16)(buffer[0] << 8) + buffer[1]) / 1000.0;
    this.Y = ((Int16)(buffer[2] << 8) + buffer[3]) / 1000.0;
    this.Z = ((Int16)(buffer[4] << 8) + buffer[5]) / 1000.0
}
```

Samples/CayenneDecoderTest/CayenneTest.cs
```diff
@@ -284,11 +280,10 @@ public void GPSLocation()
...
-            Assert.Equal(gPSLocation.Longitude, lon);
-            Assert.Equal(gPSLocation.Altitude, alt);
-            Assert.Equal(gPSLocation.Channel, channelgps);
+            Assert.Equal(cayenneDecoder.cayenneDevice.GPSLocation[0].Latitude, lat);
+            Assert.Equal(cayenneDecoder.cayenneDevice.GPSLocation[0].Longitude, lon);
+            Assert.Equal(cayenneDecoder.cayenneDevice.GPSLocation[0].Altitude, alt);
+            Assert.Equal(cayenneDecoder.cayenneDevice.GPSLocation[0].Channel, channelgps);
         }
     }
 }
```
fbeltrao: Could you please add a e2e test with multiple values of same type?

Samples/CayenneDecoder/Classes/CayenneDecoder.cs
fbeltrao: You could even get rid of the CayenneTypeSize, by having each type have a property with the Byte length.
In each constructor you check if the span has the minimum expected size, if not throw an exception.

Samples/CayenneDecoderTest/CayenneTest.cs
Ellerbach: I can try to before end of the week.

Samples/CayenneDecoder/Classes/CayenneDecoder.cs
Ellerbach: I think that in terms of architecture, if we want or need at some point to use the encoder class as well to generate the byte arays, I would leave the conversion part of the converter class and have the device objects as light as possible. I was hesitating to do it like you proposed and finally decided to leave all in the conversion class.
Having an enum class for the size makes it easier to avoid mistake with "magic numbers". because the way to get rid of them is to hard code them in the switch section.

PR #118, opened 2019-02-07
Changes to deployment templates for 0-4-0-preview release

TechPreacher:
- Removed deployment.dev.template.json
- Removed deployment.sim.template.json
- Hard code edgeAgent and edgeHub versions and environment variables in deployment template
- Set default logging to Error in deployment template
- Adapt /Template/deviceConfiguration.json


Template/deviceConfiguration.json
```diff
@@ -37,7 +46,7 @@
 					"LoRaWanNetworkSrvModule": {
 						"type": "docker",
 						"settings": {
-							"image": "loraedge/lorawannetworksrvmodule:0.3.0-preview-arm32v7",
+							"image": "loraedge/lorawannetworksrvmodule:0.4.0-preview-arm32v7",
 							"createOptions": "{\"ExposedPorts\": { \"1680/udp\": {}}, \"HostConfig\": {  \"PortBindings\": {\"1680/udp\": [  { \"HostPort\": \"1680\", \"HostIp\":\"172.17.0.1\" } ]}}}"
```
ronniesa: @Mandur these one should be agnostic right?

Mandur: Yes correct  0.4.0-preview should be name for both

TechPreacher: Updated!

fbeltrao: Same applies to pktfwd module

Mandur: Yes, both images are now ready

README.md
```diff
@@ -373,6 +373,34 @@ CidType : 6
...
+
+```
+mqttSettings__enabled: false
+httpSettings__enabled: false
+TwinManagerVersion: v2
+```
+
+You do this by clicking "Set Modules" &rarr; "Configure advanced edge runtime settings" on your IoT Edge device in Azure IoT Hub.
+
+Make sure the **DevAddr** of your ABP LoRa devices starts with **"02"**: Due to addition of NetId support in this pre-relese, ABP devices created by the template  prior to 0.4.0-preview (and all devices with an incompatible NetId in general) will be incompatible with the 0.4.0-preview. If you don't set a specific NetId, the default is 1. In this case, make sure the DevAddr of your ABP LoRa devices starts with "02".
```
ronniesa: NetId is not user defined so it is currently fix to 1

TechPreacher: Changed.

Mandur: I would remove sentence "If you don't set a specific NetId, the default is 1.". Is this ok then?

PR #124, opened 2019-02-11
Introduce request queue for data payloads

fbeltrao:
**Changes**
- Deprecated MessageProcessor. It will be removed in a future PR.
- Add task queueing to LoRaDevice class
- Moved message processing logic to DefaultLoRaDataRequestHandler
- Introduce MessageDispatcher, which internally uses the LoRaDevice queue to process data messages
- Added DeviceLoaderSynchronizer which takes care of loading devices by devAddr while queueing incoming message requests. Ensuring first-in first-out to incoming data payloads.
- Add IPacketForwarder interface that defines a contract to send data to a packet forwarder.
- UdpServer now implements IPacketForwarder, keeping also the IP address of incoming calls (in addition to udp port).
- UdpServer is using message dispatcher through a compiler directive

Fixes AB#946

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
@@ -35,10 +35,12 @@ namespace LoRaWan.NetworkServer
         {
             var timeWatcher = new LoRaOperationTimeWatcher(request.LoRaRegion, request.StartTime);
             var loraPayload = (LoRaPayloadData)request.Payload;
-            var isMultiGateway = !string.Equals(loRaDevice.GatewayID, this.configuration.GatewayID, StringComparison.InvariantCultureIgnoreCase);
-            var frameCounterStrategy = isMultiGateway ?
-                this.frameCounterUpdateStrategyProvider.GetMultiGatewayStrategy() :
-                this.frameCounterUpdateStrategyProvider.GetSingleGatewayStrategy();
+            var frameCounterStrategy = this.frameCounterUpdateStrategyProvider.GetStrategy(loRaDevice.GatewayID);
+            if (frameCounterStrategy == null)
+            {
+                Logger.Log(loRaDevice.DevEUI, $"failed to resolve frame count update strategy, device gateway: {loRaDevice.GatewayID}, message ignored", LogLevel.Error);
+                return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.ApplicationError);
+            }
 
             var payloadFcnt = loraPayload.GetFcnt();
             var requiresConfirmation = loraPayload.IsConfirmed();
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
@@ -1,0 +1,505 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.NetworkServer
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+    using System.Threading.Tasks;
+    using LoRaTools;
+    using LoRaTools.LoRaMessage;
+    using LoRaTools.LoRaPhysical;
+    using LoRaTools.Utils;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
+    {
+        private readonly NetworkServerConfiguration configuration;
+        private readonly ILoRaDeviceFrameCounterUpdateStrategyFactory frameCounterUpdateStrategyFactory;
+        private readonly ILoRaPayloadDecoder payloadDecoder;
+
+        public DefaultLoRaDataRequestHandler(
+            NetworkServerConfiguration configuration,
+            ILoRaDeviceFrameCounterUpdateStrategyFactory frameCounterUpdateStrategyFactory,
+            ILoRaPayloadDecoder payloadDecoder)
+        {
+            this.configuration = configuration;
+            this.frameCounterUpdateStrategyFactory = frameCounterUpdateStrategyFactory;
+            this.payloadDecoder = payloadDecoder;
+        }
+
+        public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaRequest request, LoRaDevice loRaDevice)
+        {
+            var timeWatcher = new LoRaOperationTimeWatcher(request.LoRaRegion, request.StartTime);
+            var loraPayload = (LoRaPayloadData)request.Payload;
+            var isMultiGateway = !string.Equals(loRaDevice.GatewayID, this.configuration.GatewayID, StringComparison.InvariantCultureIgnoreCase);
+            var frameCounterStrategy = isMultiGateway ?
+                this.frameCounterUpdateStrategyFactory.GetMultiGatewayStrategy() :
+                this.frameCounterUpdateStrategyFactory.GetSingleGatewayStrategy();
+
+            var payloadFcnt = loraPayload.GetFcnt();
+            var requiresConfirmation = loraPayload.IsConfirmed();
+
+            using (new LoRaDeviceFrameCounterSession(loRaDevice, frameCounterStrategy))
+            {
+                // Leaf devices that restart lose the counter. In relax mode we accept the incoming frame counter
+                // ABP device does not reset the Fcnt so in relax mode we should reset for 0 (LMIC based) or 1
+                var isFrameCounterFromNewlyStartedDevice = false;
+                if (payloadFcnt <= 1)
+                {
+                    if (loRaDevice.IsABP)
+                    {
+                        if (loRaDevice.IsABPRelaxedFrameCounter && loRaDevice.FCntUp >= 0 && payloadFcnt <= 1)
+                        {
+                            // known problem when device restarts, starts fcnt from zero
+                            _ = frameCounterStrategy.ResetAsync(loRaDevice);
+                            isFrameCounterFromNewlyStartedDevice = true;
+                        }
+                    }
+                    else if (loRaDevice.FCntUp == payloadFcnt && payloadFcnt == 0)
+                    {
+                        // Some devices start with frame count 0
+                        isFrameCounterFromNewlyStartedDevice = true;
+                    }
+                }
+
+                // Reply attack or confirmed reply
+                // Confirmed resubmit: A confirmed message that was received previously but we did not answer in time
+                // Device will send it again and we just need to return an ack (but also check for C2D to send it over)
+                var isConfirmedResubmit = false;
+                if (!isFrameCounterFromNewlyStartedDevice && payloadFcnt <= loRaDevice.FCntUp)
+                {
+                    // if it is confirmed most probably we did not ack in time before or device lost the ack packet so we should continue but not send the msg to iothub
+                    if (requiresConfirmation && payloadFcnt == loRaDevice.FCntUp)
+                    {
+                        if (!loRaDevice.ValidateConfirmResubmit(payloadFcnt))
+                        {
+                            Logger.Log(loRaDevice.DevEUI, $"resubmit from confirmed message exceeds threshold of {LoRaDevice.MaxConfirmationResubmitCount}, message ignored, msg: {payloadFcnt} server: {loRaDevice.FCntUp}", LogLevel.Debug);
+                            return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.ConfirmationResubmitThresholdExceeded);
+                        }
+
+                        isConfirmedResubmit = true;
+                        Logger.Log(loRaDevice.DevEUI, $"resubmit from confirmed message detected, msg: {payloadFcnt} server: {loRaDevice.FCntUp}", LogLevel.Information);
+                    }
+                    else
+                    {
+                        Logger.Log(loRaDevice.DevEUI, $"invalid frame counter, message ignored, msg: {payloadFcnt} server: {loRaDevice.FCntUp}", LogLevel.Information);
+                        return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.InvalidFrameCounter);
+                    }
+                }
+
+                var fcntDown = 0;
+                // If it is confirmed it require us to update the frame counter down
+                // Multiple gateways: in redis, otherwise in device twin
+                if (requiresConfirmation)
+                {
+                    fcntDown = await frameCounterStrategy.NextFcntDown(loRaDevice, payloadFcnt);
+
+                    // Failed to update the fcnt down
+                    // In multi gateway scenarios it means the another gateway was faster than using, can stop now
+                    if (fcntDown <= 0)
+                    {
+                        // update our fcntup anyway?
+                        // loRaDevice.SetFcntUp(payloadFcnt);
+                        Logger.Log(loRaDevice.DevEUI, "another gateway has already sent ack or downlink msg", LogLevel.Information);
+
+                        return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.HandledByAnotherGateway);
+                    }
+
+                    Logger.Log(loRaDevice.DevEUI, $"down frame counter: {loRaDevice.FCntDown}", LogLevel.Information);
+                }
+
+                if (!isConfirmedResubmit)
+                {
+                    var validFcntUp = isFrameCounterFromNewlyStartedDevice || (payloadFcnt > loRaDevice.FCntUp);
+                    if (validFcntUp)
+                    {
+                        Logger.Log(loRaDevice.DevEUI, $"valid frame counter, msg: {payloadFcnt} server: {loRaDevice.FCntUp}", LogLevel.Information);
+
+                        object payloadData = null;
+
+                        // if it is an upward acknowledgement from the device it does not have a payload
+                        // This is confirmation from leaf device that he received a C2D confirmed
+                        // if a message payload is null we don't try to decrypt it.
+                        if (loraPayload.Frmpayload.Length != 0)
+                        {
+                            byte[] decryptedPayloadData = null;
+                            try
+                            {
+                                decryptedPayloadData = loraPayload.GetDecryptedPayload(loRaDevice.AppSKey);
+                            }
+                            catch (Exception ex)
+                            {
+                                Logger.Log(loRaDevice.DevEUI, $"failed to decrypt message: {ex.Message}", LogLevel.Error);
+                            }
+
+                            var fportUp = loraPayload.GetFPort();
+
+                            if (string.IsNullOrEmpty(loRaDevice.SensorDecoder))
+                            {
+                                Logger.Log(loRaDevice.DevEUI, $"no decoder set in device twin. port: {fportUp}", LogLevel.Debug);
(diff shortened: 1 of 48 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -0,0 +1,505 @@
...
+            // if (firstWindowTime > TimeSpan.Zero)
+            //     System.Threading.Thread.Sleep(firstWindowTime);
+            var receiveWindow = timeWatcher.ResolveReceiveWindowToUse(loRaDevice);
+            if (receiveWindow == 0)
+                return null;
+
+            string datr;
+            double freq;
+            long tmst;
+            if (receiveWindow == 2)
```
debugerr: would be cool to have that magical number 2 defined as a constant that can give some context to its meaning.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -0,0 +1,505 @@
...
+                    }
+                }
+            }
+
+            if (await loRaDevice.SendEventAsync(deviceTelemetry, eventProperties))
+            {
+                var payloadAsRaw = deviceTelemetry.Data as string;
+                if (payloadAsRaw == null && deviceTelemetry.Data != null)
+                {
+                    payloadAsRaw = JsonConvert.SerializeObject(deviceTelemetry.Data, Formatting.None);
```
debugerr: You may want to wrap this in an exception handler?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -0,0 +1,330 @@
...
+            NetworkServerConfiguration configuration,
+            Action<Task> continuationAction)
+        {
+            this.loRaDeviceAPIService = loRaDeviceAPIService;
+            this.deviceFactory = deviceFactory;
+            this.devAddr = devAddr;
+            this.destinationDictionary = destinationDictionary;
+            this.initializers = initializers;
+            this.configuration = configuration;
+            this.state = LoaderState.QueryingDevices;
```
debugerr: that's the default.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -0,0 +1,330 @@
...
+            this.loadingDevicesFailed = false;
```
debugerr: Also here, that would be the default

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            try
+            {
+                // If device was not found, search in the device API, updating local cache
+                Logger.Log(this.devAddr, "querying the registry for device", LogLevel.Information);
+
+                SearchDevicesResult searchDeviceResult = null;
+                try
+                {
+                    searchDeviceResult = await this.loRaDeviceAPIService.SearchByDevAddrAsync(this.devAddr);
```
debugerr: Would be cool if you could explain to me, why a device can be found multiple times with this? I thought the devAddr is unique.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -0,0 +1,330 @@
...
+                device.Queue(request);
+            }
+            else
+            {
+                Logger.Log(device.DevEUI, $"device is not our device, ignore message", LogLevel.Debug);
+                request.NotifyFailed(device, LoRaDeviceRequestFailedReason.BelongsToAnotherGateway);
+            }
+        }
+
+        void SetState(LoaderState newState)
```
debugerr: why not use a property here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -0,0 +1,330 @@
...
+
+        private async Task<LoRaDevice> InitializeDeviceAsync(LoRaDevice loRaDevice)
+        {
+            try
+            {
+                // Calling initialize async here to avoid making async calls in the concurrent dictionary
+                // Since only one device will be added, we guarantee that initialization only happens once
+                if (await loRaDevice.InitializeAsync())
+                {
+                    loRaDevice.IsOurDevice = string.IsNullOrEmpty(loRaDevice.GatewayID) || string.Equals(loRaDevice.GatewayID, this.configuration.GatewayID, StringComparison.InvariantCultureIgnoreCase);
```
debugerr: can be confusing for a newbie like me if IsOurDevice has multiple meanings. If there is no gateway id, it's not really 'our' device right? I thought it was our device if the gatewayid was explicitly set and matches.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -421,6 +431,78 @@ internal async Task<bool> UpdateAfterJoinAsync(string devAddr, string nwkSKey, s
             return succeeded;
         }
 
+        internal void SetRequestHandler(ILoRaDataRequestHandler dataRequestHandler) => this.dataRequestHandler = dataRequestHandler;
```
debugerr: property?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -421,6 +431,78 @@ internal async Task<bool> UpdateAfterJoinAsync(string devAddr, string nwkSKey, s
...
+                    this.StartNextRequest(request);
+                }
+                else
+                {
+                    this.queuedRequests.Enqueue(request);
+                }
+            }
+        }
+
+        private void QueueNext()
```
debugerr: The name is confusing to me. You are not queueing anything, but dequeuing and processing. Maybe ProcessNext would be an option?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -421,6 +431,78 @@ internal async Task<bool> UpdateAfterJoinAsync(string devAddr, string nwkSKey, s
...
+        void StartNextRequest(LoRaRequest requestToStart)
+        {
+            async Task RunAndQueueNext(LoRaRequest request)
+            {
+                LoRaDeviceRequestProcessResult result = null;
+                Exception processingError = null;
+
+                try
+                {
+                    result = await this.dataRequestHandler.ProcessRequestAsync(request, this);
```
debugerr: Here you are still holding onto the lock and pass control to an external method you don't control. If the ProcessRequestAsync would acquire another lock you can end up in a dead lock scenario. It's only a theoretical problem here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -252,6 +252,27 @@ public async Task<bool> AbandonAsync(Message message)
...
+        /// Disconnects device client
+        /// </summary>
+        public async Task<bool> DisconnectAsync()
+        {
+            try
+            {
+                await this.deviceClient.CloseAsync();
+                this.deviceClient = null;
+
+                Logger.Log(this.devEUI, $"device client disconnected", LogLevel.Debug);
```
debugerr: no need to prefix with $ here

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
```diff
@@ -26,6 +28,9 @@ public LoRaDevice Create(IoTHubDeviceInfo deviceInfo)
                 devEUI: deviceInfo.DevEUI,
                 loRaDeviceClient: loraDeviceClient);
 
+            if (this.dataRequestHandler != null)
```
debugerr: do you really need to check for null?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
```diff
@@ -0,0 +1,315 @@
...
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Message dispatcher
+    /// </summary>
+    public class MessageDispatcher
+    {
+        // Defines Cloud to device message property containing fport value
+        internal const string FPORT_MSG_PROPERTY_KEY = "fport";
```
debugerr: arn't those the same as in Constants.cs?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -0,0 +1,505 @@
...
+            this.configuration = configuration;
+            this.frameCounterUpdateStrategyFactory = frameCounterUpdateStrategyFactory;
+            this.payloadDecoder = payloadDecoder;
+        }
+
+        public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaRequest request, LoRaDevice loRaDevice)
+        {
+            var timeWatcher = new LoRaOperationTimeWatcher(request.LoRaRegion, request.StartTime);
+            var loraPayload = (LoRaPayloadData)request.Payload;
+            var isMultiGateway = !string.Equals(loRaDevice.GatewayID, this.configuration.GatewayID, StringComparison.InvariantCultureIgnoreCase);
```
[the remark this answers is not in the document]
fbeltrao: Yes, that would be enough. Since you already made the change on your branch I decided to leave as is.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -0,0 +1,505 @@
...
+            var frameCounterStrategy = isMultiGateway ?
```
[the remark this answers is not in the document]
fbeltrao: That is correct. I started with a factory but then realize I did not have to create a new instance every time. I will change from factory to provider (ILoRaDeviceFrameCounterStrategyProvider)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -0,0 +1,505 @@
...
+                // ABP device does not reset the Fcnt so in relax mode we should reset for 0 (LMIC based) or 1
+                var isFrameCounterFromNewlyStartedDevice = false;
+                if (payloadFcnt <= 1)
+                {
+                    if (loRaDevice.IsABP)
+                    {
+                        if (loRaDevice.IsABPRelaxedFrameCounter && loRaDevice.FCntUp >= 0 && payloadFcnt <= 1)
+                        {
+                            // known problem when device restarts, starts fcnt from zero
+                            _ = frameCounterStrategy.ResetAsync(loRaDevice);
```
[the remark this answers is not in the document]
fbeltrao: If you don't discard there will be warnings

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
fbeltrao: I will add RECEIVE_WINDOW_1 and RECEIVE_WINDOW_2 as constants.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
fbeltrao: There is no getter. Made the method to enable easier unit testing

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
fbeltrao: I prefer having the StartNextRequest doing the 2 lines (assign and schedule current task) than duplicating the code in the 2 places where it is called.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
fbeltrao: True, good catch

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
fbeltrao: Using ABP the user defines the DevAddr, thus there is no guarantee of uniqueness. With OTAA we generate the DevAddr, and there is a very very small chance of conflict.

To identify the correct device we need a combination of DevAddr + checking the payload  integrity (MIC) against the device network session key.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
fbeltrao: Personal preference. If I am doing a bit of code in the setter (logging and validation) I prefer methods.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
fbeltrao: `IsOurDevice` here means whether or not we should handle the device messages. That is true if the device.GatewayID matches ours or the device does not have a gatewayID (in that case all gateways will process it's messages).

Could be changed to IsExternalDevice that would mean device.GatewayID != ourGatewayID. What do you think?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -0,0 +1,505 @@
...
+                // Confirm receiving message to device
+                fctrl = (byte)FctrlEnum.Ack;
+            }
+
+            byte? fport = null;
+            var requiresDeviceAcknowlegement = false;
+            byte[] macbytes = null;
+
+            byte[] rndToken = new byte[2];
+            Random rnd = new Random();
```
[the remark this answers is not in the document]
fbeltrao: Using the same approach as the packet forwarder:
https://github.com/Lora-net/packet_forwarder/blob/d0226eae6e7b6bbaec6117d0d2372bf17819c438/lora_pkt_fwd/src/lora_pkt_fwd.c#L2068

I am not sure what are the chances of TickCount returning the same value if called exactly at the same time.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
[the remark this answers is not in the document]
fbeltrao: I am pushing another change to make a bit cleaner.
```c#
    public interface ILoRaDeviceFrameCounterUpdateStrategyProvider
    {
        ILoRaDeviceFrameCounterUpdateStrategy GetStrategy(string deviceGatewayID);
    }
```

Finding out the correct strategy is responsibility of the provider now.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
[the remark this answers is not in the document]
debugerr: Sure. A good way to ensure unique values would also be to keep the instance around (as a static) and re-use it to generate the random values.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
debugerr: I thought there was a stylecop rule that would flag set/get naming. No big deal.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
debugerr: IsExternalDevice would make it clearer to me. A comment on the property would also help. :)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
debugerr: True

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
debugerr: Sure. Another Q: Why did you make RunAndQueueNext a local function? Just readability?

fbeltrao: Wanted to ensure that is it only called within the method

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
ronniesa: IsOurDevice means that it belongs to our IoT Hub instance (basically is in the DB behind IoT Hub

fbeltrao: With the changes introduced in 0.3-preview this flag indicates if the messages of a device should  be handled by the network module.

A device in the join process or a device that is assigned to a different gateway will have IsOurDevice to `false`. All others will have `true`

ronniesa: Then we should rename it to something more meaningful IsDeviceLinkedToThisGateway

PR #125, opened 2019-02-13
Update Dockerfile.amd64

TechPreacher:
- Create 2 separate builder environments for spidev1 and spidev2 to prevent failure caused by building the packet forwarder twice in a row.

Fixes: AB#1077


TechPreacher:
@fbeltrao Should I add the fullci label for this? There won't be much benefit as we are not testing the amd64 packet forwarder in the CI...

ronniesa:
> @fbeltrao Should I add the fullci label for this? There won't be much benefit as we are not testing the amd64 packet forwarder in the CI...

non need of full ci

PR #126, opened 2019-02-13
Pat dev 932 deduplication

debugerr:
AB#932
AB#1065
Add message deduplication support

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
@@ -1,0 +1,170 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
+{
+    using System;
+    using Microsoft.Azure.WebJobs;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+    using StackExchange.Redis;
+
+    public sealed class LoRaDeviceCache : IDisposable
+    {
+        private const string CacheKeyLockSuffix = "msglock";
+        private static readonly TimeSpan LockWaitingTimeout = TimeSpan.FromSeconds(10);
+        private static IDatabase redisCache;
+
+        private readonly string gatewayId;
+        private readonly string devEUI;
+
+        public bool IsLockOwner { get; private set; }
+
+        private string lockKey;
+
+        private LoRaDeviceCache(string devEUI, string gatewayId)
+        {
+            this.devEUI = devEUI;
+            this.gatewayId = gatewayId;
+        }
+
+        public static LoRaDeviceCache Create(ExecutionContext context, string devEUI, string gatewayId)
+        {
+            if (string.IsNullOrEmpty(devEUI))
+            {
+                throw new ArgumentNullException("devEUI");
+            }
+
+            if (string.IsNullOrEmpty(gatewayId))
+            {
+                throw new ArgumentNullException("gatewayId");
+            }
+
+            EnsureRedisInstance(context);
+            return new LoRaDeviceCache(devEUI, gatewayId);
+        }
+
+        public bool TryToLock()
+        {
+            if (this.IsLockOwner)
+            {
+                return true;
+            }
+
+            this.lockKey = this.devEUI + CacheKeyLockSuffix;
+            if (!redisCache.LockTake(this.lockKey, this.gatewayId, LockWaitingTimeout))
+            {
+                return false;
+            }
+
+            this.IsLockOwner = true;
+            return true;
+        }
+
+        public DeviceCacheInfo Initialize(int clientFCntDown = 0, int clientFCntUp = 0)
+        {
+            // it is the first message from this device
+            var newFCntDown = clientFCntDown + 1;
+            var serverStateForDeviceInfo = new DeviceCacheInfo
+            {
+                FCntDown = newFCntDown,
+                FCntUp = clientFCntUp,
+                GatewayId = this.gatewayId
+            };
+
+            this.StoreInfo(serverStateForDeviceInfo);
+            return serverStateForDeviceInfo;
+        }
+
+        public bool TryGetInfo(out DeviceCacheInfo info)
+        {
+            info = null;
+            if (!this.IsLockOwner)
+            {
+                throw new InvalidOperationException($"Trying to read information without owning the lock. Device: {this.devEUI} Gateway: {this.gatewayId}");
+            }
+
+            string cachedFCnt = redisCache.StringGet(this.devEUI, CommandFlags.DemandMaster);
+            if (string.IsNullOrEmpty(cachedFCnt))
+            {
+                return false;
+            }
+
+            info = (DeviceCacheInfo)JsonConvert.DeserializeObject(cachedFCnt, typeof(DeviceCacheInfo));
+            return info != null;
+        }
+
+        public void StoreInfo(DeviceCacheInfo info)
+        {
+            if (!this.IsLockOwner)
+            {
+                throw new InvalidOperationException($"Trying to update information without owning the lock. Device: {this.devEUI} Gateway: {this.gatewayId}");
+            }
+
+            LoRaDeviceCache.redisCache.StringSet(this.devEUI, JsonConvert.SerializeObject(info), new TimeSpan(30, 0, 0, 0), When.Always, CommandFlags.DemandMaster);
+        }
+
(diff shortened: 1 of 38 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -126,7 +126,13 @@ public async Task<bool> SendEventAsync(LoRaDeviceTelemetry telemetry, Dictionary
                     // Enable retry for this send message, off by default
                     this.SetRetry(true);
 
-                    var messageJson = JsonConvert.SerializeObject(telemetry, Formatting.None);
+                    var settings = new JsonSerializerSettings
```
fbeltrao: We could have the settings as an static and reuse every time, instead of creating a new one for every serialization?

fbeltrao: Out of curiosity: did you make this change to avoid sending `"dupmsg": false`?
If that is the case, what about using the NullValueHandling?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
```diff
@@ -71,6 +71,9 @@ public class LoRaDeviceTelemetry
         [JsonProperty("edgets")]
         public long Edgets { get; set; }
 
+        [JsonProperty("dupmsg")]
```
fbeltrao: Just to see if I understand correctly.
If you don't have an strategy for the device `"dupmsg"` won't be part of the telemetry.

On the other hand, if you have an strategy for the device the `"dupmsg"` will be present, for true or false.

Is that correct?

LoRaEngine/test/LoRaWanNetworkServer.Test/DeduplicationStrategy_End2End_Tests.cs
```diff
@@ -0,0 +1,142 @@
...
+                .ReturnsAsync((Message)null);
+
+            await this.SendTwoMessages(mode, confirmed);
+        }
+
+        private async Task SendTwoMessages(DeduplicationMode mode, bool confirmed)
+        {
+            var simulatedDevice = new SimulatedDevice(TestDeviceInfo.CreateABPDevice(1));
+
+            var loRaDevice = this.CreateLoRaDevice(simulatedDevice);
```
fbeltrao: If you are trying to simulate parallel processing from multiple gateways you should create 2 LoRaDevice instances. Reusing the same will cause the requests to be queued under the unique instance.

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
```diff
@@ -0,0 +1,170 @@
...
+    using Microsoft.Azure.WebJobs;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+    using StackExchange.Redis;
+
+    public sealed class LoRaDeviceCache : IDisposable
+    {
+        private const string CacheKeyLockSuffix = "msglock";
+        private static readonly TimeSpan LockWaitingTimeout = TimeSpan.FromSeconds(10);
+        private static IDatabase redisCache;
```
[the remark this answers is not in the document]
debugerr: Ok. That's interesting. As far as I understand, if you do use a lock, the accessed fields are being read/write with volatile memory access. I would say that you do not need a volatile marker, if you use a lock. volatile is usually used, when you have a single writer thread and multiple read threads, where locks would make it unnecessarily expensive synchronizing the value. We can discuss this one. :)

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
```diff
@@ -0,0 +1,170 @@
...
+                throw new InvalidOperationException($"Trying to read information without owning the lock. Device: {this.devEUI} Gateway: {this.gatewayId}");
+            }
+
+            string cachedFCnt = redisCache.StringGet(this.devEUI, CommandFlags.DemandMaster);
+            if (string.IsNullOrEmpty(cachedFCnt))
+            {
+                return false;
+            }
+
+            info = (DeviceCacheInfo)JsonConvert.DeserializeObject(cachedFCnt, typeof(DeviceCacheInfo));
```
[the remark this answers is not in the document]
debugerr: Copied from original, I think. Using the generic one is preferred. Thanks.

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        private static void EnsureRedisInstance(ExecutionContext context)
+        {
+            if (redisCache != null)
+            {
+                return;
+            }
+
+            lock (typeof(LoRaDeviceCache))
```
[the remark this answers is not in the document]
debugerr: Totally. Also copied from old code. Thanks for catching.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -10,6 +10,13 @@ namespace LoRaWan.NetworkServer
     using Microsoft.Azure.Devices.Shared;
     using Microsoft.Extensions.Logging;
 
+    public enum DeduplicationMode
```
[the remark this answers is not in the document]
debugerr: :) I usually also do that. However, I checked and saw that most of the enums are defined along their used classes (LoRaPayloadKeyType, LoRaMessageType, FctrlEnum, PhysicalIdentifier, CidEnum, LoRaRegion). I was trying to be a good sport and apply the practices you use. ;) I'm happy to create a separate file and move the others to dedicated source files as well.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
debugerr: Since the DupMsg is a nullable, it would actually emit "dumpmsg": null which is not what I want. I can make sure that we use a static for the settings, makes sense. Thanks.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
debugerr: It will only be included if it is true. That's basically "marking" the message. Avoiding more data to be sent to the server than necessary.

LoRaEngine/test/LoRaWanNetworkServer.Test/DeduplicationStrategy_End2End_Tests.cs
debugerr: I know what you mean. If you really want it to be completely separate I would also have to duplicate the API services. However, I don't think that makes a difference in the Unit test as it all comes down to how the strategy handle the response from the API service and how the processing proceeds. If the API service mock handles that, I'm not sure having multiple devices will give us anything else in the test?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
fbeltrao: Ok. In case you don't know, the same could be achieved using a DefaultValueHandling on the property: 

```c#
[JsonProperty("dupmsg", DefaultValueHandling = DefaultValueHandling.Ignore)]
public bool IsDuplicated { get; set; }
```

Sample: https://dotnetfiddle.net/dUoPJe

LoRaEngine/test/LoRaWanNetworkServer.Test/DeduplicationStrategy_End2End_Tests.cs
fbeltrao: Makes sense. I was not sure if you wanted to test processing the message in parallel vs handling the responses correctly. But, that would be more something to be done at the function level, but unfortunately we don't have any, right?

PR #130, opened 2019-02-14
Update issue templates

niksacdev:
added issue template for bug and enhancements

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,53 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+<!--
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,53 @@
...
+3.
+4.
+
+## Context (Environment)
+
+### Device (Host) Operating System
+<!-- Ubuntu 16.04, Ubuntu 18.04, Windows IoT Core, etc. -->
+
+### Architecture
+<!-- armv7, arm32, x86 etc. -->
```
Mandur: @ronniesa  Should the SPI dev version be mentionned here in case of x86

ronniesa: no need for this we can always ask if we think the problem could be related..... but it may confuse customers too

PR #132, opened 2019-02-16
Fix non-breaking exceptions when using the network server outside of Edge

Mandur:
When running the solution outside of IoT Edge, in Visual Studio directly on my machine, the exception log windows is literally swamped by the following exception:

```
Exception thrown: 'Microsoft.CSharp.RuntimeBinder.RuntimeBinderException' in Microsoft.CSharp.dll ("Cannot implicitly convert type 'Newtonsoft.Json.Linq.JValue' to 'string'. An explicit conversion exists (are you missing a cast?)") Exception thrown: 'Microsoft.CSharp.RuntimeBinder.RuntimeBinderException' in Microsoft.CSharp.dll ("Cannot implicitly convert type 'Newtonsoft.Json.Linq.JValue' to 'string'. An explicit conversion exists (are you missing a cast?)") 
```
This makes it impossible to break on exception when developing :( and definitely impact my dev cycle.

After some investigation I found out that adding the .Value (as it was already done some times) fixed the issue.

Do you want me to run a FULL CI for this?

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
@@ -121,7 +121,7 @@ namespace LoRaWan.NetworkServer
                 if (twin.Properties.Desired.Contains(TwinProperty.AppSKey))
                 {
                     // ABP Case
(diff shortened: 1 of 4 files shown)


fbeltrao:
This change stops exceptions from happening when a LoRa device is initialised?
I wonder why it wouldn't happen on every environment and only when debugging locally.

fbeltrao:
After looking again with @Mandur this change prevents an internal exception in the TwinCollection class. Please harmonise the changes in all twin get operations. Using cast instead of .Value.

PR #133, opened 2019-02-18
Refactor functions as a base for ADR implementation

debugerr:
- refactor functions api
- abstract redis cache
- refactor unified unit tests for API version checks
- add unit tests for function implementation


LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
```diff
@@ -4,47 +4,33 @@
...
-            string region = string.Empty;
-            string resetPin = string.Empty;
-
-            queryStrings.TryGetValue("deviceName", out deviceName);
-            queryStrings.TryGetValue("publishingUserName", out publishingUserName);
-            queryStrings.TryGetValue("publishingPassword", out publishingPassword);
-            queryStrings.TryGetValue("region", out region);
-            queryStrings.TryGetValue("resetPin", out resetPin);
+
+            queryStrings.TryGetValue("deviceName", out string deviceName);
```
fbeltrao: nit: any reason to use `string` instead of `var`?

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCacheRedisStore.cs
```diff
@@ -0,0 +1,60 @@
...
+    using Microsoft.Extensions.Configuration;
+    using StackExchange.Redis;
+
+    public class LoRaDeviceCacheRedisStore : ILoRaDeviceCacheStore
+    {
+        private IDatabase redisCache;
+
+        public LoRaDeviceCacheRedisStore(ExecutionContext context)
+        {
+            var config = new ConfigurationBuilder()
```
fbeltrao: Suggestion: this configuration could be built once, for RegistryManager and RedisStore

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCacheRedisStore.cs
```diff
@@ -0,0 +1,60 @@
...
+                throw new Exception(errorMsg);
+            }
+
+            var redis = ConnectionMultiplexer.Connect(redisConnectionString);
+            this.redisCache = redis.GetDatabase();
+        }
+
+        public bool LockTake(string key, string value, TimeSpan timeout)
+        {
+            return this.redisCache.LockTake(key, value, timeout);
```
fbeltrao: What is your take in using redis async as opposed to sync?

LoRaEngine/LoraKeysManagerFacade/LoRaRegistryManager.cs
```diff
@@ -0,0 +1,59 @@
...
+        private static readonly object SingletonLock = new object();
+
+        private static RegistryManager current;
+
+        /// <summary>
+        /// Explicit initialization of a <see cref="RegistryManager"/> implementation.
+        /// </summary>
+        /// <param name="testManager">RegistryManager to be used</param>
+        /// <remarks>Note: only used for unit testing</remarks>
+        public static void InitRegistryManager(RegistryManager testManager)
```
fbeltrao: I think you mean `registryManager` instead of `testManager`

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
debugerr: here I think it's hard to see what the type will be if you do var. To me this is easier to read and string is short enough.

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCacheRedisStore.cs
debugerr: True. Will change that. There is another thing I noticed: we add the option to reload on all of these, if the file changes, but we do not really do anything if we reload: 
1. are we ever going to need to reload (do we support updating the config file)
2. if 1 is true, we should add code to actually refresh the various objects we build from it
3. if 1 is false, we should remove the reload option.

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCacheRedisStore.cs
debugerr: Generally, I like to use async where they are provided. I sticked to this because it has been used before and I was assuming the perf characteristics were good enough for this to be a sync call. I think for now, I would keep it this way, unless we see a problem with scalability (which is unlikely in my opinion)

LoRaEngine/LoraKeysManagerFacade/LoRaRegistryManager.cs
debugerr: I only want this to be used in unit tests, that's why I named it this way to make it more obvious.

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCacheRedisStore.cs
fbeltrao: AFAIK if the configuration file changes the Azure Function will be restarted. It might be useful for local development, but I wouldn't invest any time on it.

debugerr: Ok. Then I'll disable the re-load flag.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -19,126 +19,82 @@ namespace LoraKeysManagerFacade
...
                         {
-                            return (ActionResult)new BadRequestObjectResult("UsedDevNonce");
+                            throw new DeviceNonceUsedException();
                         }
 
-                        redisCache.StringSet(cacheKey, devNonce, new TimeSpan(0, 1, 0), When.Always, CommandFlags.DemandMaster);
+                        deviceCache.SetValue(devNonce, TimeSpan.FromMinutes(1));
 
-                        IoTHubDeviceInfo iotHubDeviceInfo = new IoTHubDeviceInfo();
                         var device = await registryManager.GetDeviceAsync(devEUI);
```
ronniesa: Here is something that we have never implemented but as a security measuere we should check that the lengh is exaclty the one of DevEUI (I belive 16) and is a valid hex format. In case the function key get compromise you can get only the lora devices out and not all the other iot hub devices. You don't neet to add it to this PR but would make sense to have it in here at some point

PR #138, opened 2019-02-22
Add support to class C devices

fbeltrao:
- Added support to class C devices in decoder and module direct method
- Added documentation to enable class C devices
- Moved Downlink message creating to DownLinkMessageBuild (shared by class A and C processing)
- Consolidated fcnt down resolution in message processing
- Fixed a bug that could sometimes cause a deduplication loser to retry resolving a fcnt down
- Reformatted cloud to device messages using the payload instead of a combination of payload and properties
- Removed obsolete MessageProcessor class and ILoRaDeviceRegistry.GetDeviceForPayloadAsync methods
- Added method to obtain device by DevEUI in Azure Function (class C scenario)
- Optimised fcnt changes in device twin to happen once a difference of 10 is identified (fcnt up or down)
- Fixed a bug in telemetry sent to IoT Hub. The decrypted value will always be available in base64 format in `rawpayload` property
- Move join accept logic to JoinRequestMessageHandler
- Optimised integration test to build only once

**Important**
CI might failed due to problem receiving direct method in module. Restarting edgeHub/networksrvmodule usually fixes the problem

Fixes AB#698


fbeltrao:
- All tests (except class C) working: https://dev.azure.com/epicstuff/Azure%20IoT%20Edge%20LoRaWAN%20Starter%20Kit/_build/results?buildId=2328

- Class C test only (rm edgeHub container & restart LoRaWanNetworkSrvModule): https://dev.azure.com/epicstuff/Azure%20IoT%20Edge%20LoRaWAN%20Starter%20Kit/_build/results?buildId=2329

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -102,9 +121,11 @@ public LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClient loRaDeviceCli
             this.IsABPRelaxedFrameCounter = true;
             this.PreferredWindow = 1;
             this.hasFrameCountChanges = false;
-            this.fcntLock = new object();
+            this.fcntSync = new object();
             this.confirmationResubmitCount = 0;
+            this.queueSync = new object();
             this.queuedRequests = new Queue<LoRaRequest>();
+            this.classType = LoRaDeviceClassType.A;
```
debugerr: All of the initializers that are not passed into the ctor could be directly assigned when you declare the field. The advantage I see with that is that if you ever add another contstructor overload, you have the fields set to the correct defaults. 

I think I made that comment before, and it's probably personal preference, but the defaults for the various types, would not need to be re-initialized (it's additional code that's running):

bool is by default false (does not need to be initialized)
int is by default 0
classType is by default type.A

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -205,6 +233,14 @@ public async Task<bool> InitializeAsync()
                     this.Deduplication = mode;
                 }
 
+                if (twin.Properties.Desired.Contains(TwinProperty.ClassType))
```
debugerr: I understand there are no other class types we use explicitly at the moment, but woudln't it make sense to add the parsing for all values in the enum?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -306,9 +342,20 @@ public async Task<bool> SaveFrameCountChangesAsync()
...
+        /// </summary>
+        void AcceptFrameCountChanges(int savedFcntUp, int savedFcntDown)
+        {
+            lock (this.fcntSync)
             {
-                this.hasFrameCountChanges = false;
+                this.lastSavedFcntUp = savedFcntUp;
+                this.lastSavedFcntDown = savedFcntDown;
+
+                this.hasFrameCountChanges = this.fcntDown != this.lastSavedFcntDown || this.fcntUp != this.lastSavedFcntUp;
```
debugerr: Nice. There was a race here before that I also fixed in our feature branch. Will have to make sure we get all of your changes. :)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIServiceBase.cs
```diff
@@ -34,10 +35,20 @@ public abstract class LoRaDeviceAPIServiceBase
...
+
         /// <summary>
         /// Sets the new URL value
         /// </summary>
-        public void SetURL(string value) => this.URL = value;
+        public void SetURL(string value) => this.URL = this.SanitizeApiURL(value);
+
+        private string SanitizeApiURL(string value)
+        {
+            if (string.IsNullOrEmpty(value) || value.EndsWith('/'))
```
debugerr: You may also want to trim it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
```diff
@@ -362,6 +334,7 @@ public void ResetDeviceCache()
         {
             var oldResetCacheToken = this.resetCacheToken;
             this.resetCacheToken = new CancellationTokenSource();
+            this.resetCacheChangeToken = new CancellationChangeToken(this.resetCacheToken.Token);
```
debugerr: I think you need to lock this. Otherwise you can do this wil adding items to the cache and invalidating them right away.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
```diff
@@ -44,105 +60,137 @@ public async ValueTask<JObject> DecodeMessageAsync(byte[] payload, byte fport, s
...
+                            var externalDecoderResponse = JsonConvert.DeserializeObject<Dictionary<string, object>>(externalRawResponse);
+                            if (externalDecoderResponse.TryGetValue("cloudToDeviceMessage", out var cloudToDeviceObject))
+                            {
+                                if (cloudToDeviceObject is JObject jsonObject)
+                                {
+                                    loRaCloudToDeviceMessage = jsonObject.ToObject<LoRaCloudToDeviceMessage>();
+                                }
+                                else
+                                {
+                                    loRaCloudToDeviceMessage = (LoRaCloudToDeviceMessage)cloudToDeviceObject;
```
debugerr: is this a save direct cast or should this be an indirect one?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaWan.NetworkServer.csproj
```diff
@@ -26,6 +26,11 @@
   <ItemGroup>
     <AdditionalFiles Include="../../../../stylecop.json" Link="stylecop.json" />
   </ItemGroup>
+  <ItemGroup>
```
debugerr: probably not something you want to check in?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/RawPayloadJsonConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.NetworkServer
+{
+    using System;
+    using Newtonsoft.Json;
+
+    public class RawPayloadJsonConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType) => true;
```
debugerr: A comment would be good here.

LoRaEngine/LoraKeysManagerFacade/SearchDeviceByDevEUI.cs
```diff
@@ -0,0 +1,69 @@
...
+        public static async Task<IActionResult> GetDeviceByDevEUI([HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)]HttpRequest req, ILogger log, ExecutionContext context)
+        {
+            return await RunGetDeviceByDevEUI(req, log, context, ApiVersion.LatestVersion);
+        }
+
+        private static async Task<IActionResult> RunGetDeviceByDevEUI(HttpRequest req, ILogger log, ExecutionContext context, ApiVersion currentApiVersion)
+        {
+            try
+            {
+                VersionValidator.Validate(req);
```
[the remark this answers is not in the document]
fbeltrao: I changed to remain consistent with other functions. However, all tests I've written are against the outer method. For a test to succeed the api-version must be part of the test request.

LoRaEngine/LoraKeysManagerFacade/SearchDeviceByDevEUI.cs
```diff
@@ -0,0 +1,69 @@
...
+                    });
+                }
+
+                log.LogDebug($"Search for {devEUI} found 1 device");
+                return new OkObjectResult(result);
+            }
+            else
+            {
+                log.LogInformation($"Search for {devEUI} found 0 devices");
+                return new NotFoundObjectResult(result);
```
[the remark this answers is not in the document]
fbeltrao: As you mentioned, this will return 404 which means content not found. If clients are implemented correct that should be enough to indicate that there is no device with the requested DevEUI.

Reasoning behind returning an empty array as content is for those clients that are not checking the status code. They just serialise the return content to a list of devices that will be empty.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -31,18 +31,15 @@ public class LoRaPayloadData : LoRaPayload
         /// <summary>
         /// Gets the LoRa payload fport as value
         /// </summary>
-        public byte FPort
+        public byte GetFPort()
```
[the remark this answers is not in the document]
fbeltrao: My rule of thumb is that properties should not allocate memory and be simple. If those requirements are not met I prefer methods as they give the sense that you should not be calling all the time, but using a local variable to keep the result.

In this particular case, keeping the value of FPort in a local variable to not keep checking if is null all the time. 

Being a property might be even more confusing with the backing property: `Memory<byte> Fport { get; set; }`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DecodedPayloadValue.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// Defines a simple decoded payload value
+    /// Represents the value as: { "value":1 }
+    /// </summary>
+    public class DecodedPayloadValue
+    {
+        [JsonProperty("value")]
+        public object Value { get; set; }
+
+        public DecodedPayloadValue(object value)
```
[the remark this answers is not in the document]
fbeltrao: A default constructor is needed for deserialise. This class is only to deserialize the default provided decoded payload.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DecodingFailedPayload.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class DecodingFailedPayload
+    {
+        [JsonProperty("error")]
+        public string Error { get; set; }
+
+        [JsonProperty("errorDetail", NullValueHandling = NullValueHandling.Ignore)]
+        public string ErrorDetail { get; set; }
+
+        public DecodingFailedPayload(string error, string errorDetail)
```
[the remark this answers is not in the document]
fbeltrao: Same as before, server side instance create to send over the wire.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -292,21 +318,19 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
                             {
                                 // The message coming from the device was not confirmed, therefore we did not computed the frame count down
                                 // Now we need to increment because there is a C2D message to be sent
-                                fcntDown = await frameCounterStrategy.NextFcntDown(loRaDevice, payloadFcnt);
+                                fcntDown = await this.EnsureHasFcntDownAsync(loRaDevice, fcntDown, payloadFcnt, frameCounterStrategy);
 
-                                if (fcntDown == 0)
+                                if (!fcntDown.HasValue || fcntDown <= 0)
```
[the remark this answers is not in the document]
fbeltrao: Logically not, but being an int it could.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -152,71 +160,89 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
+                                        if (string.IsNullOrEmpty(decodePayloadResult.CloudToDeviceMessage.DevEUI) || string.Equals(loRaDevice.DevEUI, decodePayloadResult.CloudToDeviceMessage.DevEUI, StringComparison.InvariantCultureIgnoreCase))
+                                        {
+                                            // sending c2d to same device
+                                            cloudToDeviceMessage = decodePayloadResult.CloudToDeviceMessage;
+                                            fcntDown = await this.EnsureHasFcntDownAsync(loRaDevice, fcntDown, payloadFcnt, frameCounterStrategy);
+
+                                            if (!fcntDown.HasValue || fcntDown <= 0)
+                                            {
+                                                // We did not get a valid frame count down, therefore we should not process the message
+                                                _ = cloudToDeviceMessage.AbandonAsync();
```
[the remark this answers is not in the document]
fbeltrao: The intention is to not wait for it to be finished. So, afaik, it will return as soon as the amqp/mqtt will be sent. Could use `Task.Run` to make 100% in parallel.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -365,185 +390,60 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
-                    requiresDeviceAcknowlegement = true;
-                    loRaDevice.LastConfirmedC2DMessageID = cloudToDeviceMessage.MessageId ?? Constants.C2D_MSG_ID_PLACEHOLDER;
-                }
-
-                if (cloudToDeviceMessage.Properties.TryGetValueCaseInsensitive("fport", out var fPortValue))
-                {
-                    fport = byte.Parse(fPortValue);
-                }
+            var actualMessage = await loRaDevice.ReceiveCloudToDeviceAsync(timeAvailableToCheckCloudToDeviceMessages);
+            if (actualMessage != null)
```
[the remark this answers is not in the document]
fbeltrao: Weird that Stylecop did not complain about it. Changed to ternary return.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -365,185 +390,60 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
-
-                // if specific twins are set, specify second channel to be as specified
-                else
+                // Not valid if there is no mac command or there is a payload
+                if (cloudToDeviceMsg.MacCommands?.Count == 0 || cloudToDeviceMsg.GetPayload().Length > 0)
                 {
-                    freq = this.configuration.Rx2DataFrequency;
-                    datr = this.configuration.Rx2DataRate;
-                    Logger.Log(loRaDevice.DevEUI, $"using custom DR second receive windows freq : {freq}, datr:{datr}", LogLevel.Information);
+                    Logger.Log(loRaDevice.DevEUI, $"invalid mac command fport usage in C2D message '{cloudToDeviceMsg.MessageId}'", LogLevel.Error);
```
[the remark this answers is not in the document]
fbeltrao: They should be lower case since they have the device identifier as prefix.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -0,0 +1,289 @@
...
+                foreach (var requestedMacCommand in requestedMacCommands)
+                {
+                    switch (requestedMacCommand.Cid)
+                    {
+                        case CidEnum.LinkCheckCmd:
+                        {
+                            if (rxpk != null)
+                            {
+                                var linkCheckAnswer = new LinkCheckAnswer(rxpk.GetModulationMargin(), 1);
+                                macCommands.Add((int)CidEnum.LinkCheckCmd, linkCheckAnswer);
```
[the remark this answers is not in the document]
fbeltrao: Changed to TryAdd, ensuring that logging will happen once.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaCloudToDeviceMessageWrapper.cs
```diff
@@ -0,0 +1,118 @@
...
+                this.invalidErrorMessage = "Cloud message does not have a body";
+            }
+        }
+
+        public byte Fport
+        {
+            get
+            {
+                if (this.parseCloudToDeviceMessage != null)
+                    return this.parseCloudToDeviceMessage.Fport;
```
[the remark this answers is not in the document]
fbeltrao: Replaced by `return this.parseCloudToDeviceMessage?.Fport ?? 0;`

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaCloudToDeviceMessageWrapper.cs
```diff
@@ -0,0 +1,118 @@
...
+    class LoRaCloudToDeviceMessageWrapper : ILoRaCloudToDeviceMessage
+    {
+        private readonly LoRaDevice loRaDevice;
+        private readonly Message message;
+        private LoRaCloudToDeviceMessage parseCloudToDeviceMessage;
+        private string invalidErrorMessage;
+
+        public LoRaCloudToDeviceMessageWrapper(LoRaDevice loRaDevice, Message message)
+        {
+            this.loRaDevice = loRaDevice;
```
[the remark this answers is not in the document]
fbeltrao: I've added, but this should be done everywhere. Adding input validation will break a few tests that are parsing `null` for objects not being used by the workload being tested.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
fbeltrao: I see the advantage of setting the value together with the property declaration, what I dislike is that during a debugging session stepping into the ctor will go through many lines. But, this is probably better than introducing bugs by adding constructors. Another way to address this issues is by having ctors calling other ctors.

Regarding initialising members with default values, specifically enums, the advantage I see is that the code is resilient to changes to the enum definition (i.e. defining as first enum None or NotSet could break the code).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
fbeltrao: Current behavior will treat any value other than "C" as a class A device. Notice that there is a 'B' defined in the enum, but it is not yet supported. I have chosen to assume "A" if any invalid value is set.

However, if the requirement is to fail if the device class is set to "B" I will have to make changes.

@ronniesa what is your take? Fail or switch silently to class A device if the value is B or invalid?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
fbeltrao: Do you mean adding a lock to prevent invalidating items that are being added as the cache is cleared?

If that is the case, I think it is better to use lock and let items that were just added to cache be invalidated instead of adding a lock before adding anything to the cache. Cache clear is supposed to be called not too often. If we start to rely on it more heavily we should be able to invalidate only parts of the cache.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
fbeltrao: This was wrong, the object would never be the expected type. It will be of type `JObject` or `JValue`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaWan.NetworkServer.csproj
fbeltrao: I don't mind adding a class diagram. Removing for now.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/RawPayloadJsonConverter.cs
fbeltrao: This was an experimentation that I forgot to remove. File has been deleted.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaWan.NetworkServer.csproj
debugerr: Me neither, but I think it was referencing a local path. Usually the trickiest part is to keep a checked in CD up to date.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
debugerr: Not sure, but I think we mean the same thing: 
What I find dangerous here is that while calling ResetDeviceCache there might be entries coming in. Some of them might get removed and invalidated (added after the new resetCacheToken was initialized) and others might be removed (those that still got the original token).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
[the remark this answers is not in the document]
debugerr: I agree that we have to be careful with the use of properties that are extensive, but here, you don't have a single Heap allocation. All you do is copying stuff on the stack, which should be very fast.

Here from the book Effective C#: 
<book>
Property accessors also establish performance expectations for your users. A property access looks like a data field access. It should not have performance characteristics that are significantly different from those of a simple data access. Property accessors should not perform lengthy computations, or make cross-application calls (such as perform database queries), or do other lengthy operations that would be inconsistent with your users’ expectations for a property accessor.
</book>

I personally like the Properties syntax better, but after all it's a personal preference. In the end, they are both methods, in one case it is obvious that you call a method, in the other less (most of the people know though). 

I would use more properties, but cache the results of those on the stack if you need them multiple times. That's a good practice anyway imho.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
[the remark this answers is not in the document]
debugerr: Ok. If you do not want to await, can't you just remove the _ = ? Is there a style rule complaining? I wouldn't start another thread for this.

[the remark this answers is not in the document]
fbeltrao: Discard was added to suppress warning [CS4014](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014).

PR #143, opened 2019-03-06
Add DlSettings join properties

Mandur:
* Add RX1DROffset Join Property
* Add RX2DR Join Property

Fix AB#1128

NB. Testing RX1DROffset on our current integration platform seems not to be possible because of device limitation. I tested manually, and have Simulated tests that ensure it is performing good and to prevent any regression. Testing this feature might be added in the future and I will add a work item to keep it on track.


Mandur:
I will resubmit the PR after a rebase on Dev.

PR #145, opened 2019-03-08
Add DlSettings join properties

Mandur:
* Add RX1DROffset Join Property
* Add RX2DR Join Property

Added since #143  a combination Desired/Reported to better reflect current device state and only allow changes on those properties on join requests-

Fix AB#1128


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -71,23 +71,11 @@ internal static class DownlinkMessageBuilder
...
-                    datr = loRaRegion.DRtoConfiguration[loRaRegion.RX2DefaultReceiveWindows.dr].configuration;
-                }
-                else
-                {
-                    // if specific twins are set, specify second channel to be as specified
-                    freq = configuration.Rx2DataFrequency;
-                    datr = configuration.Rx2DataRate;
-                    Logger.Log(loRaDevice.DevEUI, $"using custom DR second receive windows freq : {freq}, datr:{datr}", LogLevel.Information);
-                }
+                (freq, datr) = request.Rxpk.GetDownstreamRX2DRAndFreq(loRaDevice.DevEUI, configuration.Rx2DataRate, configuration.Rx2DataFrequency, loRaRegion, loRaDevice.ReportedRX2DataRate);
```
debugerr: can you add a comment explaining why the calculation for the RX2 has to go through the Rxpk vs. the calculation below for the RX1? To me that's not obvious.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -175,6 +164,37 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
                 // Build join accept downlink message
                 Array.Reverse(netId);
                 Array.Reverse(appNonceBytes);
+                byte[] dlSettings = new byte[1]
+                {
```
debugerr: no need to initialize. It will be 0.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -175,6 +164,37 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
...
+                }
+                else
+                {
+                    Logger.Log(devEUI, $"twin RX2 datarate values are not within acceptable values", LogLevel.Error);
+                }
+
+                if (loRaDevice.DesiredRX1DROffset >= 0 && loRaDevice.DesiredRX1DROffset < request.LoRaRegion.RX1DROffsetTable.GetUpperBound(1))
+                {
+                    var rx1droffset = (byte)(loRaDevice.DesiredRX1DROffset << 4);
+                    if (dlSettings == null)
```
debugerr: can never be null

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -175,6 +164,37 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
...
+                    {
+                        dlSettings = new byte[1] { rx1droffset };
+                    }
+                    else
+                    {
+                        dlSettings[0] = (byte)(dlSettings[0] + rx1droffset);
```
debugerr: Is the init here deliberately adding to the existing value? If so, a comment would be good explaining what we do here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -132,6 +145,8 @@ public LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClient loRaDeviceCli
             this.queueSync = new object();
             this.queuedRequests = new Queue<LoRaRequest>();
             this.classType = LoRaDeviceClassType.A;
+            this.ReportedRX1DROffset = 0;
```
debugerr: no need to set these to 0. They already are guaranteed to be 0 at this point.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -520,6 +545,25 @@ internal async Task<bool> UpdateAfterJoinAsync(string devAddr, string nwkSKey, s
...
+                reportedProperties[TwinProperty.RX1DROffset] = null;
+            }
+
+            if (this.DesiredRX2DataRate != DefaultJoinValues)
+            {
+                reportedProperties[TwinProperty.RX2DataRate] = this.DesiredRX2DataRate;
+            }
+            else
+            {
+                reportedProperties[TwinProperty.RX2DataRate] = null;
```
debugerr: why do you blank it out if they are the same value? Wouldn't we want to keep the value to ensure the same behavior will apply next time this runs?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
Mandur: I am not sure I follow you is your question asking why I did not put the method in the LoRaRegion and in the Rxpk instead?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
```diff
@@ -50,6 +50,10 @@ public class LoRaPayloadJoinAccept : LoRaPayload
         /// </summary>
         public Memory<byte> Fcnt { get; set; }
 
+        public int Rx1DrOffset => (this.DlSettings.Span[0] >> 4) & 0b00000111;
```
debugerr: right shift and then mask it? What is the value in DlSettings.Span[0]? if you right shift it to the first 4 bits, you will always end up with 7 no?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
```diff
@@ -122,5 +122,45 @@ public uint GetModulationMargin()
 
             return (uint)signedMargin;
         }
+
+        public (double freq, string datr) GetDownstreamRX2DRAndFreq(string devEUI, string nwkSrvRx2Dr, double nwkSrvRx2Freq, Region loraRegion, int? rx2DrFromTwins)
```
debugerr: Francisco will love this. ;)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
```diff
@@ -122,5 +122,45 @@ public uint GetModulationMargin()
 
             return (uint)signedMargin;
         }
+
+        public (double freq, string datr) GetDownstreamRX2DRAndFreq(string devEUI, string nwkSrvRx2Dr, double nwkSrvRx2Freq, Region loraRegion, int? rx2DrFromTwins)
+        {
+            double freq = 0;
+            string datr = string.Empty;
```
debugerr: init to null? looks like you always going to assign a string?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -265,13 +265,21 @@ public double GetDownstreamChannelFrequency(Rxpk upstreamChannel)
...
             this.EnsureValidRxpk(upstreamChannel);
             if (this.LoRaRegion == LoRaRegion.EU868)
             {
-                // in case of EU, you respond on same frequency as you sent data.
-                return upstreamChannel.Datr;
+                // If the rx1 offset is a valid value we use it, otherwise we keep answering on normal datar
+                if (rx1DrOffset < this.RX1DROffsetTable.GetUpperBound(1))
+                {
+                    // in case of EU, you respond on same frequency as you sent data.
+                    return this.DRtoConfiguration[(uint)this.RX1DROffsetTable[this.GetDRFromFreqAndChan(upstreamChannel.Datr), rx1DrOffset]].configuration;
```
debugerr: is it worth performing some bounds checking before indexing into the table to avoid potential out of bounds exceptions?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionLimits.cs
```diff
@@ -11,6 +11,18 @@ public class RegionLimits
 
         public List<string> DatarateRange { get; set; }
 
+        public bool IsCurrentDRIndexWithinAcceptableValue(uint dr)
+        {
+            if (dr < this.DatarateRange.Count)
```
debugerr: can be replaced with return dr < this.DatarateRange.Count

LoRaEngine/test/LoRaWan.Test.Shared/TestDeviceInfo.cs
```diff
@@ -50,6 +50,10 @@ public class TestDeviceInfo
 
         public char ClassType { get; set; } = 'A';
 
+        public int RX2DataRate { get; set; } = 0;
```
debugerr: no need to initialize either of them.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
debugerr: this comment didn't make it for some reason: what I don't understand is, you right shift by 4 and then mask the first 4 bytes with 01111 which will result to 7 always, unless DlSettings.Span is larger than 8 bytes? Not sure what it is really. Did you mean to left shift?

Mandur: I need the three bits located in position 4:6. I could mask it and shift shift, I don't know if the order have an impact here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
Mandur: Why so? I did put a Capital letter !

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
Mandur: Well, normally StyleCop complains when I don't initalize variables, So I guess I just initialize everything now. I could change this to remove the String.Empty

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
Mandur: The EnsureValidRxpk Method is checking the Datar, and the rx1DrOffset is checked in the condition just above. However I realize the EnsureValidRxpk is throwing an exception, but nothing catches it. I will fix this.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
debugerr: basically, I'm just saying, I don't understand what's going on when looking at the code and suggest to add a comment to make it clear why the 2 windows take such different routes when it comes to the freq / datr calculation.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
debugerr: All good.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
fbeltrao: Thinking behind I suppose is to save twin size. Not having a property or having with default value results in the same.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -302,34 +363,33 @@ public uint GetMaxPayloadSize(string datr)
...
             if (this.LoRaRegion == LoRaRegion.EU868)
             {
                 if (rxpk.Freq < EU868.RegionLimits.FrequencyRange.min ||
-                    rxpk.Freq > EU868.RegionLimits.FrequencyRange.max)
-                {
-                    throw new RegionLimitException($"Frequency {rxpk.Freq} in {this.LoRaRegion} region was not within the acceptable range of frequencies.", RegionLimitExceptionType.Frequency);
-                }
-
-                if (!EU868.RegionLimits.DatarateRange.Contains(rxpk.Datr))
+                    rxpk.Freq > EU868.RegionLimits.FrequencyRange.max ||
```
fbeltrao: nit: regions limits should be a property of the Region instance defined during the construction. This way IsValidRxpk does not need to a if (region == x, y, z).

The code would look like this:
```c#
bool IsValidRxpk(Rxpk rxpk)
{
    if (rxpk.Freq < this.RegionLimits.FrequencyRange.min || ...)
}
```

LoRaEngine/test/LoRaWanNetworkServer.Test/MessageProcessorJoinTest.cs
```diff
@@ -418,5 +418,409 @@ public async Task When_Api_Returns_DevAlreadyUsed_Should_Return_Null()
...
+            // Create Rxpk
+            var rxpk = joinRequest.SerializeUplink(simulatedDevice.LoRaDevice.AppKey).Rxpk[0];
+
+            var devNonce = ConversionHelper.ByteArrayToString(joinRequest.DevNonce);
+            var devAddr = string.Empty;
+            var devEUI = simulatedDevice.LoRaDevice.DeviceID;
+            var appEUI = simulatedDevice.LoRaDevice.AppEUI;
+
+            // Device twin will be queried
+            var twin = new Twin();
```
fbeltrao: fyi: there is an extension method to create OTAA/ABP twins, so you don't have to hardcode all values.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
Mandur: Good suggestion I will add it to the region limit PR I am having

LoRaEngine/test/LoRaWanNetworkServer.Test/MessageProcessorJoinTest.cs
Mandur: TBH, I just inspired myself from the test When_Getting_Device_Information_From_Twin_Returns_JoinAccept but this is a good point. I guess as we are in Strict mode I would still need to setup gettwinasync right? The method you mention is simulatedDevice.CreateOTAATwin correct?

fbeltrao: Correct. Just for the next time you write a test ;)

Mandur: I can change them, it's no problems

Docs/quickstart.md
```diff
@@ -154,10 +154,23 @@ Customization's to lora devices are set by creating specific twin desired proper
...
 
 **Important**: changes made to twin desired properties in devices that are already connected will only take effect once the network server is restarted or [cache is cleared](#cache-clearing).
 
+### Device Reported Properties
+
+This section provide a brief description of the device reported properties you can see appear on your device.
+
+|Name|Description|Used By|
+|-|-|-|
+|DataRate|The Device Current Datarate|This value will be only reported if you are using Adaptive Data Rate||
```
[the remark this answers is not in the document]
Mandur: Frankly, I am not sure how I could write this line without mentioning DataRate 3 times...  Would  |DataRate (DR)|The Device Current DR|This value will be only reported if your device has Adaptive Data Rate (ADR) enabled|| be better ?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
ronniesa: setting it to null basically deletes this twin property. Not sure that is the intention here but this would be the effect

Mandur: Yes, this was the intended behavior

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
fbeltrao: Because of usage of Tuples, I guess

PR #148, opened 2019-03-15
Implement RXDelay

Mandur:
* Implement RX Delay, time between receiving a message and starting to transmit.

Fix AB#1130


Mandur:
I will add fullci once dev has been built

LoRaEngine/test/LoRaWan.Test.Shared/IntegrationTestFixtureBase.Asserts.cs
```diff
@@ -135,6 +135,52 @@ public async Task AssertNetworkServerModuleLogExistsAsync(Func<string, bool> pre
             }
         }
 
+        /// <summary>
+        /// Check for The timing between an upstream message and a downstream one.
+        /// The method stop at first occurence found and is not expected to perform against multiple DS/US messages.
+        /// </summary>
+        public async Task CheckAnswerTimingAsync(uint rxDelay, bool isSecondWindows)
```
debugerr: isSecondWindow instead?

LoRaEngine/test/LoRaWan.Test.Shared/IntegrationTestFixtureBase.Asserts.cs
```diff
@@ -135,6 +135,52 @@ public async Task AssertNetworkServerModuleLogExistsAsync(Func<string, bool> pre
...
+            }
+        }
+
+        /// <summary>
+        /// Helper method to find the time of the message that contain the message argument.
+        /// </summary>
+        async Task<FindTimeResult> TryFindMessageTimeAsync(string message)
+        {
+            var log = await this.SearchUdpLogs(x => x.Contains(message));
+            int timeIndexStart = log.FoundLogResult.IndexOf(@"""tmst"":") + 7;
```
debugerr: +7 is likely just moving to the end of the token. What I usually do here, is create a const token (const string token = @"""tmst"":") and then use the token.Length to make it obvious.

LoRaEngine/test/LoRaWan.Test.Shared/IntegrationTestFixtureBase.Asserts.cs
```diff
@@ -135,6 +135,52 @@ public async Task AssertNetworkServerModuleLogExistsAsync(Func<string, bool> pre
...
+            bool success = false;
+            if (timeIndexStart > 0 && timeIndexStop > 0)
+            {
+                if (uint.TryParse(log.FoundLogResult.Substring(timeIndexStart, timeIndexStop - timeIndexStart), out parsedValue))
+                {
+                    success = true;
+                }
+            }
+
+            var result = new FindTimeResult()
```
debugerr: directly return the result instead of assigning first? return new FindTimeResult ...

LoRaEngine/test/LoRaWan.IntegrationTest/C2DMessageTest.cs
```diff
@@ -108,6 +109,21 @@ public async Task Test_OTAA_Confirmed_Receives_C2D_Message()
...
 
+                // Check that RXDelay was correctly used
+                if (this.ArduinoDevice.SerialLogs.Where(x => x.StartsWith("+CMSG: RXWIN1")).Count() > 0)
+                {
+                    await this.TestFixtureCi.CheckAnswerTimingAsync(2000000, false);
+                }
+                else
+                if (this.ArduinoDevice.SerialLogs.Where(x => x.StartsWith("+CMSG: RXWIN2")).Count() > 0)
+                {
+                    await this.TestFixtureCi.CheckAnswerTimingAsync(2000000, false);
```
debugerr: shouldn't the 2nd parameter be true here?

LoRaEngine/test/LoRaWanNetworkServer.Test/TestLoRaOperationTimeWatcher.cs
```diff
@@ -0,0 +1,27 @@
...
+        public TestLoRaOperationTimeWatcher(Region loraRegion, TimeSpan constantElapsedTime)
+            : base(loraRegion)
+        {
+            this.constantElapsedTime = constantElapsedTime;
+        }
+
+        /// <summary>
+        /// Gets time passed since start
+        /// </summary>
+        protected internal override TimeSpan GetElapsedTime() => this.constantElapsedTime;
```
debugerr: I see why the base class's method is virtual now. :)

Docs/quickstart.md
```diff
@@ -157,6 +157,7 @@ Customization's to lora devices are set by creating specific twin desired proper
...
+| RX Delay | Allows setting a custom wait time between receiving and transmission as specified in the specification. | Add twin desired property `"RXDelay": # ` where # is a valid specification number to set the RxDelay to wait for # seconds. Note that values 0 and 1 are default. Example: if a value of 2 is provided, The RX1 transmit windows will open 2 second after receiving the message. RX2 windows will always stay 1 second after RX1 | Setting a custom RXDelay could be usefull in case of long processing time, this could give more time to the server/device to compute their answers and lower the risk of missing the transmission windows. This is an advanced feature, it is not recommanded to be setup in normal cases.|
```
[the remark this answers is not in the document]
Mandur: This is true, I will adapt

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -234,14 +235,15 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
             double freq,
             long tmst,
             string devEUI,
-            byte[] dlSettings)
+            byte[] dlSettings,
+            ushort rxDelay)
         {
```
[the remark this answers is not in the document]
Mandur: Correct, not sure it should be in scope of this PR thought, but I could include in next one

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
```diff
@@ -139,7 +148,7 @@ public LoRaPayloadJoinAccept(byte[] inputMessage, string appKey)
             Array.Copy(inputMessage, 11, dlSettings, 0, 1);
             this.DlSettings = new Memory<byte>(dlSettings);
             var rxDelay = new byte[1];
-            Array.Copy(inputMessage, 11, rxDelay, 0, 1);
+            Array.Copy(inputMessage, 12, rxDelay, 0, 1);
```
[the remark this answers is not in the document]
Mandur: Yes

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionFactory.cs
```diff
@@ -37,23 +37,41 @@ public static bool TryGetRegion(LoRaRegion value, out Region region)
         }
```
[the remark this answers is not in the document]
Mandur: The region situation is a bit messy atm. I have another PR ready that revamp the region management.

LoRaEngine/test/LoRaWan.IntegrationTest/C2DMessageTest.cs
```diff
@@ -108,6 +109,21 @@ public async Task Test_OTAA_Confirmed_Receives_C2D_Message()
                 // +CMSG: ACK Received
                 await AssertUtils.ContainsWithRetriesAsync("+CMSG: ACK Received", this.ArduinoDevice.SerialLogs);
 
+                // Check that RXDelay was correctly used
+                if (this.ArduinoDevice.SerialLogs.Where(x => x.StartsWith("+CMSG: RXWIN1")).Count() > 0)
+                {
+                    await this.TestFixtureCi.CheckAnswerTimingAsync(2000000, false);
```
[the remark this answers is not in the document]
Mandur: I added a constant to better describe. the packet forwarder represent time in millionth of seconds

LoRaEngine/test/LoRaWan.Test.Shared/IntegrationTestFixtureBase.Asserts.cs
Mandur: I could do this in part, but I can't use the token.length as my string contains escape characters...

LoRaEngine/test/LoRaWan.IntegrationTest/C2DMessageTest.cs
Mandur: Yes this is an error, thanks for spotting!

LoRaEngine/test/LoRaWan.Test.Shared/IntegrationTestFixtureBase.Asserts.cs
debugerr: The length of the string should be correct no matter if you are using escape chars or not. If you put this in a const and check the Length, it will be 7.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaRequest.cs
```diff
@@ -52,5 +52,7 @@ public virtual void NotifySucceeded(LoRaDevice loRaDevice, DownlinkPktFwdMessage
         internal void SetPayload(LoRaPayload loRaPayload) => this.Payload = loRaPayload;
 
         internal void SetRegion(Region loRaRegion) => this.Region = loRaRegion;
+
+        public virtual LoRaOperationTimeWatcher GetTimeWatcher() => new LoRaOperationTimeWatcher(this.Region, this.StartTime);
```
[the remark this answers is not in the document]
Mandur: I am not sure of all the implication that would have on the tests,I don't think a singleton would work in this case. Let's discuss and integrate in another PR.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -71,15 +71,14 @@ internal static class DownlinkMessageBuilder
 
             if (receiveWindow == Constants.RECEIVE_WINDOW_2)
             {
-                tmst = rxpk.Tmst + timeWatcher.GetReceiveWindow2Delay(loRaDevice) * 1000000;
-
+                tmst = CalculateTime(rxpk.Tmst, timeWatcher.GetReceiveWindow2Delay(loRaDevice), loRaDevice.ReportedRXDelay);
```
fbeltrao: nit

tmst will always be rxpk.Tmst + delay.

The delay is computed based on the window we are plus device specific properties (as delaying the 1st window). For me, it is a bit more readable like this:

```c#
tmst = rxpk.Tmst + timeWatcher.GetWindowDelay(loRaDevice, receiveWindow);
```

You can now unit test all combinations of possible delays in the LoRaOperationTimeWatcher tests.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaRequest.cs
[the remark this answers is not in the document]
fbeltrao: Agree, we can call this CreateTimeWatcher in a new PR

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionFactory.cs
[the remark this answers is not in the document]
fbeltrao: This method is like any other bool TryXXX(out object outParameter).
If we cannot resolve we return false and set outParameter to null.

This method was created because the current implementation sets the result into a static variable making it unsafe for tests.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -381,5 +380,17 @@ private static ushort ValidateAndConvert16bitFCnt(uint fcntDown)
 
             return macCommands.Values;
         }
+
+        private static long CalculateTime(uint baseTime, int windowTime, ushort rXDelay)
+        {
+            if (rXDelay > 1 && rXDelay < 16)
+            {
+                return baseTime + (windowTime + rXDelay - 1) * 1000000;
```
fbeltrao: Can we use the constant CONVERT_TO_PKT_FWD_TIME?

PR #149, opened 2019-03-19
Refactor Region implementation

Mandur:
* Refactor region implementation to simplify
* Fix a bug occurring if out of bound join parameters are provided
* Add changes suggested in #148 

Fix [AB#1133](https://dev.azure.com/epicstuff/web/wi.aspx?pcguid=b4ad2b30-c669-4749-9c24-e42a8da5d35d&id=1133)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs
```diff
@@ -1,201 +1,228 @@
...
-    {
-        public static Region CurrentRegion
-        {
-            get; set;
-        }
-
-        /// <summary>
-        /// Tries to get the <see cref="LoRaRegion"/> based on <paramref name="value"/>
-        /// </summary>
-        public static bool TryGetRegion(LoRaRegion value, out Region region)
```
debugerr: I would make this a simple TranslateToRegion or something like that. You not really trying, as the enum is specified and there are only 2 values.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -399,5 +260,17 @@ public int GetDRFromFreqAndChan(string datr)
         {
             return (int)this.DRtoConfiguration.FirstOrDefault(x => x.Value.configuration == datr).Key;
         }
+
+        public bool IsValidRX1DROffset(uint rx1DrOffset) => rx1DrOffset >= 0 && rx1DrOffset <= this.RX1DROffsetTable.GetUpperBound(1);
+
+        public bool IsValidRXDelay(ushort desiredRXDelay)
+        {
+                if (desiredRXDelay >= 0 && desiredRXDelay <= MAX_RX_DELAY)
```
debugerr: wrong indenting and you could change this to

return (desiredRXDelay >= 0 && desiredRXDelay <= MAX_RX_DELAY)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/RegionImplementationTest.cs
```diff
@@ -24,8 +24,9 @@ public class RegionImplementationTest
             // in EU in standard parameters the expectation is that the reply arrive at same place.
             var expFreq = freq;
             var expDatr = datr;
-            Assert.Equal(RegionFactory.CreateEU868Region().GetDownstreamChannelFrequency(rxpk[0]), expFreq);
-            Assert.Equal(RegionFactory.CreateEU868Region().GetDownstreamDR(rxpk[0]), expDatr);
+            RegionManager.EU868.TryGetDownstreamChannelFrequency(rxpk[0], out double frequency);
```
debugerr: you have multiple of those. If you call a try, you should likely validate the return value of that (not going to add that comment all over :))

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/LoRaRegionEnum.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace LoRaTools.Regions
 {
-    public enum LoRaRegion
+    public enum LoRaRegionEnum
     {
```
debugerr: I don't think we should be naming an enum Enum. Can we call it LoRaRegion?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -154,7 +154,10 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
                 if (windowToUse == Constants.RECEIVE_WINDOW_1)
                 {
                     datr = loraRegion.GetDownstreamDR(request.Rxpk);
-                    freq = loraRegion.GetDownstreamChannelFrequency(request.Rxpk);
+                    if (!loraRegion.TryGetDownstreamChannelFrequency(request.Rxpk, out freq) || datr == null)
```
debugerr: that's a bit weird. If you do not get a valid frequency, can you continue?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -184,30 +187,38 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
...
+                {
+                    rxDelay = loRaDevice.DesiredRXDelay;
+                }
+                else
+                {
+                    Logger.Log(devEUI, $"twin RX Delay value are not within acceptable values", LogLevel.Error);
+                }
+
+                var loRaPayloadJoinAccept = new LoRaTools.LoRaMessage.LoRaPayloadJoinAccept(
+               LoRaTools.Utils.ConversionHelper.ByteArrayToString(netId), // NETID 0 / 1 is default test
```
debugerr: indent is wrong here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
debugerr: Indent is wrong.
Can be changed to 
return desiredRXDelay >= 0 && desiredRXDelay <= MAX_RX_DELAY;

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs
Mandur: good point, I will change the naming. Keep in mind that the enum will grow to ~10 values.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/RegionImplementationTest.cs
Mandur: Correct, I did that in the normal process, As those are in the tests, I assumed I would let it like this, but I will wrap those with an Assert.True

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/LoRaRegionEnum.cs
Mandur: Well, I changed the name due to the fact that there is another class named Region, and I always get mixed between the two, at least with this it is clear. Open for suggestion and/or best practices here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
Mandur: Shouldn't Stylecop detect this? :/

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
Mandur: Thank you for the comment, this is a bug. The same occur inDefaultDataRequestHandler, I tried to correct it

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultClassCDevicesMessageSender.cs
```diff
@@ -51,7 +51,7 @@ public async Task<bool> SendAsync(IReceivedLoRaCloudToDeviceMessage cloudToDevic
                     return false;
                 }
 
-                if (!RegionFactory.TryGetRegion(loRaDevice.LoRaRegion, out var region))
+                if (!RegionManager.TranslateToRegion(loRaDevice.LoRaRegion, out var region))
```
fbeltrao: If it is a bool TranslateToRegion(param1, out param2) it should be `TryTranslateToRegion` imo. It can fail (Region.NotSet is used for instance)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -77,7 +77,12 @@ internal static class DownlinkMessageBuilder
             else
             {
                 datr = loRaRegion.GetDownstreamDR(rxpk, (uint)loRaDevice.ReportedRX1DROffset);
-                freq = loRaRegion.GetDownstreamChannelFrequency(rxpk);
+                if (!loRaRegion.TryGetDownstreamChannelFrequency(rxpk, out freq) && datr != null)
```
fbeltrao: Here will continue if we cannot get a frequency and datr. Shouldn't it stop if any of those fail (or)?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
```diff
@@ -51,15 +51,15 @@ public void DispatchRequest(LoRaRequest request)
 
             if (this.loraRegion == null)
             {
-                if (!RegionFactory.TryResolveRegion(request.Rxpk))
+                if (!RegionManager.TryResolveRegion(request.Rxpk))
```
fbeltrao: I would recommend using the overloaded that is not based on the static. This way we can one day remove the static `CurrentRegion`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/LoRaRegionEnum.cs
fbeltrao: Not a big fan of calling enum XxxEnums too. Maybe LoRaRegionType.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -234,16 +121,19 @@ public Region(LoRaRegion regionEnum, byte loRaSyncWord, byte[] gFSKSyncWord, (do
...
         /// <param name="upstreamChannel">the channel at which the message was transmitted</param>
-        public double GetDownstreamChannelFrequency(Rxpk upstreamChannel)
+        public bool TryGetDownstreamChannelFrequency(Rxpk upstreamChannel, out double frequency)
         {
+            frequency = 0;
+
             if (this.IsValidRxpk(upstreamChannel))
             {
-                if (this.LoRaRegion == LoRaRegion.EU868)
+                if (this.LoRaRegion == LoRaRegionEnum.EU868)
```
fbeltrao: Food for thought: shouldn't this kind part of each Region instance, defined during creation? This method will grow for each new region we support.

A possible solution would be to extract this into a class. For instance:

```c#
// create eu region
...
euRegion.DownstreamChannelFrequencyResolver = new SameAsUpstreamChannelFrequencyResolver();

// create us
usRegion.DownstreamChannelFrequencyResolver = new USUpstreamChannelFrequencyResolver();
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
Mandur: I am not sure I understand, I am returning an empty object that should be ignored by the receiver. What changes do you intend?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
Mandur: Do you mean request.Region ?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
fbeltrao: I think the intend is to stop processing if we could not find a datr or frequency. It looks to me that the code will only stop if we cannot find a frequency **but found a datr**.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
fbeltrao: Yes, set request.Region to a local variable instead of static Current.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
Mandur: I think the logic is indeed wrong, I changed the && to a or

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
ronniesa: this is odd should stylecop detect this for sure....

PR #150, opened 2019-03-20
Fixed configuration of SPIDEV.

gerfen:
Needed to flip the key and the value to properly set SPIDEV from the .env file.


Mandur:
Thank you for spotting it, this looks indeed correct!

PR #154, opened 2019-03-22
Update IoT SDK to v1.19.0

fbeltrao:
Updating IoT SDK to v1.19.0 from v1.18.0


fbeltrao:
Reverting to SDK 1.18.1 due to timeout bug in DeviceClient.ReceiveAsync

PR #156, opened 2019-03-22
Fine tune device caching

fbeltrao:
Changes:
* Absolute device cache expiration
* Device join attempt cache increased to 30 min (from 2 min)
* Update AzureDevOpsAgent to 2.148.2

Fixes [AB#666](https://dev.azure.com/epicstuff/web/wi.aspx?pcguid=b4ad2b30-c669-4749-9c24-e42a8da5d35d&id=666), [AB#1165](https://dev.azure.com/epicstuff/web/wi.aspx?pcguid=b4ad2b30-c669-4749-9c24-e42a8da5d35d&id=1165)

### DIFF
--- a/LoRaEngine/test/LoRaWan.IntegrationTest/IntegrationTestFixtureCI.cs
+++ b/LoRaEngine/test/LoRaWan.IntegrationTest/IntegrationTestFixtureCI.cs
@@ -122,6 +122,28 @@ namespace LoRaWan.IntegrationTest
             }
         }
 
+        public string GetKey16(int deviceId)
+        {
+            var format = string.IsNullOrEmpty(this.Configuration.DeviceKeyFormat) ? "0000000000000000" : this.Configuration.DeviceKeyFormat;
+            if (format.Length < 16)
+            {
+                format = format.PadLeft(16, '0');
+            }
+
+            return deviceId.ToString(format);
+        }
+
+        public string GetKey32(int deviceId)
+        {
+            var format = string.IsNullOrEmpty(this.Configuration.DeviceKeyFormat) ? "00000000000000000000000000000000" : this.Configuration.DeviceKeyFormat;
+            if (format.Length < 32)
+            {
+                format = format.PadLeft(32, '0');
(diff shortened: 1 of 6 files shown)


fbeltrao:
Forgot to make the test device key changes...

fbeltrao:
@ronniesa test OTAA_Join_With_Valid_Device_Updates_DeviceTwin failed once in CI. Could not read the information from IoT Hub on time. The properties were there as the EventHub log shows.

LoRaEngine/test/LoRaWan.IntegrationTest/IntegrationTestFixtureCI.cs
```diff
@@ -122,26 +122,48 @@ protected override void Dispose(bool disposing)
...
+            {
+                format = format.PadLeft(16, '0');
+            }
+
+            return deviceId.ToString(format);
+        }
+
+        public string GetKey32(int deviceId)
+        {
+            var format = string.IsNullOrEmpty(this.Configuration.DeviceKeyFormat) ? "00000000000000000000000000000000" : this.Configuration.DeviceKeyFormat;
```
TechPreacher: Might become a bit more readable and less prone to error if instead of "00000000000000000000000000000000" you would use new String('0', 32)

fbeltrao: That is true @TechPreacher. Will make this change in a next PR. Thanks!

TechPreacher: Or maybe simplify the 2 methods to:

public string GetKey(int deviceId, int length)
{
    var format = string.IsNullOrEmpty(this.Configuration.DeviceKeyFormat) ? new string('0', length) : this.Configuration.DeviceKeyFormat;
    if (format.Length < length)
    {
        format = format.PadLeft(length, '0');
    }

    return deviceId.ToString(format);
}

PR #163, opened 2019-03-27
Prepare template for release

Mandur:
Prepare template and pin the Azure function version

### DIFF
--- a/Template/azuredeploy.json
+++ b/Template/azuredeploy.json
@@ -46,7 +46,7 @@
         "storageAccountType": "Standard_LRS",
         "functionAppName": "[concat(parameters('uniqueSolutionPrefix'), 'function')]",
         "gitUsername": "Azure",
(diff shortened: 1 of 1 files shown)


Template/azuredeploy.json
```diff
@@ -129,7 +129,7 @@
                         },
                         {
                             "name": "FUNCTIONS_EXTENSION_VERSION",
-                            "value": "~2"
+                            "value": "2.0.12342.0"
```
TechPreacher: Do we want to hard code the functions extensions version? I know it works around a current bug but it will also prevent users from future updates...

ronniesa: We are aware but is the only way to make it work right now. Btw this should also be on the dev guide @debugerr can you add a note on the dev guide section please

fbeltrao: To be honest, based on my experience, I would recommend pinning versions for anything going to production. I have seen newer version being updated with regression problems.

I would approach function runtime version the same way we do nuget packages or OS versions: you just don't put on auto update and hope for the best. Proper testing is needed.

PR #167, opened 2019-03-28
merging bug fixes

debugerr:
- fix null ref in ADR execution item
- remove ADR multi GW API calls

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
@@ -84,9 +84,14 @@ namespace LoraKeysManagerFacade
                 string cacheKey = devEUI + devNonce;
                 using (var deviceCache = new LoRaDeviceCache(this.cacheStore, devEUI, gatewayId, cacheKey))
                 {
+                    if (deviceCache.HasValue())
+                    {
+                        throw new DeviceNonceUsedException();
+                    }
+
                     if (deviceCache.TryToLock(cacheKey + "joinlock"))
                     {
-                        if (deviceCache.TryGetValue(out _))
+                        if (deviceCache.HasValue())
                         {
                             throw new DeviceNonceUsedException();
                         }
@@ -106,6 +111,10 @@ namespace LoraKeysManagerFacade
 
                             this.cacheStore.KeyDelete(devEUI);
                         }
+                    }
+                    else
+                    {
+                        throw new DeviceNonceUsedException();
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/ADRExecutionItem.cs
```diff
@@ -19,7 +19,7 @@ public ADRExecutionItem(ILoRaADRManager adrManager)
         public async Task<FunctionBundlerExecutionState> ExecuteAsync(IPipelineExecutionContext context)
         {
             context.Result.AdrResult = await this.HandleADRRequest(context.DevEUI, context.Request.AdrRequest);
-            context.Result.NextFCntDown = context.Result?.AdrResult.FCntDown > 0 ? context.Result.AdrResult.FCntDown : (uint?)null;
+            context.Result.NextFCntDown = context.Result.AdrResult != null && context.Result.AdrResult.FCntDown > 0 ? context.Result.AdrResult.FCntDown : (uint?)null;
```
Mandur: Isn't that similar?

debugerr: Don't think so, right? The first one was testing if Result was null and the new one does validate AdrResult.

fbeltrao: PSA: I shot myself in the foot with a bug using nullables. They are dangerous for == and != comparison.
```csharp
int? a = null;
if (a == 0) {
    // will not go in since null != 0
}

if (a != 0) {
    // will go in, since 0 != null
}
```

What you probably want is:

```csharp
if ((a ?? 0) == 0) {
    // will go in if a == null || a == 0
}

if ((a ?? 0) != 0) {
    // will go in only if 'a' has a value and is not zero
}
```

fbeltrao: So `.Result` can never be null, right?

Mandur: Correct I missed a ? :P

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -84,9 +84,14 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
                 string cacheKey = devEUI + devNonce;
                 using (var deviceCache = new LoRaDeviceCache(this.cacheStore, devEUI, gatewayId, cacheKey))
                 {
+                    if (deviceCache.HasValue())
```
fbeltrao: Just to add context, that was a bug that caused the function to not return "dev nonce already used" if the loser would not get the lock.

LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/ADRExecutionItem.cs
debugerr: @fbeltrao Correct, .Result can never be null. It has a property initializer. The null ref came from AdrResult. Here the FCntDown (on the AdrResult) is not a nullable. Should all be ok here.

PR #174, opened 2019-04-02
fix server races

debugerr:
fixes: 
- deduplication race with device reset (ABP relaxed mode)
- fix OTAA join when having a gateway in the twin
- add a blocking lock to the cache

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
@@ -93,7 +93,7 @@ namespace LoraKeysManagerFacade
                     var cacheKeyDevNonce = string.Concat(devEUI, ":", devNonce);
                     var lockKeyDevNonce = string.Concat(cacheKeyDevNonce, ":joinlockdevnonce");
 
-                    if (await this.cacheStore.LockTakeAsync(lockKeyDevNonce, gatewayId, TimeSpan.FromSeconds(10)))
+                    if (await this.cacheStore.LockTakeAsync(lockKeyDevNonce, gatewayId, TimeSpan.FromSeconds(10), block: false))
                     {
                         try
                         {
@@ -199,23 +199,24 @@ namespace LoraKeysManagerFacade
                         this.cacheStore.ObjectSet(cacheKeyJoinInfo, joinInfo, TimeSpan.FromMinutes(60));
                         log?.LogDebug("updated cache with join info '{key}':{gwid}", devEUI, gatewayId);
                     }
-
-                    if (string.IsNullOrEmpty(joinInfo.PrimaryKey))
-                    {
-                        throw new JoinRefusedException("Not in our network.");
-                    }
-
-                    if (!string.IsNullOrEmpty(joinInfo.DesiredGateway) && gatewayId != joinInfo.DesiredGateway)
-                    {
-                        throw new JoinRefusedException("Not the owning gateway");
-                    }
-
-                    log?.LogDebug("got LogInfo '{key}':{gwid} attached gw: {desiredgw}", devEUI, gatewayId, joinInfo.DesiredGateway);
                 }
                 finally
                 {
                     var released = this.cacheStore.LockRelease(lockKeyJoinInfo, gatewayId);
                 }
+
+                if (string.IsNullOrEmpty(joinInfo.PrimaryKey))
+                {
+                    throw new JoinRefusedException("Not in our network.");
+                }
+
+                if (!string.IsNullOrEmpty(joinInfo.DesiredGateway) &&
+                    !joinInfo.DesiredGateway.Equals(gatewayId, StringComparison.OrdinalIgnoreCase))
+                {
+                    throw new JoinRefusedException($"Not the owning gateway. Owning gateway is '{joinInfo.DesiredGateway}'");
+                }
+
+                log?.LogDebug("got LogInfo '{key}':{gwid} attached gw: {desiredgw}", devEUI, gatewayId, joinInfo.DesiredGateway);
             }
             else
             {
--- a/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
@@ -85,25 +85,18 @@ namespace LoraKeysManagerFacade
 
             if (devEUI != null)
             {
+                var joinInfo = await this.TryGetJoinInfoAndValidateAsync(devEUI, gatewayId, log);
+
                 // OTAA join
                 using (var deviceCache = new LoRaDeviceCache(this.cacheStore, devEUI, gatewayId))
                 {
-                    var joinInfo = await this.GetJoinInfoAsync(devEUI, gatewayId, log);
-
                     var cacheKeyDevNonce = string.Concat(devEUI, ":", devNonce);
                     var lockKeyDevNonce = string.Concat(cacheKeyDevNonce, ":joinlockdevnonce");
 
-                    if (await this.cacheStore.LockTakeAsync(lockKeyDevNonce, gatewayId, TimeSpan.FromSeconds(10), block: false))
+                    if (this.cacheStore.StringSet(cacheKeyDevNonce, devNonce, TimeSpan.FromMinutes(5), onlyIfNotExists: true))
                     {
                         try
                         {
-                            var freshValue = this.cacheStore.StringSet(cacheKeyDevNonce, devNonce, TimeSpan.FromMinutes(5), onlyIfNotExists: true);
-                            if (!freshValue)
-                            {
-                                log?.LogDebug("dev nonce already used. Ignore request '{key}':{gwid}", devEUI, gatewayId);
-                                throw new DeviceNonceUsedException();
-                            }
-
                             var iotHubDeviceInfo = new IoTHubDeviceInfo
                             {
                                 DevEUI = devEUI,
@@ -129,6 +122,7 @@ namespace LoraKeysManagerFacade
                     }
                     else
                     {
+                        log?.LogDebug("dev nonce already used. Ignore request '{key}':{gwid}", devEUI, gatewayId);
                         throw new DeviceNonceUsedException();
                     }
                 }
@@ -169,7 +163,7 @@ namespace LoraKeysManagerFacade
             return results;
         }
 
-        private async Task<JoinInfo> GetJoinInfoAsync(string devEUI, string gatewayId, ILogger log)
+        private async Task<JoinInfo> TryGetJoinInfoAndValidateAsync(string devEUI, string gatewayId, ILogger log)
         {
             var cacheKeyJoinInfo = string.Concat(devEUI, ":joininfo");
             var lockKeyJoinInfo = string.Concat(devEUI, ":joinlockjoininfo");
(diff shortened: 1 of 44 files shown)


LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -153,5 +168,61 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
...
             return results;
         }
+
+        private async Task<JoinInfo> GetJoinInfoAsync(string devEUI, string gatewayId, ILogger log)
+        {
+            var cacheKeyJoinInfo = string.Concat(devEUI, ":joininfo");
+            var lockKeyJoinInfo = string.Concat(devEUI, ":joinlockjoininfo");
+            JoinInfo joinInfo = null;
+
+            if (await this.cacheStore.LockTakeAsync(lockKeyJoinInfo, gatewayId, TimeSpan.FromMinutes(5)))
```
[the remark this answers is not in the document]
fbeltrao: Should this LockTakeAsync block be `true`? If not, racing GetJoins will return a null object during the instance creation. That could be the case when the correct gatewayID is unable to adquire the join info lock.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -153,5 +168,61 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
...
+                        this.cacheStore.ObjectSet(cacheKeyJoinInfo, joinInfo, TimeSpan.FromMinutes(60));
+                        log?.LogDebug("updated cache with join info '{key}':{gwid}", devEUI, gatewayId);
+                    }
+
+                    if (string.IsNullOrEmpty(joinInfo.PrimaryKey))
+                    {
+                        throw new JoinRefusedException("Not in our network.");
+                    }
+
+                    if (!string.IsNullOrEmpty(joinInfo.DesiredGateway) && gatewayId != joinInfo.DesiredGateway)
```
fbeltrao: Shouldn't we compare gatewayID ignoring case?

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -153,5 +168,61 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
...
+                    {
+                        throw new JoinRefusedException("Not the owning gateway");
```
fbeltrao: Perhaps add the assigned gateway to make troubleshooting easier

LoRaEngine/LoraKeysManagerFacade/FCntCacheCheck.cs
```diff
@@ -42,41 +43,63 @@ public FCntCacheCheck(ILoRaDeviceCacheStore deviceCache)
...
             if (!string.IsNullOrEmpty(abpFcntCacheReset))
             {
-                this.deviceCache.KeyDelete(devEUI);
+                using (var deviceCache = new LoRaDeviceCache(this.deviceCache, devEUI, gatewayId))
+                {
+                    if (await deviceCache.TryToLockAsync())
+                    {
+                        if (deviceCache.TryGetInfo(out var deviceInfo))
+                        {
+                            // only reset the cache if the current valud is larger
```
fbeltrao: Small typo on comment `valud` &rarr; `value`

`continued processing` &rarr; `continue processing`

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
```diff
@@ -36,27 +37,28 @@ public LoRaDeviceCache(ILoRaDeviceCacheStore cacheStore, string devEUI, string g
             this.cacheStore = cacheStore;
             this.devEUI = devEUI;
             this.gatewayId = gatewayId;
-            this.cacheKey = cacheKey ?? devEUI;
+            this.cacheKey = devEUI;
         }
 
-        public bool TryToLock(string lockKey = null)
+        public async Task<bool> TryToLockAsync(string lockKey = null, bool block = true)
```
fbeltrao: I find the naming confusing, plus we changed the default implementation to "wait until lock is obtained". Here block is true by default where in RedisStore is false.

Perhaps split in two methods:
- TryToLock or TryToLockOnce that only goes the happy path, a single attempt
- WaitForLock or EnsureHasLock that waits until the lock is obtained

LoRaEngine/LoraKeysManagerFacade/FCntCacheCheck.cs
Mandur: value typo

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -55,7 +55,7 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
                 if (loRaDevice == null)
                 {
                     request.NotifyFailed(devEUI, LoRaDeviceRequestFailedReason.UnknownDevice);
-                    Logger.Log(devEUI, "join refused: unknown device", LogLevel.Information);
+                    Logger.Log(devEUI, "join refused", LogLevel.Information);
```
fbeltrao: Isn't that a duplicated log from LoRaDeviceRegistry line (289)?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
```diff
@@ -10,13 +10,13 @@ namespace LoRaTools
 
     public class OTAAKeysGenerator
     {
+        private static readonly Random RndDevAddr = new Random();
```
fbeltrao: Better than before, but I think we could have problems using shared object in multiple threads.

https://devblogs.microsoft.com/pfxteam/getting-random-numbers-in-a-thread-safe-way/

https://stackoverflow.com/questions/3049467/is-c-sharp-random-number-generator-thread-safe

LoRaEngine/LoraKeysManagerFacade/FCntCacheCheck.cs
```diff
@@ -42,41 +43,63 @@ public FCntCacheCheck(ILoRaDeviceCacheStore deviceCache)
...
                 {
                     if (deviceCache.TryGetInfo(out DeviceCacheInfo serverStateForDeviceInfo))
                     {
                         newFCntDown = ProcessExistingDeviceInfo(deviceCache, serverStateForDeviceInfo, gatewayId, clientFCntUp, clientFCntDown);
                     }
                     else
                     {
-                        var state = deviceCache.Initialize(clientFCntUp, clientFCntDown + 1);
-                        newFCntDown = state.FCntDown;
+                        newFCntDown = clientFCntDown + 1;
```
Mandur: Is there a reason for this change? The code looks similar to me.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
Mandur: Is static random considered as best practice? Should we think about using the same one across the app or that would generate a similar random value?

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
ronniesa: yes

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
ronniesa: If we have it yes would be nice

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
ronniesa: You thought you would sneak it in this one here..... ah ah.... but no we keep it like this.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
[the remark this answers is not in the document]
debugerr: Default is to block. We can chat about the additional get.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
debugerr: allright. fixing this.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
debugerr: added

LoRaEngine/LoraKeysManagerFacade/FCntCacheCheck.cs
debugerr: continued is right here. continued refers to the initial resetter and that carried on (continued) processing and has added it again.

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
debugerr: I'll update to default to true for all. The idea was to have the DeviceCache default to true, but others false (as that seemed to be the general perception of the existing callers). However, I think in general, acquiring a lock, at least in my understanding, should block and wait to get the lock.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
debugerr: It is. I thought we talked about that. I find it difficult to ditch this as we would make an assumption about the implementation of the device registry component logging that particular one. We should not do that, since if the implementation changes, we do not get a log.
The one in the device registry is now more specific.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
debugerr: Ok. Didn't know about the fact that this thing is not thread safe. Doesn't make much sense, as it will be used as a static almost everywhere (if you want to make sure you get unique numbers for sure). I was sceptical that this was still a problem in the current framework implementations (Jave btw. does properly lock it apparently), but I was able to repro and get it into a state where it only returned 0. I'll add a lock to protect from that, but will keep the static.

LoRaEngine/LoraKeysManagerFacade/FCntCacheCheck.cs
debugerr: Fixed. Thanks.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
debugerr: The problem with using a new instance everytime is that the implementation uses the system timer as a seed. That resolution is not high enough to guarantee unique numbers if you hit that code very quickly, you could end up with the same value. 
In general, we could have a wrapper class that is using a static and is using a lock internally to provide the values.

PR #175, opened 2019-04-03
CLI Device Provisioning Tool: bulkverify and new twin options: RXDelay and KeepAliveTimeout

TechPreacher:
- Added new verb: bulkverify to verify all devices in IoT Hub with LoRa device twin configuration.
- Added RXDelay and KeepAliveTimeout twin properties.
- Refactored logging to allow for verbose/non-verbose logging between modes (verify / bulkverify)
- Consolidated various duplicate code blocks into methods
- Moved from synchronous Main() to new, async Main() in console app.
- Updated docs.
- Moved version to 1.0.0.0 in sync with main project.


.gitignore
```diff
@@ -296,3 +296,5 @@ Doc
 appsettings.local.json
 Docs/.DS_Store
 .DS_Store
+
+Tools/Cli-LoRa-Device-Provisioning/settings\.local\.json
```
fbeltrao: We should generalize local settings file names to avoid adding multiple .gitignore files. I can see there is only already to `appsetings.local.json`.
My suggestion is to either use the same name or change the gitignore rule to *.local.json

Tools/Cli-LoRa-Device-Provisioning/Helpers/ConfigurationHelper.cs
```diff
@@ -26,34 +26,35 @@ public bool ReadConfig()
                 var configurationBuilder = new ConfigurationBuilder()
                     .SetBasePath(Directory.GetCurrentDirectory())
                     .AddJsonFile("settings.json", optional: false, reloadOnChange: true)
+                    .AddJsonFile("settings.local.json", optional: true, reloadOnChange: true)
```
fbeltrao: If you don't use IOptionsMonitor or [IOptionsSnapshot](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-2.2#reload-configuration-data-with-ioptionssnapshot) setting `reloadOnChange: true` does nothing. It is just an overhead.

I would recommend setting to false. What are the chances that someone invokes the cmd line and changes the config during the execution?

Tools/Cli-LoRa-Device-Provisioning/Helpers/IoTDeviceHelper.cs
```diff
@@ -13,7 +13,8 @@ namespace LoRaWan.Tools.CLI.Helpers
 
     public class IoTDeviceHelper
     {
-        private string[] classTypes = { "A", "C" };
+        private readonly string[] classTypes = { "A", "C" };
```
fbeltrao: If you are using as constants I recommend changing to static. Thus preventing a new value being created every time a new IoTDeviceHelper is instanciated.

.gitignore
TechPreacher: I am happy to switch. 
Should I change both files to appsettings.json and appsettings.local.json? Makes sense to have consistency.
Then undo the .gitignore changes...

Tools/Cli-LoRa-Device-Provisioning/Helpers/ConfigurationHelper.cs
TechPreacher: Yeah. That's a silly setting for a command line tool.

.gitignore
fbeltrao: I think that is a good approach, specially because Azure Functions takes `appsettings.local.json` by default.

Tools/Cli-LoRa-Device-Provisioning/Helpers/IoTDeviceHelper.cs
TechPreacher: Will do!

.gitignore
TechPreacher: Done!

PR #177, opened 2019-04-04
Fix reported properties sync bug

Mandur:
Our current implementation of the additional join properties was not syncing the reported properties across different gateway. Hence it was non working in multi gateway scenarios.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
@@ -572,9 +572,9 @@ namespace LoRaWan.NetworkServer
                 var fcntDownDelta = this.FCntDown >= this.LastSavedFCntDown ? this.FCntDown - this.LastSavedFCntDown : this.LastSavedFCntDown - this.FCntDown;
 
                 if (reportedProperties.Count > 0 ||
-                            fcntDownDelta >= Constants.MAX_FCNT_UNSAVED_DELTA ||
-                            fcntUpDelta >= Constants.MAX_FCNT_UNSAVED_DELTA ||
-                            (this.hasFrameCountChanges && force))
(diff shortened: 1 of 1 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -560,9 +576,9 @@ public async Task<bool> SaveChangesAsync(TwinCollection reportedProperties = nul
                 var fcntDownDelta = this.FCntDown >= this.LastSavedFCntDown ? this.FCntDown - this.LastSavedFCntDown : this.LastSavedFCntDown - this.FCntDown;
 
                 if (reportedProperties.Count > 0 ||
-                    fcntDownDelta >= Constants.MAX_FCNT_UNSAVED_DELTA ||
-                    fcntUpDelta >= Constants.MAX_FCNT_UNSAVED_DELTA ||
-                    (this.hasFrameCountChanges && force))
+                            fcntDownDelta >= Constants.MAX_FCNT_UNSAVED_DELTA ||
+                            fcntUpDelta >= Constants.MAX_FCNT_UNSAVED_DELTA ||
+                            (this.hasFrameCountChanges && force))
```
debugerr: I prefer the othe rstyle where things are aligned with the first line

TechPreacher:
Looks good!

fbeltrao:
My feedback is to provide a description of what this PR is supposed to do. Is the problem that we handle RX delay properties in all devices types, instead of OTAA only?

ronniesa:
> My feedback is to provide a description of what this PR is supposed to do. Is the problem that we handle RX delay properties in all devices types, instead of OTAA only?

No there was a bug that on first message we did not read the reported twin for the RX delay. So the cache version after a join works but in case of a ns restart or a multigateway it would fail.

Mandur:
I forced push this branch to clean up history. It is now ready for test and review.

PR #178, opened 2019-04-04
Fixes for ADR caching and FCnt down in functions

debugerr:
- fix adr caching (redis locks)
- ensure nextfcnt down only runs once

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
@@ -14,27 +14,36 @@ namespace LoraKeysManagerFacade
 
     public class LoRaADRRedisStore : LoRaADRStoreBase, ILoRaADRStore
     {
-        const string CacheToken = "ADR";
-        const string LockToken = "_lock";
+        const string CacheToken = ":ADR";
+        const string LockToken = ":lock";
         IDatabase redisCache;
 
         sealed class RedisLockWrapper : IDisposable
         {
+            private static readonly TimeSpan LockTimeout = TimeSpan.FromSeconds(10);
             private string lockKey;
             private string owner;
             private IDatabase redisCache;
             private bool ownsLock;
 
-            internal RedisLockWrapper(string devEUI, IDatabase redisCache, string owner = "_LoRaRedisStore")
+            internal RedisLockWrapper(string devEUI, IDatabase redisCache, string owner = ":LoRaRedisStore")
             {
                 this.lockKey = GetEntryKey(devEUI) + LockToken;
                 this.redisCache = redisCache;
                 this.owner = owner;
             }
 
-            internal bool TakeLock()
+            internal async Task<bool> TakeLockAsync()
             {
-                return this.ownsLock = this.redisCache.LockTake(this.lockKey, this.owner, TimeSpan.FromSeconds(10));
+                var start = DateTime.UtcNow;
+                while (!(this.ownsLock = await this.redisCache.LockTakeAsync(this.lockKey, this.owner, TimeSpan.FromSeconds(10))))
+                {
+                    if (DateTime.UtcNow - start > LockTimeout)
+                        break;
+                    await Task.Delay(100);
+                }
+
+                return this.ownsLock;
             }
 
             public void Dispose()
@@ -56,7 +65,7 @@ namespace LoraKeysManagerFacade
         {
             using (var redisLock = new RedisLockWrapper(devEUI, this.redisCache))
             {
-                if (redisLock.TakeLock())
+                if (await redisLock.TakeLockAsync())
                 {
                     await this.redisCache.StringSetAsync(GetEntryKey(devEUI), JsonConvert.SerializeObject(table));
                 }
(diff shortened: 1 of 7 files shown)


LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
```diff
@@ -14,27 +14,36 @@ namespace LoraKeysManagerFacade
...
-            internal bool TakeLock()
+            internal async Task<bool> TakeLockAsync()
             {
-                return this.ownsLock = this.redisCache.LockTake(this.lockKey, this.owner, TimeSpan.FromSeconds(10));
+                var start = DateTime.UtcNow;
+                while (!(this.ownsLock = await this.redisCache.LockTakeAsync(this.lockKey, this.owner, TimeSpan.FromSeconds(10))))
+                {
+                    if (DateTime.UtcNow - start > LockTimeout)
+                        break;
+                    await Task.Delay(100);
```
TechPreacher: Are these timings going to be fixed or could they maybe be declared as constants?

LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
```diff
@@ -14,27 +14,36 @@ namespace LoraKeysManagerFacade
...
                 this.redisCache = redisCache;
                 this.owner = owner;
             }
 
-            internal bool TakeLock()
+            internal async Task<bool> TakeLockAsync()
             {
-                return this.ownsLock = this.redisCache.LockTake(this.lockKey, this.owner, TimeSpan.FromSeconds(10));
+                var start = DateTime.UtcNow;
+                while (!(this.ownsLock = await this.redisCache.LockTakeAsync(this.lockKey, this.owner, TimeSpan.FromSeconds(10))))
```
TechPreacher: Same for the fixed timings.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -141,8 +146,6 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
                     // In multi gateway scenarios it means the another gateway was faster than using, can stop now
                     if (fcntDown <= 0)
                     {
```
fbeltrao: Did you remove the logging because it is done somewhere else?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -576,14 +579,17 @@ void ProcessAndSendMacCommands(LoRaPayloadData payloadData, ref Dictionary<strin
...
+                // if it was previously calculated, use that value
                 return fcntDown.Value;
+            }
 
             var newFcntDown = await frameCounterStrategy.NextFcntDown(loRaDevice, payloadFcnt);
 
             // Failed to update the fcnt down
             // In multi gateway scenarios it means the another gateway was faster than using, can stop now
-            if (newFcntDown <= 0)
+            if (newFcntDown > 0)
```
fbeltrao: Shouldn't this be the opposite? If the newFcntDown > 0 means we have a valid frame down counter. If it is <= 0 then is invalid.

LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
debugerr: The delay you mean? That's really just a loop delay. We can make the lock duration a constant. Not sure if you meant that.

Generally: I'm not 100% happy about the code at the moment. The problem is that we have 2 wrappers around redis cache, which is harder to maintain. That's also why we didn't see this locking issue when we fixed it in the first wrapper. Something we should cleanup if we get a chance.

fbeltrao: I second that, the Redis wrapper are confusing. We should revisit whenever possible.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
debugerr: Yes. This was a duplicate logging event.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
debugerr: Good catch. Thanks. Would have resulted in a wrong debug message.

debugerr:
Added a backlog item to track refactoring the redis store wrappers: 1206

PR #199, opened 2019-07-03
Add SPI Speed option for pkt fwd

ZhuQI77:
modify spi speed for rak2245.

### DIFF
--- a/LoRaEngine/example.env
+++ b/LoRaEngine/example.env
@@ -24,6 +24,8 @@ EDGEHUB_ROUTE='FROM /* INTO $upstream'
 PKT_FWD_VERSION=1.0.0
 #SPI DEV version information. Only needed for x86 gateways. Possible values are 1 or 2.
 PKT_FWD_SPI_DEV=2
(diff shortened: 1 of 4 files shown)


Mandur:
Hello @ZhuQI77 thank you for your contribution and the nice enhancement for the pkt_fwd. I will test to see if any amendment is necessary.

Mandur:
@TechPreacher Could you please test on rak2245 and let us know if current PR works? Thanks

LoRaEngine/example.env
```diff
@@ -24,6 +24,8 @@ EDGEHUB_ROUTE='FROM /* INTO $upstream'
 PKT_FWD_VERSION=1.0.0
 #SPI DEV version information. Only needed for x86 gateways. Possible values are 1 or 2.
 PKT_FWD_SPI_DEV=2
+#set a custom speed for your SPI
```
TechPreacher: I would add for clarity:
#set custom speed for your SPI or, if not needed, comment or remove PKT_FWD_SPI_DEV.

PR #201, opened 2019-07-05
Enable Spi speed and Spi dev on arm template

Mandur:
& prepare for release 1.0.1

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
@@ -40,6 +40,8 @@ namespace LoraKeysManagerFacade
             queryStrings.TryGetValue("publishingPassword", out string publishingPassword);
             queryStrings.TryGetValue("region", out string region);
             queryStrings.TryGetValue("resetPin", out string resetPin);
+            queryStrings.TryGetValue("spiSpeed", out string spiSpeed);
+            queryStrings.TryGetValue("spiDev", out string spiDev);
 
             bool.TryParse(Environment.GetEnvironmentVariable("DEPLOY_DEVICE"), out bool deployEndDevice);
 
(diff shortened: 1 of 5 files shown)


LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
```diff
@@ -146,6 +148,40 @@ public CreateEdgeDevice(RegistryManager registryManager)
...
+                string.IsNullOrEmpty(spiSpeed))
+            {
+                // default case
+                json = json.Replace("[$spi_speed]", string.Empty);
+            }
+            else
+            {
+                json = json.Replace(
+                    "[$spi_speed]",
+                    ",'SPI_SPEED':{'value':'2'}");
```
debugerr: This seems weird. Why would this be a constant 2? Looks like the passed in spiSpeed is never used and should be used here?

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
```diff
@@ -146,6 +148,40 @@ public CreateEdgeDevice(RegistryManager registryManager)
...
+                string.IsNullOrEmpty(spiDev))
+            {
+                // default case
+                json = json.Replace("[$spi_dev]", string.Empty);
+            }
+            else
+            {
+                json = json.Replace(
+                    "[$spi_dev]",
+                    ",'SPI_DEV':{'value':'1'}");
```
debugerr: same with spiDev. Right now, we never add the passed in value to the json string. If it is 2 the result will be empty, in all other cases it will be one, no matter what value was passed in. If that's really the desired behavior, it needs comments.

Template/azuredeploy.json
```diff
@@ -36,6 +36,28 @@
             "metadata": {
                 "description": "In what region is your gateway deployed?"
             }
+        },
+        "spiSpeed": {
+            "type": "int",
+            "allowedValues": [
```
debugerr: ok, from here it really looks like 8 and 2 are the only allowed values... and the same for spiDev down there, 1 & 2

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
debugerr: Looking at the ARM template further down, it looks like there are only 2 values allowed. Need to clarify.

PR #203, opened 2019-07-08
1355 Add Hex decoder to built-in decoders

TechPreacher:
- Added DecoderHexSensor to LoRaPayloadDecoder.cs
- Added Unit tests to LoRaPayloadDecoderTest.cs
- Updated quickstart.md to show use of Hex decoder

Fixes: [AB#1355](https://dev.azure.com/epicstuff/7acd4d9f-4949-49e3-ada7-2485edc00c16/_workitems/edit/1355)

### DIFF
--- a/Docs/quickstart.md
+++ b/Docs/quickstart.md
@@ -217,7 +217,7 @@ If you want the raw decrypted payload to be sent to IoT Hub as **Hex encoded** v
 }
 ```
 
-To add the sample `"DecoderValueSensor"` or `"DeoderHexSensor`" to the sample LoRa device configured above, change it's desired properties in IoT Hub as follows for option 1:
+To add the sample `"DecoderValueSensor"` or `"DecoderHexSensor`" to the sample LoRa device configured above, change it's desired properties in IoT Hub as follows for option 1:
 
 ```json
 "desired": {
--- a/Docs/quickstart.md
+++ b/Docs/quickstart.md
@@ -189,7 +189,7 @@ This section provide a brief description of the device reported properties you c
 
 ### Decoders
 
-The SensorDecoder tag is used to define which method will be used to decode the LoRa payload. If you leave it out or empty it will send the raw decrypted payload in the data field of the json message as Base64 encoded value to IoT Hub.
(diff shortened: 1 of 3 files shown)


Docs/quickstart.md
```diff
@@ -207,7 +207,17 @@ In both cases, we have already provided a simple decoder called `"DecoderValueSe
...
+
+```json
+{
+  .....
+    "data": {"value": "01020408FF"}
+  .....
+}
+```
+
+To add the sample `"DecoderValueSensor"` or `"DeoderHexSensor`" to the sample LoRa device configured above, change it's desired properties in IoT Hub as follows for option 1:
```
fbeltrao: typo in "DeoderHexSensor"

TechPreacher: Fixed! Thanks for noticing.

LoRaEngine/test/LoRaWanNetworkServer.Test/LoRaPayloadDecoderTest.cs
```diff
@@ -156,6 +176,28 @@ public async Task When_Decoder_Is_DecoderValueSensor_Return_In_Value(string deco
             Assert.Equal(expectedValue, decodedPayloadValue.Value);
         }
 
+        [Theory]
+        [InlineData("DecoderHexSensor", "1234", "31323334", 1)]
```
debugerr: Confusing to me to use the fport, as it is ignored all the way in the implementation. I would rather use a fixed fport.

TechPreacher: Should I change this in both the tests for DecoderValueSensor and DecoderHexSensor, @fbeltrao ?

fbeltrao: Maybe it would be interesting to run using different fports to ensure that it returns always the same value.

The included decoders don't use it, but the caller expects it to be there, so we can't remove it from the method signature.

TechPreacher: That's how it's now. The same calls are made using different fports for both DecoderValueSensor and DecoderHexSensor.

PR #207, opened 2019-07-11
Adding retry handling for tests

debugerr:
- adding xUnit test attributes to mark retries within tests (Theory and Fact)
- we currently retry 3 times and have an exponential back off up to 240s max
- even if the test passes after retry, the build will be red - retries are only helping analyzing the problem
- adjusted the pipeline templates to use individual steps when performing integration tests
- enabled retries on multi gw scenarios only


azure-pipeline-integration-test-steps-template.yaml
```diff
@@ -16,38 +17,14 @@ steps:
...
-
-- template: azure-pipeline-single-integration-test-template.yaml
-  parameters:
-    testName: MultiGatewayTests
-
-- template: azure-pipeline-single-integration-test-template.yaml
-  parameters:
-    testName: ClassCTest
-
+- ${{ each test in parameters.testsToRun }}:
```
fbeltrao: Nice one. As you mentioned, we should update the documentation on this!

Issue #225, opened 2020-02-14
AS923 Support


Mandur:
**Is your feature request related to a problem? Please describe.**
Support of AS923 frequency range

**Describe the solution you'd like**
LoRa starter kit should support AS 923 freq range

**Describe alternatives you've considered**
N/A

**Additional context**
Requested by community

jsaez868:
OK, so we figured we need to implement the new region in the RegionManager using https://lora-alliance.org/sites/default/files/2018-04/lorawantm_regional_parameters_v1.1rb_-_final.pdf as reference, correct?

Mandur:
Hello! Thank you for the interest, yes you are Correct. I tried here below to list the changes needed to implement a new region. In case you have questions or you need additional help, I'd be happy to help you out. My skype handle is mandulevrai.

* Add a region enum in [LoraRegion](https://github.com/Azure/iotedge-lorawan-starterkit/blob/master/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/LoRaRegionType.cs)
* Add correct mapping of downstream frequency in [TryGetUpstreamChannelFrequency](https://github.com/Azure/iotedge-lorawan-starterkit/blob/4815a61cc3822ad9a24bd69518f41201b0b26799/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs#L124) .
* most changes need to be done in [RegionManager](https://github.com/Azure/iotedge-lorawan-starterkit/blob/master/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs)
  * We are using upstream frequency of first messages to detect current region. I would suggest changing max us frequencies to 915 and setting as range of 923-925 for the as region. [TryResolveRegion](https://github.com/Azure/iotedge-lorawan-starterkit/blob/4815a61cc3822ad9a24bd69518f41201b0b26799/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs#L49)
  * Create the region similarly to [CreateEU868Region](https://github.com/Azure/iotedge-lorawan-starterkit/blob/4815a61cc3822ad9a24bd69518f41201b0b26799/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs#L89) with adequate settings from the spec. 
[...]

PR #227, opened 2020-03-19
Correct rx2 calculation from twins

Mandur:
#225  revealed current implementation of RX2 desired property lead to a bug that was not detectable in EU freq range as it defaults to 0 but could have impacted other frequency plans where default is not 0.

* Rx2 is now nullable in twins. 
* calculation of rx2 and twins are now split


Mandur:
Abandonning this pr as it seems to increase tests flakiness. i will dive deeper for next release

PR #239, opened 2020-06-25
Fix RX2 implementation in twin

Mandur:
Current implementation had RX2 as non nullable type, therefore the RX2 always defaulted to 0, it was not a problem for EU and US, but it is a problem for other frequency plan we could implement in the future

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/CaseInsensitiveEnvironmentVariables.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/CaseInsensitiveEnvironmentVariables.cs
@@ -49,7 +49,7 @@ namespace LoRaWan.NetworkServer
             return value;
         }
 
-        public double GetEnvVar(string key, double defaultValue)
+        public double? GetEnvVar(string key, double? defaultValue)
         {
             var value = defaultValue;
             if (this.envVars.TryGetValue(key, out var envValue))
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/CaseInsensitiveEnvironmentVariables.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/CaseInsensitiveEnvironmentVariables.cs
@@ -49,7 +49,7 @@ namespace LoRaWan.NetworkServer
             return value;
         }
 
-        public double? GetEnvVar(string key, double? defaultValue)
+        public double? GetEnvVar(string key, double? defaultValue = null)
         {
             var value = defaultValue;
             if (this.envVars.TryGetValue(key, out var envValue))
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
@@ -129,11 +129,7 @@ namespace LoRaWan.NetworkServer
             config.GatewayID = envVars.GetEnvVar("IOTEDGE_DEVICEID", string.Empty);
             config.HttpsProxy = envVars.GetEnvVar("HTTPS_PROXY", string.Empty);
             config.Rx2DataRate = envVars.GetEnvVar("RX2_DATR", string.Empty);
-            if (config.Rx2Frequency.HasValue)
-            {
-                config.Rx2Frequency = envVars.GetEnvVar("RX2_FREQ", config.Rx2Frequency.Value);
-            }
-
+            config.Rx2Frequency = envVars.GetEnvVar("RX2_FREQ", (double?)null);
             config.IoTEdgeTimeout = envVars.GetEnvVar("IOTEDGE_TIMEOUT", config.IoTEdgeTimeout);
             config.FacadeServerUrl = envVars.GetEnvVar("FACADE_SERVER_URL", string.Empty);
             config.FacadeAuthCode = envVars.GetEnvVar("FACADE_AUTH_CODE", string.Empty);
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
@@ -129,7 +129,7 @@ namespace LoRaWan.NetworkServer
             config.GatewayID = envVars.GetEnvVar("IOTEDGE_DEVICEID", string.Empty);
             config.HttpsProxy = envVars.GetEnvVar("HTTPS_PROXY", string.Empty);
             config.Rx2DataRate = envVars.GetEnvVar("RX2_DATR", string.Empty);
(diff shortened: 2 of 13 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -157,52 +157,69 @@ public bool TryGetUpstreamChannelFrequency(Rxpk upstreamChannel, out double freq
...
+            if (nwkSrvRx2Freq.HasValue)
+            {
+                Logger.Log(devEUI, $"using custom gateway RX2 frequency {nwkSrvRx2Freq}", LogLevel.Debug);
+                return nwkSrvRx2Freq.Value;
+            }
+            else
+            {
+                // default frequency
+                Logger.Log(devEUI, $"using standard region RX2 frequency {this.RX2DefaultReceiveWindows.frequency}", LogLevel.Debug);
+                return this.RX2DefaultReceiveWindows.frequency;
```
p-schuler, line 177: nit: frequency should be upper case?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs
```diff
@@ -56,7 +56,7 @@ public static bool TryResolveRegion(Rxpk rxpk, out Region region)
                 region = EU868;
                 return true;
             }// US902-928
-            else if (rxpk.Freq <= 928 && rxpk.Freq >= 902)
+            else if (rxpk.Freq <= 915 && rxpk.Freq >= 902)
```
p-schuler, line 59: either the comment is wrong or the value.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -157,52 +157,69 @@ public bool TryGetUpstreamChannelFrequency(Rxpk upstreamChannel, out double freq
...
+                    // If not we use the region default.
+                    var datr = this.DRtoConfiguration[this.RX2DefaultReceiveWindows.dr].configuration;
+                    Logger.Log(devEUI, $"using standard region RX2 datarate {datr}", LogLevel.Debug);
+                    return datr;
                 }
                 else
                 {
-                    datr = this.DRtoConfiguration[this.RX2DefaultReceiveWindows.dr].configuration;
-                    Logger.Log(devEUI, $"device rx2 ({rx2Dr}) is invalid, using default: {this.RX2DefaultReceiveWindows.dr}, datr: {datr}, region: {this.LoRaRegion}", LogLevel.Debug);
+                    var datr = nwkSrvRx2Dr;
```
p-schuler, line 218: can we refactor the 2 default cases and put them at the end. we return early if we have other configurations.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs
Mandur, line 59: Precised the comment

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -95,7 +129,11 @@ public static NetworkServerConfiguration CreateFromEnviromentVariables()
             config.GatewayID = envVars.GetEnvVar("IOTEDGE_DEVICEID", string.Empty);
             config.HttpsProxy = envVars.GetEnvVar("HTTPS_PROXY", string.Empty);
             config.Rx2DataRate = envVars.GetEnvVar("RX2_DATR", string.Empty);
-            config.Rx2DataFrequency = envVars.GetEnvVar("RX2_FREQ", config.Rx2DataFrequency);
```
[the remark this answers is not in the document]
Mandur: Thanks for the catch

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/CaseInsensitiveEnvironmentVariables.cs
```diff
@@ -49,7 +49,7 @@ public uint GetEnvVar(string key, uint defaultValue)
             return value;
         }
 
-        public double GetEnvVar(string key, double defaultValue)
+        public double? GetEnvVar(string key, double? defaultValue)
```
fbeltrao, line 52: Is this method being used in a scenario that a defaultValue other than null is set?
If it returns nullable of double you could return null in case you cannot get a valid value.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -95,7 +129,7 @@ public static NetworkServerConfiguration CreateFromEnviromentVariables()
...
+            config.Rx2Frequency = envVars.GetEnvVar("RX2_FREQ", (double?)null);
```
fbeltrao, line 132: As mentioned before, if we always use the double? null as default we could set null as default parameter value.
Another option is to remove the parameter and return null if not found or invalid.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -217,7 +230,7 @@ public string GetDownstreamDR(Rxpk upstreamChannel, uint rx1DrOffset = 0)
                 if (rx1DrOffset <= this.RX1DROffsetTable.GetUpperBound(1))
                 {
                     // in case of EU, you respond on same frequency as you sent data.
-                    return this.DRtoConfiguration[(uint)this.RX1DROffsetTable[this.GetDRFromFreqAndChan(upstreamChannel.Datr), rx1DrOffset]].configuration;
+                    return this.DRtoConfiguration[(ushort)this.RX1DROffsetTable[this.GetDRFromFreqAndChan(upstreamChannel.Datr), rx1DrOffset]].configuration;
```
fbeltrao, line 233: Should we add log if for some reason the value we want is not found in the dictionary? This could make our life easier if it breaks here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/CaseInsensitiveEnvironmentVariables.cs
Mandur: It is currently not used for another default value, however some other methods signature are. I thought keeping a similar signature between method might be good, additionally maybe the functionality could be needed in the future.

fbeltrao: Perhaps add a default value:
`public double? GetEnvVar(string key, double? defaultValue = null)`

this way callers don't have to have to set a null nullable value as the default.

Mandur: Good idea will do

Mandur:
Sorry for the force push, I forgot my origin branch was disjoint from upstream, should be good now

Issue #253, opened 2020-08-13
Failed to deploy path that does not exist


levalencia:
I got the master branch and I am trying to deploy the Azure Facade Function.
I changed the remote function setting from 3 to 2 as in the previous Issue I just marked as closed.

However I get this error:
Failed to deploy path that does not exist: c:\Users\username\Downloads\Repos\iotedge-lorawan-starterkit\LoRaEngine\LoraKeysManagerFacade\bin\Release\netstandard2.0\publish

Build is working fine:

```

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:01.24

Terminal will be reused by tasks, press any key to close it.

> Executing task: dotnet publish --configuration Release <

Microsoft (R) Build Engine version 16.6.0+5ff7b0c9e for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  Logger -> C:\Users\username\Downloads\Repos\iotedge-lorawan-starterkit\LoRaEngine\modules\LoRaWanNetworkSrvModule\Logger\bin\Release\netcoreapp3.1\Logger.dll
  LoRaTools -> C:\Users\username\Downloads\Repos\iotedge-lorawan-starterkit\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoraTools\bin\Release\netcoreapp3.1\LoRaTools.dll
  LoraKeysManagerFacade -> C:\Users\username\Downloads\Repos\iotedge-lorawan-starterkit\LoRaEngine\LoraKeysManagerFacade\bin\Release\netcoreapp3.1\bin\LoraKeysManagerFacade.dll
  LoraKeysManagerFacade -> C:\Users\username\Downloads\Repos\iotedge-lorawan-starterkit\LoRaEngine\LoraKeysManagerFacade\bin\Release\netcoreapp3.1\publish\
```

I am not sure what I am doing wrong from my side

Thanks

Mandur:
From which branch are you trying to deploy? Master or dev? It looks to me like you are somewhat using .net 3.1 locally?

levalencia:
master branch, when I checked the bin folder in VS Code, there is not netstandard2.0 in the folder structure.

Mandur:
Okay, I think you might not have .NET 2 installed. If you let me 15 minutes I was working on the 1.0.4 release that should simplify this for you.

Mandur:
[...]

Issue #251, opened 2020-08-12
Documentation error


levalencia:
In the developer guidance page:
https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/Docs/devguide.md

There is a broken link:, in the text: Build and deploy the Azure function.
The Azure Functions Link is broken

Mandur:
Thank you for reporting, I will ammend

Mandur:
It was fixed in #254. Thank you very much @levalencia for bringing this up.

PR #254, opened 2020-08-13
Upgrade Azure function guide

Mandur:
* Fix documentation for #251 
* Upgrade Azure function dev experience to .net 3.1 #253 
* Upgrade deploy to Azure button to deploy version 1.0.4


fbeltrao:
In case someone have the v2 of the Azure Function, is there any steps that should be taken (i.e. update the function runtime in the settings)?

Mandur:
Not I am aware of, all the dependencies have been updated as part of the previous PR. this is related to how the vs code extension deploy Azure function zip package

Issue #258, opened 2020-08-25
LoraKeysManagerFacade fails when run locally with VS2019


Feiko:
## Expected Behavior
Debugging LoraKeysManagerFacade locally using Azure Functions

## Current Behavior
Launching LoraKeysManagerFacade with Visual Studio or Visual Studio Code outputs the following error:

[25-8-2020 08:01:56] A host error has occurred during startup operation '2b8f3a21-cd48-4742-9ffa-6dbebe4972d4'.
[25-8-2020 08:01:56] func: Invalid host services. Microsoft.Azure.WebJobs.Script.WebHost: The following service registrations did not match the expected services:
[25-8-2020 08:01:56]   [Invalid] ServiceType: Microsoft.Extensions.Hosting.IHostedService, Lifetime: Singleton, ImplementationType: LoraKeysManagerFacade.DevAddressCacheManager.
Value cannot be null. (Parameter 'provider')

## Steps to Reproduce

1. Clone the repository 
2. Open LoraEngine Solution
3. Add local.settings.json to LoraKeysManagerFacade project
4. added the connectionstrings to IoTHub Redis and WebjobsStorage
5. Run LoRaKeysManagerFacade

## Context (Environment)
C:\Users\xxxxxx\AppData\Local\AzureFunctionsTools\Releases\3.10.0\cli_x64>func --version
3.0.2798

### Device (Host) Operating System
Windows 10 x64

### Architecture
x64

### Container Operating System

none

#### Docker

none

## Logs
Azure Functions Core Tools (3.0.2798 Commit hash: d9d99872a51c04b1c2783198b1ee1104da6b064f)
Function Runtime Version: 3.0.14191.0
[25-8-2020 08:14:56] Building host: startup suppressed: 'False', configuration suppressed: 'False', startup operation id: '5ca9f522-d9df-4281-b059-e62ca5fef475'
[25-8-2020 08:14:56] Reading host configuration file 'C:\Users\gorte506\Desktop\iotedge-lorawan-starterkit-1.0.4\LoRaEngine\LoraKeysManagerFacade\bin\Debug\netcoreapp3.1\host.json'
[25-8-2020 08:14:56] Host configuration file read:
[25-8-2020 08:14:56] {
[25-8-2020 08:14:56]   "version": "2.0"
[25-8-2020 08:14:56] }
[25-8-2020 08:14:56] Loading functions metadata
[25-8-2020 08:14:56] Reading functions metadata
[...]

PR #259, opened 2020-08-28
Change Dev Addr Cache to be updated with a cron function instead of a worker job

Mandur:
Currently the Dev Addr cache sync is working as a background job. This model is not appropriate with the Azure function and triggers errors (#258 ) in a non deterministic fashion.
Proposal is to change the hosted service to a function triggered every 5 minutes. This change would need no logic change on the current application logic.


LoRaEngine/LoraKeysManagerFacade/SyncDevAddrCache.cs
```diff
@@ -0,0 +1,31 @@
...
+        public SyncDevAddrCache(LoRaDevAddrCache loRaDevAddrCache, RegistryManager registryManager)
+        {
+            this.loRaDevAddrCache = loRaDevAddrCache;
+            this.registryManager = registryManager;
+        }
+
+        [FunctionName("SyncDevAddrCache")]
+        public async Task Run([TimerTrigger("0 */5 * * * *")]TimerInfo myTimer, ILogger log)
+        {
+            log.LogDebug($"SyncDevAddrCache function executed at: {DateTime.Now}");
```
fbeltrao, line 27: Prefer DateTime.UtcNow as it is deterministic regardless of the host.

fbeltrao, line 27: Wonder if we should include `myTimer` to the logs. It has [interesting information](https://docs.microsoft.com/azure/azure-functions/functions-bindings-timer?tabs=csharp#usage)

LoRaEngine/LoraKeysManagerFacade/SyncDevAddrCache.cs
```diff
@@ -0,0 +1,31 @@
...
+        private RegistryManager registryManager;
+
+        public SyncDevAddrCache(LoRaDevAddrCache loRaDevAddrCache, RegistryManager registryManager)
+        {
+            this.loRaDevAddrCache = loRaDevAddrCache;
+            this.registryManager = registryManager;
+        }
+
+        [FunctionName("SyncDevAddrCache")]
+        public async Task Run([TimerTrigger("0 */5 * * * *")]TimerInfo myTimer, ILogger log)
```
fbeltrao, line 25: Should we add tests to the function to ensure the device cache:
-  is being cleared 
- operation is waited
- exception are raised

?

fbeltrao (approved):

LoRaEngine/LoraKeysManagerFacade/SyncDevAddrCache.cs
ronniesa: Sounds like a good idea to have the timer info on the logs

ronniesa (approved):

LoRaEngine/LoraKeysManagerFacade/SyncDevAddrCache.cs
Mandur: I think a lot of tests regarding cache function are already present [here](https://github.com/Azure/iotedge-lorawan-starterkit/blob/fdbc3c17bf5bb0805bbf6578d8bd6d60f229395b/LoRaEngine/test/LoraKeysManagerFacade.Test/DevAddrCacheTest.cs).
Regarding the operation is waited, what exactly do you mean? And what exception do you have in mind?

LoRaEngine/LoraKeysManagerFacade/SyncDevAddrCache.cs
Mandur: I added some information to the log, let me know if you think it's enough

LoRaEngine/LoraKeysManagerFacade/SyncDevAddrCache.cs
fbeltrao: If loRaDevAddrCache.PerformNeededSyncs throws an exception

fbeltrao (approved):

PR #260, opened 2020-09-04
Fix missleading logs

Mandur:
A wrong "the provided RX2DataRate is not valid" was displayed on the join logs in case a custom RX2DR is not set.


fbeltrao:
If I understand correctly the change is that we only validate if a RX2 was actually defined. Before, the log indicating invalid value would be written even if one was not set.

fbeltrao (approved):

fbeltrao:
Do we have unit tests that cover the desired value being out of the limits?

Mandur:
Yes in Check_Correct_RXPK_Datr_Are_Accepted  and Check_incorrect_RXPK_Datr_Are_Refused in rxpktest.cs

Issue #270, opened 2020-11-03
dotnet publish -c Release -r win10-x64 errors


levalencia:
I am trying to use the cli tools to register devices with the comand below:

dotnet publish -c Release -r win10-x64 errors

But I get all these errors.
```


Determining projects to restore...
C:\Users\user\Downloads\Repos\iotedge-lorawan-starterkit\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.csproj : error NU1605: Detected package downgrade: System.Runtime.InteropServices from 4.3.0 to 4.1.0. Reference the package directly from the project to select a different version.  [C:\Users\user\Downloads\Repos\iotedge-lorawan-starterkit\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.sln]
C:\Users\user\Downloads\Repos\iotedge-lorawan-starterkit\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.csproj : error NU1605:  loradeviceprovisioning -> Microsoft.Azure.Devices 1.22.0 -> Microsoft.Azure.Amqp 2.4.4 -> System.Net.WebSockets.Client 4.0.2 -> Microsoft.Win32.Primitives 4.0.1 -> runtime.win.Microsoft.Win32.Primitives 4.3.0 -> System.Runtime.InteropServices (>= 4.3.0)  [C:\Users\user\Downloads\Repos\iotedge-lorawan-starterkit\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.sln]
C:\Users\user\Downloads\Repos\iotedge-lorawan-starterkit\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.csproj : error NU1605:  loradeviceprovisioning -> Microsoft.Azure.Devices 1.22.0 -> Microsoft.Azure.Amqp 2.4.4 -> System.Net.WebSockets.Client 4.0.2 -> System.Runtime.InteropServices (>= 4.1.0) [C:\Users\user\Downloads\Repos\iotedge-lorawan-starterkit\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.sln]
[...]

PR #271, opened 2020-11-04
Bug #270 Fix error NU1605: Detected package downgrade on dotnet publish

TechPreacher:
Fixes Bug #270 

Problem fixed according to solution described in https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1605#issue-1

Problem
Certain combinations of packages which shipped with .NET Core 1.0 and 1.1 are not compatible with each other when they are referenced together in a .NET Core 3.0 or higher project, and a RuntimeIdentifier is specified. The problematic packages generally start with System. or Microsoft., and have version numbers between 4.0.0 and 4.3.1. In this case, the downgrade message will have a package starting with runtime. in the dependency chain.

Solution
To work around this issue, add the following PackageReference:
```xml
<PackageReference Include="Microsoft.NETCore.Targets" Version="3.0.0" PrivateAssets="all" />
```


Mandur (approved):

TechPreacher:
Should we close this PR? @ronniesa, @fbeltrao?

PR #279, opened 2020-12-20
Fix Desired Datarate Null check

Mandur:
A bug was introduced in dev in #260 . Due to one device of the gateway being out use, this was not picked up until it was brought online again. 
This PR should also make the CI functional again.


Mandur:
Accidentally submitted it, I need to carry more checks

PR #298, opened 2021-05-18
[OneWeek] Abstract common logic

Mandur:
Abstract common logic from the  UdpServer to the PhysicalClient


danigian (approved):
Looks good to me

PR #318, opened 2021-08-12
commit to test some PR feature at @mandur's request

kebetsi:


Mandur:
Thank you very much for doing this @kebetsi, we were able to assert some behaviors of the Github actions CI thanks to this!

PR #320, opened 2021-09-03
change tmmst to ulong

Mandur:
Currently the base implementation of the RXPK (Physical layer) use uint to represent the time at which to send payload. It seems this pause problems on windows. We propose switching it to ulong


Mandur:
This cause too much issue and would need a deeper implementation, I think this will be covered by the migration to the LNS

PR #327, opened 2021-09-12
Fix github action syntax

Mandur:
Current syntax is invalid. I am trying this fix to correct it. Also trying to reduce the set of permission given to the workflow


TechPreacher:
Looks good.

TechPreacher (approved):

PR #328, opened 2021-09-13
CI agent IP address is set on secrets

Mandur:
Due to changes in network topology on where the CI is running, we need to update its IP address. In order to avoid PRs for every time this happens, I propose to use a secret. Downside is that we cannot see the current value.


Mandur:
Closing as #334 provided a better solution

PR #331, opened 2021-09-14
Fix dev experience

Mandur:
The current PR aims to fix the current VS Code dev experience. In cases where the variables SPI_DEV and SPI_SPEED are not set the packet forwarder will crash. 
Also, amd64 version of the packet forwarder is depending on dotnet images no longer available, updating to plain debian (image size does not matter as it is only a build image).

### DIFF
--- a/LoRaEngine/modules/LoRaWanPktFwdModule/start_pktfwd.sh
+++ b/LoRaEngine/modules/LoRaWanPktFwdModule/start_pktfwd.sh
@@ -20,7 +20,7 @@ fi
 #get current architecture for the mess processor
 arch="$(uname -m)"
 if [[ $arch != *"arm"* ]]; then
-    if [ -z "$SPI_DEV" ]; then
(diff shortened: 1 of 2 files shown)


LoRaEngine/modules/LoRaWanPktFwdModule/start_pktfwd.sh
```diff
@@ -20,7 +20,7 @@ fi
 #get current architecture for the mess processor
 arch="$(uname -m)"
 if [[ $arch != *"arm"* ]]; then
-    if [ -z "$SPI_DEV" ]; then
+    if [ -z "$SPI_DEV" ] || [ "$SPI_DEV" == '$PKT_FWD_SPI_DEV' ]; then
```
p-schuler, line 23: nit: we could add some debug output to indicate the value was not set.

p-schuler (approved):

LoRaEngine/modules/LoRaWanPktFwdModule/start_pktfwd.sh
Mandur: Good idea! I will add logs indicate we are taking the default

PR #332, opened 2021-09-14
Additional documentation on values for template

spygi:
While following the Quick start guide, the values that I chose caused template validation errors or/and failed deployments after the deployment started.

This PR adds documentation on possible values that are user-provided to avoid these failures and allow a more smooth onboarding.


p-schuler (approved):
lgtm

PR #337, opened 2021-09-16
Fix E2E results

Mandur:
The CI E2E is currently outputting error as -even when a test stage is skipped because of label- we try to upload test result as job output. The current PR aims at fixing this issue


Mandur:
The CI is

danigian (approved):

Issue #361, opened 2021-09-22
Fix Dependabot issue with `tmpl`


roel4ez:
see https://github.com/Azure/iotedge-lorawan-starterkit/security/dependabot/Samples/UniversalDecoder/package-lock.json/tmpl/open

PR #363, opened 2021-09-23
Fix dependabot warning on templ library

Mandur:
# Fix issue #361 

## What is being addressed

Closing #361. Succesfull run of the tests for the universal decoders [here](https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1265665240)

## How is this addressed

We upgrade the libraries from node in the package.lock.json

### DIFF
--- a/Samples/UniversalDecoder/package-lock.json
+++ b/Samples/UniversalDecoder/package-lock.json
@@ -35,29 +35,29 @@
             }
         },
         "node_modules/@babel/compat-data": {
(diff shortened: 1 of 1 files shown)


roel4ez (approved):

p-schuler:
We could automate that step: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/enabling-and-disabling-version-updates

PR #426, opened 2021-09-29
Pin .NET Core SDK to 3.1.4xx

atifaziz:
## What is being addressed

The same build tools and SDK is used on the developer's machine as well as the CI.

## How is this addressed

The SDK used (3.1) is pinned via [`global.json`](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x). During CI builds, `dotnet --info` is executed to list the installed SDKs and runtimes as well as get a confirmation of the SDK being actually used.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,11 +62,11 @@ jobs:
       - env_var
     if: needs.env_var.outputs.RunTestsOnly != 'true'
     steps:
-    - name: .NET (Core) SDK Information
-      run: dotnet --info
-
     - uses: actions/checkout@v2
       name: Checkout current branch
+
+    - name: .NET (Core) SDK Information
+      run: dotnet --info
 
     # build LoRa Engine
     - name: Build LoRa Engine
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,6 +62,9 @@ jobs:
       - env_var
     if: needs.env_var.outputs.RunTestsOnly != 'true'
     steps:
+    - name: .NET (Core) SDK Information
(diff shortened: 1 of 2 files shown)


global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "3.1.400",
+    "rollForward": "latestFeature"
```
Mandur, line 4: Wouldn't latestfeature pin it to 3.1.xxx and latestPatch to 3.1.4xx? 
(based on [this article](https://andrewlock.net/exploring-the-new-rollforward-and-allowprerelease-settings-in-global-json/#understanding-the-various-rollforward-policies))

.github/workflows/ci.yaml
```diff
@@ -65,6 +65,9 @@ jobs:
     - uses: actions/checkout@v2
       name: Checkout current branch
 
+    - name: .NET (Core) SDK Information
```
Mandur, line 68: Do we want to keep this in the ci, or was it just for debugging?

Mandur (approved):

global.json
atifaziz: That's correct, but here we are setting 3.1.400 as the minimum baseline. So we pin to 3.1._N_ where _N_ >= 400.

.github/workflows/ci.yaml
atifaziz: I think it's harmless and worthwhile. Since we allow rolling-forward by feature, it could help if we start seeing different tooling behaviours being observed on CI and development environments. Perhaps then we'll need to pin down more strictly, but which can prove to be cumbersome for developers who may have a more recent version.

Issue #427, opened 2021-09-29
Invalid `mkdir` does not fail CI


atifaziz:
The `mkdir` on line 87 of `ci.yaml` is invalid:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/116e353bd61133acde13dd9ed6f96ca7156544d1/.github/workflows/ci.yaml#L84-L88

During a CI run, it produces the following error (see [`logs_502.zip`](https://github.com/Azure/iotedge-lorawan-starterkit/files/7253507/logs_502.zip) for an example):

    /home/runner/work/_temp/2c161a66-5967-4a06-b812-1765310d4bc9.sh: line 1: Build.SourcesDirectory: command not found

## Expected Behavior

Expected the step to fail and the following `dotnet test` to not run. The [default behaviour for `bash` and built-in shell is to fail fast](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference).

## Current Behavior

`mkdir` fails silently and continues to run tests.

atifaziz:
This requires an investigation into why a command that failed did not fail the run step. It means that other problems could potentially creep in and go unnoticed.

atifaziz:
It seems like what's going on here is that `mkdir` is _succeeding_ by creating a directly called `results`. The `$(Build.SourcesDirectory)` part gets interpreted as a [command substitution](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03) where `Build.SourcesDirectory` becomes the command to execute. This is why the shell reports `command not found`. Unfortunately, it seems that the error handling behaviour (with `-e`) is only defined for _simple commands_ and here it seems [command substitution errors are ignored](https://unix.stackexchange.com/questions/23026/how-can-i-get-bash-to-exit-on-backtick-failure-in-a-similar-way-to-pipefail). For more background and depth, see “[How can I get bash to exit on backtick failure in a similar way to pipefail?](https://unix.stackexchange.com/questions/23026/how-can-i-get-bash-to-exit-on-backtick-failure-in-a-similar-way-to-pipefail)”.

Since we can't do much here as a follow-up action (except to avoid trouble by using simple commands only), I am closing this.

PR #428, opened 2021-09-29
Remove invalid `mkdir` command in CI workflow

atifaziz:
## What is being addressed

The invalid `mkdir` command identified and described in issue #427.

## How is this addressed

The command has been completely removed by this PR as it was invalid and serving no purpose.


Mandur (approved):
Probably a leftover from ADO

PR #430, opened 2021-09-30
Fix CI to test same configuration as being built

atifaziz:
## What is being addressed

The unit test step in the CI workflow did not test the same configuration as [the one being built (release)](https://github.com/Azure/iotedge-lorawan-starterkit/blob/b34f8014100051c1a6a224ccead9ca856c7d640f/.github/workflows/ci.yaml#L21). For example, if the build step was building the release configuration:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/b34f8014100051c1a6a224ccead9ca856c7d640f/.github/workflows/ci.yaml#L71-L73

the test step would _re-build_ the debug configuration before testing:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/b34f8014100051c1a6a224ccead9ca856c7d640f/.github/workflows/ci.yaml#L87-L90

That's because the `--configuration` switch missing and the default configuration for `dotnet test` in debug.

## How is this addressed

Add the missing `--configuration` switch to `dotnet test` and passed the value of `buildConfiguration`.

---
PS I also put each `dotnet test` invocation on its line for readability since it would have the same effect as logically and-ing them with `&&`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,7 +87,9 @@ jobs:
     # Run unit tests
     - name: Run unit tests
       run: |
-        dotnet test --logger trx LoRaEngine/test/LoRaWanNetworkServer.Test/*.csproj -r LoRaEngine/test/TestResults/  &&  dotnet test --logger trx LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/*.csproj -r LoRaEngine/test/TestResults/ && dotnet test --logger trx LoRaEngine/test/LoraKeysManagerFacade.Test/*.csproj -r LoRaEngine/test/TestResults/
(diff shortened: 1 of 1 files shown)


Mandur:
Where you able to understand the && syntax? did it really only run following tests when the first one succeeded?

Mandur (approved):

atifaziz:
> Where you able to understand the && syntax? did it really only run following tests when the first one succeeded?

Yes. According to the documentation, [the default behaviour for `bash` and built-in shell is to fail fast](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference). I believe `bash` is started with the `-e` option therefore that's the effectively the same as invoking simple commands separately or and-ing together with `&&`.

Issue #366, opened 2021-09-24
Investigate and setup Codecov


roel4ez:
We want to get a good overview of the current state of the code coverage, and we want to be able to monitor the progress.

This story deals with investigating and setting up [Codecov](https://about.codecov.io/)

- ~~ADR with decision on how and why codecov is used~~
- [x] Investigate integration
- [x] Codecov is implemented in a GitHub action
- [x] Add coverage badge

Mandur:
I am not sure we need an ADR for that, we could do this in the parking lot

atifaziz:
Closed via #431, #432 and #434.

PR #431, opened 2021-09-30
Report code coverage via unit tests to Codecov

atifaziz:
# PR for story #366

## What is being addressed

Code coverage reporting via [Codecov](https://about.codecov.io/).

## How is this addressed

The code coverage results from running the unit tests is uploaded to [codecov.io](https://about.codecov.io/) using Codecov's supplied loader.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,8 +91,7 @@ jobs:
     # Run unit tests
     - name: Run unit tests
       run: |
-        dotnet test --configuration ${{ env.buildConfiguration }} --logger trx -r \
-          LoRaEngine/test/TestResults \
+        dotnet test --configuration ${{ env.buildConfiguration }} --logger trx -r LoRaEngine/test/TestResults \
           /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura \
           --filter 'FullyQualifiedName~LoRaWanTest|FullyQualifiedName~LoRaWan.NetworkServer.Test|FullyQualifiedName~LoraKeysManagerFacade.Test' \
           LoRaEngine
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,7 +94,7 @@ jobs:
         dotnet test --configuration ${{ env.buildConfiguration }} --logger trx -r \
           LoRaEngine/test/TestResults \
           /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura \
-          --filter 'FullyQualifiedName~LoRaWanTest|FullyQualifiedName~LoRaWan.NetworkServer.Test|FullyQualifiedName~LoraKeysManagerFacade.Test)' \
+          --filter 'FullyQualifiedName~LoRaWanTest|FullyQualifiedName~LoRaWan.NetworkServer.Test|FullyQualifiedName~LoraKeysManagerFacade.Test' \
           LoRaEngine
       
     # Upload test results as artifact
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,7 +91,11 @@ jobs:
     # Run unit tests
     - name: Run unit tests
       run: |
-        dotnet test --logger trx LoRaEngine/test/LoRaWanNetworkServer.Test/*.csproj -r LoRaEngine/test/TestResults/  &&  dotnet test --logger trx LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/*.csproj -r LoRaEngine/test/TestResults/ && dotnet test --logger trx LoRaEngine/test/LoraKeysManagerFacade.Test/*.csproj -r LoRaEngine/test/TestResults/
+        dotnet test --configuration ${{ env.buildConfiguration }} --logger trx -r \
+          LoRaEngine/test/TestResults \
+          /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura \
+          --filter 'FullyQualifiedName~LoRaWanTest|FullyQualifiedName~LoRaWan.NetworkServer.Test|FullyQualifiedName~LoraKeysManagerFacade.Test)' \
+          LoRaEngine
       
     # Upload test results as artifact
     - uses: actions/upload-artifact@v1
@@ -99,6 +103,12 @@ jobs:
       with:
         name: unit-test-results
         path: LoRaEngine/test/TestResults
+
+    - name: Upload to Codecov
+      run: |
+        curl -OsSL https://uploader.codecov.io/v0.1.0_4653/linux/codecov
+        chmod +x codecov
+        ./codecov
 
   check_if_deploy:
     needs: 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -91,7 +91,11 @@ jobs:
     # Run unit tests
     - name: Run unit tests
       run: |
-        dotnet test --logger trx LoRaEngine/test/LoRaWanNetworkServer.Test/*.csproj -r LoRaEngine/test/TestResults/  &&  dotnet test --logger trx LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/*.csproj -r LoRaEngine/test/TestResults/ && dotnet test --logger trx LoRaEngine/test/LoraKeysManagerFacade.Test/*.csproj -r LoRaEngine/test/TestResults/
+        dotnet test --configuration ${{ env.buildConfiguration }} --logger trx -r \
+          LoRaEngine/test/TestResults \
+          /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura \
+          --filter 'FullyQualifiedName~LoRaWanTest|FullyQualifiedName~LoRaWan.NetworkServer.Test|FullyQualifiedName~LoraKeysManagerFacade.Test' \
```
p-schuler, line 97: I was looking for the tools tests, until I realized, they are the LoRaWanTest.

p-schuler (approved):
LGTM

atifaziz:
@p-schuler, @spygi: FYI, the [Codecov bot is now publishing the coverage report into the PR](https://github.com/Azure/iotedge-lorawan-starterkit/pull/431#issuecomment-930954179).

.github/workflows/ci.yaml
atifaziz: Yeah, there was quite a bit of confusion for me there as well. It seems some of the assembly names, project names and namespaces are not aligned. It would be good to clean that up eventually. I am [taking notes](https://github.com/Azure/iotedge-lorawan-starterkit/projects/8#card-69834110).

spygi, line 97: [nit] we don't need to repeat `FullyQualifiedName` for each test class (didn't find documentation on this but tried locally) 
NB: I don't recommend removing fully the explicit filter [as is mentioned here](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test#:~:text=An%20expression%20without%20an%20%3Coperator%3E%20is%20automatically%20considered%20as%20a%20contains%20on%20FullyQualifiedName%20property%20(for%20example%2C%20dotnet%20test%20--filter%20xyz%20is%20same%20as%20dotnet%20test%20--filter%20FullyQualifiedName~xyz).) - only not to repeat it every time.

LoRaEngine/test/LoRaWan.Test.Shared/LoRaWan.Test.Shared.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <IsTestProject>false</IsTestProject>
```
spygi, line 5: 👌 because otherwise we were getting some "false positives" (trying to find tests in this project and failing)

.github/workflows/ci.yaml
```diff
@@ -100,6 +104,12 @@ jobs:
         name: unit-test-results
         path: LoRaEngine/test/TestResults
 
+    - name: Upload to Codecov
+      run: |
+        curl -OsSL https://uploader.codecov.io/v0.1.0_4653/linux/codecov
+        chmod +x codecov
+        ./codecov
+
```
spygi, line 112: Seems a bit weird to have to do this with CURL-ing a specific version especially and making it executable because of updates - we would have to bump up the version manually here. We could use [the latest always](https://uploader.codecov.io/linux/latest) to avoid this but shouldn't we use a [Github action for this](https://github.com/marketplace/actions/codecov#usage) - v2 is not deprecated if I understand correctly.

spygi (commented):
Can be merged from my side (it works) - though I think it would make sense to use the Github action.

.github/workflows/ci.yaml
spygi: Also, the token is not provided anywhere, I assume because Azure (the org) has given permission to the Codecov application?

atifaziz: Considering that it's version 0._x_, I assume that there can be breaking changes between minor versions and therefore would prefer to stick with a version known to be working and upgrade manually when needed. Perhaps one can review this when it goes 1._x_. 

Downloading the binary, making it executable and then executing it seemed simple enough that it didn't seem to warrant the use of a whole GitHub Action when the usage was already so straightforward (and frankly what I was used to doing from other projects when the action did not exist). That said, [I've made a note](https://github.com/Azure/iotedge-lorawan-starterkit/projects/8#card-69845696) on the [backlog](https://github.com/Azure/iotedge-lorawan-starterkit/projects/8) to consider using the GitHub Action.

.github/workflows/ci.yaml
atifaziz: > didn't find documentation on this but tried locally

If it's not documented then I'd rather not go with it if you don't mind. In fact, the filters didn't work reliably for me when _listing_, but that could be just bugs in the .NET Core SDK 3.1 tooling. I don't think we should be relying on the filtering on names in the longer run anyway. It would be much better to go with categories but I didn't want to expand the scope of this PR.

LoRaEngine/test/LoRaWan.Test.Shared/LoRaWan.Test.Shared.csproj
atifaziz: If you don't do this, you get an error from `dotnet test` along the lines of:

    Unable to find testhost.dll. Please publish your test project and retry.

I changed the `dotnet test` invocation in the CI workflow to run all tests in the solution, with some filters applied. However, some projects, like **LoRaWan.Tests.Shared**, reference [xunit](https://www.nuget.org/packages/xunit/) and contain code to be shared by actual test projects. Unless you flag them with `IsTestProject` as `false`, `dotnet test` mistakes them for test projects and fails with the above error when trying to run them.

spygi: I know :D that's why the 👍

.github/workflows/ci.yaml
spygi: I don't see it the same way: since we have the tools (Github action) we should use them. We can resolve here since this is a small/insignificant example but if we apply the same principle in general we will re-invent a lot of things, just food for thought.

p-schuler: Same here. I think since there is an action, it seems easier to use that to me.

atifaziz: I had already moved to the action version in PR #432, shortly after wrapping up this one. It is indeed simpler.

PR #434, opened 2021-09-30
Add coverage badge to read-me doc

atifaziz:
# PR for issue #366

## What is being addressed

Coverage badge for the project.

## How is this addressed

Added to the read-me document of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![LoRa CI](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml/badge.svg?branch=master)](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml)
 [![LoRa CI](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml/badge.svg?branch=dev)](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml)
 [![Markdown](https://github.com//Azure/iotedge-lorawan-starterkit/actions/workflows/md-linter.yaml/badge.svg?branch=dev)](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/md-linter.yaml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,7 @@
 [![LoRa CI](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml/badge.svg?branch=master)](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml)
 [![LoRa CI](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml/badge.svg?branch=dev)](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/ci.yaml)
 [![Markdown](https://github.com//Azure/iotedge-lorawan-starterkit/actions/workflows/md-linter.yaml/badge.svg?branch=dev)](https://github.com/Azure/iotedge-lorawan-starterkit/actions/workflows/md-linter.yaml)
+[![codecov](https://codecov.io/gh/Azure/iotedge-lorawan-starterkit/branch/dev/graph/badge.svg?token=OTwrbFLcPS)](https://codecov.io/gh/Azure/iotedge-lorawan-starterkit)
```
danigian, line 6: What about this token? Is it a secret one?

atifaziz: > What about this token?

This was provided in the URL in the [badges page of the project](https://docs.codecov.com/docs/status-badges).

> Is it a secret one?

I don't know, to be honest. It doesn't seem to be documented. If I was to take a guess, I imagine it's for private repos, but this one is public. I tried the URL without the token and it works fine without it so I've removed it with efd1f28d8bc8213af4ee7aea6d805077f6af1519.

danigian (approved):

README.md
p-schuler: If this was a secret we would need to rotate it now.

atifaziz: I checked with Codecov support and am told it's not considered a secret. Quoting the reply:

> That is called the graph token, and it is good only for calls to the graph endpoint for badges and the charts. It's not needed for public repos, but it is for private ones.
>
> We don't consider is a secret, based on it's read-only nature to coverage numbers…

There is no UI for regenerating the token and it's not linked to the upload token (which can be regenerated).

atifaziz: While the token is no secret, I worked with Codecov support to generate a new one, just to err on the safe side.

Issue #436, opened 2021-10-01
Update Code Analyzer and code style handling


roel4ez:
Consistent and documented code style

PR #446, opened 2021-10-06
Add common version file with analyzers

Mandur:
# PR for issue #445

Dependabot PRs are failing are multiple components reference the library on different version and it seems to be only applying the updated one by one.

## How is this addressed

We provide a centralized file "Directory.Build.props" to configure the version used by the project so that dependabot can upgrade all project at once. As a side benefit, we also centralize StyleCop and FxCop rules there.

> The current configuration **removes** the "threatwarningaserror" to make the modern analyzer pass until the list of rules are set in #436


LoRaEngine/Directory.Build.props
```diff
@@ -0,0 +1,30 @@
+<Project>
+    <ItemGroup>
+        <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.2">
```
atifaziz, line 3: Since we are on the .NET 5 SDK since merge of PR #437, [the analyzers are already included](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview) and can be enabled via the following properties:

```xml
        <EnableNETAnalyzers>true</EnableNETAnalyzers>
        <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
        <AnalysisMode>AllEnabledByDefault</AnalysisMode>
        <AnalysisLevel>latest</AnalysisLevel>
```

Issue #449, opened 2021-10-06
[Story] LoRaWAN Primitives


roel4ez:
This [implementation](https://github.com/lorabasics/basicstation/blob/master/src/rt.h) could help as reference

## Acceptance Criteria

- [x] Primitives (ID6, EUI, INT) are implemented
- [x] Primitives have unit tests

PR #464, opened 2021-10-07
Add LoRaWAN primitives object model

atifaziz:
# PR for story #449

## What is being addressed

LoRaWAN primitives like `DevAddr`, `DevEUI`, etc., are cumbersome to deal with. The current code base has no encapsulation of the bit manipulations, parsing and formatting of such elements.

## How is this addressed

This PR adds adds a strong-typed object model for the primitives, where each type encapsulates the various details of multi-field octets or multi-octet fields, instead of having them spread throughout the code base. The following types have been added to a new project called **LoRaWan**:

- `DevAddr`
- `DevNonce`
- `DevEui`, `JoinEui` & `StationEui`
- `FrameControl` & `FCtrlFlags` (**FCtrl**)
- `Hertz` (to represent frequencies)
- `AppKey`, `AppSessionKey` (**AppSKey**) & `NetworkSessionKey` (**NwkSKey**)
- `MacHeader` (**MHDR**)
- `MacMessage` (**MType**)
- `Mic` (**MIC**)
- `NetId` (**NetID**)
- `Id6` (parsing and formatting of addresses in **ID6** format)

The following supporting types have also been added that are not LoRaWan-specific:

- `Hexadecimal` (parsing and formatting of bytes to and from hexadecimal strings)
- `Data128` (optimised storage for byte data that's 128 bits wide like AES keys above)

---
PS The `Mic.ComputeForJoinRequest` method computes the MIC for a _join request_ that's designed to serve as a template for other MIC computations that will be added in a future PR as needed. This PR was primarily focused on adding an object model and less on computational or algorithmic aspect of LoRaWAN.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/FrameControlTests.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaToolsTest/FrameControlTests.cs
@@ -38,6 +38,8 @@ namespace LoRaWanTest
             public const bool DownlinkFramePending = false;
         }
 
+        const FCtrlFlags AllFCtrlFlags = FCtrlFlags.Adr | FCtrlFlags.AdrAckReq | FCtrlFlags.Ack | FCtrlFlags.FPending;
+
         [Theory]
         [InlineData(FCtrlFlags.None         , 0, False.Adr, False.AdrAckRequested, False.Ack, False.DownlinkFramePending, 0)]
         [InlineData(FCtrlFlags.None         , 3, False.Adr, False.AdrAckRequested, False.Ack, False.DownlinkFramePending, 3)]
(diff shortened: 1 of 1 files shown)


atifaziz:
> I see we didn't run the CI for this.

From what I can see [it did run successfully](https://github.com/Azure/iotedge-lorawan-starterkit/runs/3906313115). Where are you looking?

> I would have tried to see if all the builds are happy with the T4 addition.
> Can we make sure this is not breaking anything?

T4 does not run during the build, right now, so its use should not break anything. It requires explicit invocation from the IDE when someone changes the template to refresh the generated code. In fact, VS automatically run the template to refresh the code whenever it is saved. I have a [card on the backlog](https://github.com/Azure/iotedge-lorawan-starterkit/projects/8#card-70874622) to make sure that the generated code does not go out of sync with the template. For example, if someone changes the template but does not refresh the generated code then the CI build should break. This would serve as a guard if someone is editing the template using another IDE or editor than VS.

PR #495, opened 2021-10-12
Refactoring of region unit tests

MaggieSalak:
# PR for issue #452  

## What is being addressed

The region implementation unit tests were placed in a single class `RegionImplementationTest`. This PR aims to refactor the unit tests by splitting into separete classes, with shared methods implemented in a base class. 

## How is this addressed

- In place of the existing test class, new classes have been added for tests related to specific regions, with a base class `RegionTestBase` for the common code used in the test methods.
- In case of US915, the way the test data is generated has been changed, to instead use `MemberData` with a method generating possible combinations of data rates and frequencies.


MaggieSalak:
> Very nice, did you per case ran coverage on the region implementation to see if there were methods that were not covered? Asking as these tests will be the only checks we can do for China

I did. Local run in VS shows that both RegionEU868 and RegionUS915 have now 100% coverage. Here the codecov report does seem to show much lower coverage, not sure why, because we definitely have unit tests hitting the lines that here are highlighted in red. Could it be that it only shows that the introduced changes cover?

PR #530, opened 2021-10-14
Feature/436 address warnings

bastbu:
# PR for issue #436

## What is being addressed

This PR fixes warnings CA1829, CA1002, CA1008, CA1012, CA1028, CA1032, CA1024, CA1052, CA1036 and CA1027


p-schuler:
I don't see a change in .editorconfig. I don't know what was agreed on, but it feels like we should update the .editorconfig along with those changes, to make sure we enable all the rules we are fixing. I would prefer having them break the build rather than only generate warnings.

Issue #507, opened 2021-10-12
[Bug] Solution file should reside at the root - not within LoRaEngine


spygi:
## Expected Behavior
Main solution file should reside at the root folder in order to make it easier to discover and renamed to LoRaWan.sln (or similar)

## Current Behavior
Currently it's within LoRaEngine. That made more sense when the tests where also in the same folder but since now they are moved outside of LoRaEngine, the solution file should be moved as well.

PR #573, opened 2021-10-19
Move VS solution file to the root

atifaziz:
# PR for story #454

## What is being addressed

The VS solution file location is no longer a parent or sibling of the projects it contains since some test projects have been moved as part of other PRs for #454.

## How is this addressed

The VS solution file is moved to the root.

---
PS The also has the advantage of simplifying building or testing the solution as simple as writing `dotnet build` or `dotnet test` after cloning the repo.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,7 +75,7 @@ jobs:
 
     # build LoRa Engine
     - name: Build LoRa Engine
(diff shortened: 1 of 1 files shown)


spygi:
LGTM but VS can't open the solution - is it just me? Don't see what is wrong though with it.

spygi:
Closes #507

Issue #574, opened 2021-10-19
T4 generated code may go out of sync unnoticed


atifaziz:
PR #464 added code generation via T4 templates. A developer can change the template and forget to run it (either in VS or on the command line via [`dotnet t4`](https://www.nuget.org/packages/dotnet-t4)) to refresh the generated code.

## Expected Behavior

The generated code (`*.g.cs`) that is checked-in is always fresh and reflects its template (`*.g.tt`).

## Current Behavior

The generated code can go out of sync.

PR #575, opened 2021-10-19
Add CI check that T4 generated code is fresh

atifaziz:
# PR for issue #574

## What is being addressed

PR #464 uses T4 templates to generate some C# code. If someone updates the template but forgets to run it to update the generated code then the two can go out of sync.

## How is this addressed

Adds a check during CI that the generated code is fresh with respect to the template. This is done by running `dotnet t4` on the templates and then check with Git via `git diff --exitcode` if any files in the working tree have changes. If executing the template generated different code then what's committed then the working tree state will change to dirty and `git diff --exitcode` will return a non-zero exit code and fail the workflow step.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Keys.g.tt
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Keys.g.tt
@@ -15,7 +15,6 @@ namespace LoRaWan
     using System;
 <#  foreach (var t in new[]
         {
-            "NewKey",
             "AppKey",
             "AppSessionKey",
             "NetworkSessionKey",


atifaziz:
Note for reviewer:

Commit 2520a6826fc66cca7d6860206a72ce84544beaae was conducted as a test where the template was changed but not executed. The [full log](https://github.com/Azure/iotedge-lorawan-starterkit/files/7372508/logs_1480.zip) from CI shows that the [check effectively failed the workflow](https://github.com/Azure/iotedge-lorawan-starterkit/runs/3938043747#step:6:63). See the logs from the new steps, especially the last one where the check fails, for more details:

- [`4_Restore Tools.txt`](https://github.com/Azure/iotedge-lorawan-starterkit/files/7372520/4_Restore.Tools.txt)
- [`5_Refresh T4 Generated Code.txt`](https://github.com/Azure/iotedge-lorawan-starterkit/files/7372532/5_Refresh.T4.Generated.Code.txt)
- [`6_Check for Modifications.txt`](https://github.com/Azure/iotedge-lorawan-starterkit/files/7372536/6_Check.for.Modifications.txt)
 
The change was reverted with d8bb996cbb844734d38cbd1702e3e1dc48f680ea.

Issue #585, opened 2021-10-20
[Story] The CI Should be split up


Mandur:
## Description

Our current CI is big and complex. It also prevents parallel execution and serve two purposes: Run the unit/integration tests and the end-to-end test. Now, one CI run without // execution that could be an issue in the future. Splitting the pipeline will reduce the logic and complexity and provide a clear separation of concern (in the E2E pipeline we know that we want to deploy)

## Acceptance criteria

- [x] The Build and test and the E2E pipeline are split into two CIs
- [x] The concurrency control on the E2E should be done at a workflow level using the [Concurrency feature](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency)
- [x] Both pipelines trigger on PR (E2E only if the label is set)
- [x] All E2E variables should be provided as environment variable
- [x] Path triggers are changed to only trigger on path where code is available.
- [ ] build and test also build the containers locally

PR #602, opened 2021-10-20
Add editorconfig to docker build to fix the CI (again)

Mandur:
This fix the E2E CI, otherwise the build break with stylecop issues.

cf : https://github.com/Azure/iotedge-lorawan-starterkit/runs/3953546026?check_suite_focus=true#step:4:349


Mandur:
Very good point @bastbu, It is in the success criteria of #585

PR #605, opened 2021-10-21
Fix CI for integration test reporting

Mandur:
The test publishing CI report is not reporting the integration test status as the path is not correctly exported. This PR fixed that, you should now in the checks see an integration section and not only a unit test as in [the previous PRs](https://github.com/Azure/iotedge-lorawan-starterkit/pull/602/checks?check_run_id=3956001067)

### DIFF
--- a/.github/workflows/test_report.yaml
+++ b/.github/workflows/test_report.yaml
@@ -15,5 +15,5 @@ jobs:
       with:
         artifact: /(.*)results/
         name: '$1 Results'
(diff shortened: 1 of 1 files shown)


Mandur:
https://github.com/Mandur/iotedge-lorawan-starterkit/pull/30/checks?check_run_id=3968893041 Here is a view of the PR with the changes (the reacting PR is coming from dev here)

Issue #664, opened 2021-10-29
[Task] Cache station configuration in LNS


bastbu:
Story: #579

**Description**

Cache the station configuration in-memory per LNS. Make sure to attempt to load the configuration already during the discovery phase - reject any connection request if the configuration is not present in IoT Hub.

PR #665, opened 2021-10-29
Feature/664 cache station info

bastbu:
# PR for issue #664 

## What is being addressed

This PR introduces in-memory caching on the LNS for station configuration files.

## How is this addressed

The `router_config` message is cached in an in-memory cache on the LNS.


danigian (approved):
LGTM, maybe @p-schuler could have another look since he worked on caching for this project

p-schuler:
Looks good. Generally: the way we handle the device connection is not following the same practice as with other devices. It's a special case and it's ok for now. Let's see if the connection management becomes easier after the changes we plan in that area and we can then re-visit.

PR #673, opened 2021-10-31
Enforce file header validation

Mandur:
Add File header and break when it's missing

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -115,5 +115,5 @@ dotnet_diagnostic.IDE1007.severity=suggestion
 dotnet_diagnostic.IDE1008.severity=suggestion
 
  # Enforce file header
-dotnet_diagnostic.IDE0073.severity = warning
+dotnet_diagnostic.IDE0073.severity = error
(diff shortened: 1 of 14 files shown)


atifaziz (approved):

.editorconfig
```diff
@@ -113,3 +113,7 @@ dotnet_diagnostic.IDE1003.severity=suggestion
 dotnet_diagnostic.IDE1004.severity=suggestion
 dotnet_diagnostic.IDE1007.severity=suggestion
 dotnet_diagnostic.IDE1008.severity=suggestion
+
+ # Enforce file header
+dotnet_diagnostic.IDE0073.severity = warning
```
bastbu, line 118: I would suggest to either:
- remove this, since the `IDE0073` state that when `file_header_template` is set it requires a file header. Since we did not specify any rules explicitly in this file that should be a warning, we might as well be consistent and omit it
- Set it to `error` if we always want to enforce it.

bastbu (approved):

.editorconfig
Mandur: Fair point, I will set it to error

Issue #637, opened 2021-10-26
Socket Management


roel4ez:
There needs to be a way to keep the socket open and keeping track of it (single connection to concentrator)
Manage multiple upstream messages and use the same socket for downstream messages
The reconnection needs to be handled.

PR #676, opened 2021-11-01
Socket management 🚧

atifaziz:
# PR for story #637


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -24,158 +22,150 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
...
-            if (httpContext is null) throw new ArgumentNullException(nameof(httpContext));
+            using var socket = await httpContext.WebSockets.AcceptWebSocketAsync();
+            this.logger.Log(LogLevel.Debug, $"WebSocket connection from {httpContext.Connection.RemoteIpAddress} established");
+
+            try
+            {
+                await handler(httpContext, socket, cancellationToken);
+                await socket.CloseAsync(WebSocketCloseStatus.NormalClosure, "Goodbye", cancellationToken);
+            }
+            catch (OperationCanceledException ex)
```
bastbu, line 57: Not sure if we need to handle this at this stage - we handle this at `WebSocketWriterRegistry.SendAsync` already

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -24,79 +21,146 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
...
-                var networkInterface = NetworkInterface.GetAllNetworkInterfaces()
-                                                       .SingleOrDefault(ni => ni.GetIPProperties()
+                try
+                {
+                    var scheme = httpContext.Request.IsHttps ? "wss" : "ws";
+                    var url = new Uri($"{scheme}://{httpContext.Request.Host}{BasicsStationNetworkServer.DataEndpoint}/{stationEui}");
+
+                    var networkInterface = NetworkInterface.GetAllNetworkInterfaces()
+                                                           .SingleOrDefault(ni => ni.GetIPProperties()
                                                                                 .UnicastAddresses
```
danigian, line 97: nit: these rows seem misaligned to me 👍

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+                do
+                {
+                    result = await socket.ReceiveAsync(buffer.Memory, cancellationToken);
+#pragma warning disable IDE0010 // Add missing cases (all are covered)
+                    switch (result.MessageType)
+#pragma warning restore IDE0010 // Add missing cases
+                    {
+                        case WebSocketMessageType.Close:
+                            yield break;
+                        case var type and not WebSocketMessageType.Text:
```
danigian, line 62: nit: can't it just be WebSocketMessageType.Binary here? Instead of the pattern matching

Tests/Unit/LoRaWanTests/LinkedTimeoutCancellationTokenTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using Xunit;
+
+    public class LinkedTimeoutCancellationTokenTests
+    {
+        /*[Fact]
+        public void Equality()
+        {
```
danigian, line 15: What about this commented code?

danigian (approved):

p-schuler:
Would be great to follow the naming convention on the branches.

Tests/Unit/LoRaWanTests/LinkedTimeoutCancellationTokenTests.cs
atifaziz: Yes, just clarified with @bastbu, who said he couldn't get it to work. I'll address it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
atifaziz: Indeed. Fixed with 889fb2ce5d4ac3c429c4f59013f670d20aee7a7a.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketExtensions.cs
atifaziz: I'll admit this is being a bit defensive in case there are other message types in the future.

Tests/Unit/LoRaWanTests/LinkedTimeoutCancellationTokenTests.cs
bastbu: It was less about getting it to work and more about what kind of equality we want to implement on the `LinkedTimeoutCancellationToken`. Thanks for addressing it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <summary>
+        /// Removes closed sockets at default interval periods until cancellation is requested.
+        /// </summary>
+        public Task RunPrunerAsync(CancellationToken cancellationToken) =>
+            RunPrunerAsync(WebSocketWriterRegistry.DefaultPruningInterval, cancellationToken);
+
+        /// <summary>
+        /// Removes closed sockets at a given a interval period until cancellation is requested.
+        /// </summary>
+        public Task RunPrunerAsync(TimeSpan interval, CancellationToken cancellationToken)
```
p-schuler, line 147: I don't see this is being used anywhere except for test?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/TimeoutLinkedCancellationToken.cs
```diff
@@ -0,0 +1,63 @@
...
+                        tempTimeoutCts = null;
+                    }
+                    else
+                    {
+                        Token = this.timeoutCts.Token;
+                    }
+                }
+                finally
+                {
+                    tempTimeoutCts?.Dispose();
```
p-schuler, line 41: I'm probably reading this wrong, but if it is a reference, we dispose the timeoutCts but on line 36 we could have assigned that token to the Token property. If I understand the implementation of the CancellationTokenSource correctly, the timeout would not fire when you dispose it. Hence the token wouldn't be canceled?

p-schuler (changes requested):
Let's discuss the things I raised.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <remarks>
+        /// If the send fails because the socket has been closed prematurely then the socket writer
+        /// is unregistered before this method returns.
+        /// </remarks>
+        private async ValueTask SendAsync(TKey key, TMessage message, CancellationToken cancellationToken)
+        {
+            IWebSocketWriter<TMessage> socketWriter;
+
+            lock (this.sockets)
+                socketWriter = this.sockets[key].Object;
```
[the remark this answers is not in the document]
atifaziz: You would get a `KeyNotFoundException`. Ideally we should have a more meaningful exception type (`SocketDisconnectedException`?), but for now, the point is that there would an exception thrown exactly at this point. A null reference won't result.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
atifaziz: True. This was added in anticipation of being used, but then when I got to integrating the registry into the `LnsProtocolMessageProcessor`, it didn't seem absolutely necessary. If the socket gets closed then everything should clean up in an orderly fashion. If it's closed abnormally then an error will result during the next attempt to read or send a message and either will cause a deregistration. This could come under YAGNI so happy to remove for now and bring it back if and when we need it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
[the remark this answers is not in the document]
p-schuler: True - the point was more that we have an unhandled exception bubbling up that is probably not expected up in the stack. As discussed over the phone, let's fix this as part of the exception handling story.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/TimeoutLinkedCancellationToken.cs
p-schuler: The bug was fixed as part of a re-design of this.

p-schuler (approved):
LGTM

PR #680, opened 2021-11-01
Feature/411 Region test coverage improvements

MaggieSalak:
# PR for issue #411 

## What is being addressed

Improvements in Region test coverage

## How is this addressed

- Code coverage was analyzed and additional unit tests were added to increase the coverage where it was lacking.
- Several unit tests were removed from `RxpkTest` class as they were either duplicates of already existing unit tests in Region-specific test classes or they were moved to the `RegionTestBase` class as part of this PR.
- Newly added unit tests are implemented in `RegionTestBase` and use `MemberData` referencing data-provider methods in region-specific test classes. This was done in order to reduce code duplication in the test classes and make sure all regions have the same test logic.
- The previously existing region tests (with logic in region-specific classes) were not refactored yet - please let me know what you think of the approach and then I'll refactor the other tests (see also my comment in this PR).


MaggieSalak:
Please let me know your feedback regarding the change to have all test logic in the base class as opposed to each region-specific class calling into the method in base. If you agree with the change, I will refactor the remaining region tests.

Mandur (approved):
Very cool idea

p-schuler:
Very nice improvement! <3 Thanks!

Issue #557, opened 2021-10-18
Fix unobserved task execution


p-schuler:
We have a few instances where we fire async tasks and don't observe the result. The default behavior in .NET >= 4.5 has changed to not crash the process for unobserved task exceptions. Hence, those will go un-noticed. 

## Expected Behavior

At least log the exceptions when they are happening, if we decide not to observe the result of a task.

## Current Behavior

When a task is not observed, any exceptions that might get thrown, are not logged.

## Steps to Reproduce

```c#
_ = Task.Run(() => throw new Exception("unhandled exception"));
await Task.Delay(500);
```

spygi:
Got bitten by this one while trying to fix broken integration tests: because calls are not observed, tests either don't fail or fail with a random assertion error which makes it hard to debug (need to step through the code). Not sure if we want to reopen this or prioritize to observe calls.
/cc @p-schuler and @bastbu since it was discussed with him

PR #682, opened 2021-11-01
Bug/557 Analyze unobserved task execution occurrences

bastbu:
# PR for issue #557

## What is being addressed

In this PR we analyse unobserved task executions occurring in our production code, to ensure that they are logged if they are not handled.

## How is this addressed

We are adding comments to ensure that in places where we handle general exceptions, we are aware that this (partly) is because the tasks are unobserved.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
@@ -466,6 +466,7 @@ namespace LoRaWan.NetworkServer
         {
             if (this.classCDeviceMessageSender != null)
             {
+                // Exception is logged as part of DefaultClassCDevicesMessageSender.SendAsync.
                 _ = Task.Run(() => this.classCDeviceMessageSender.SendAsync(cloudToDeviceMessage));
             }
         }
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
@@ -466,8 +466,19 @@ namespace LoRaWan.NetworkServer
         {
             if (this.classCDeviceMessageSender != null)
             {
-                // Exception is logged as part of DefaultClassCDevicesMessageSender.SendAsync.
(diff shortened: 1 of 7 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -466,6 +466,7 @@ private void SendClassCDeviceMessage(IReceivedLoRaCloudToDeviceMessage cloudToDe
         {
             if (this.classCDeviceMessageSender != null)
             {
+                // Exception is logged as part of DefaultClassCDevicesMessageSender.SendAsync.
                 _ = Task.Run(() => this.classCDeviceMessageSender.SendAsync(cloudToDeviceMessage));
```
atifaziz, line 470: Technically we are calling an interface here and we are relying on the implementation to do the logging. I would lift the exception handler and logging out of `DefaultClassCDevicesMessageSender.SendAsync` and put it here. After `UdpServer` is retired, this will be the only call site.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -25,7 +25,8 @@ public JoinRequestMessageHandler(NetworkServerConfiguration configuration, ILoRa
 
         public void DispatchRequest(LoRaRequest request)
         {
-            _ = Task.Run(async () => await ProcessJoinRequestAsync(request));
+            // Unobserved task exceptions are logged as part of ProcessJoinRequestAsync.
+            _ = Task.Run(() => ProcessJoinRequestAsync(request));
```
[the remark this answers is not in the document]
bastbu: Addressed with a9375acc4c0efccd7e5ccb7be5869a38c5b2bf7a

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: Since that would mean that we refactor many tests and also change the signature of the method (it would not return a `Task<bool>` any longer): What do you think about renaming the method to `TrySendAsync`? That would indicate that it should always succeed, and it would leave the signature intact.

bastbu: Addressed with 1c55bfb109342723c3ac68b6a650843c8a14f7f1

Issue #674, opened 2021-11-01
Add a thread safe cache / DevEUI where we register the incoming fcnt


spygi:
With unit tests.

PR #684, opened 2021-11-01
Cache of DeviceEUI -> concentrator on LNS level

spygi:
# PR for issue #674

## What is being addressed

This PR covers the case of a message that is picked up by 2 concentrators that talk to the same LNS. Here we assume that we have available the deviceEUI on LNS (i.e. the cache has not expired yet)

## How is this addressed

- We add the ConcentratorId to the Request DTO for now with an empty value. 
- We register a new service that handles duplicate messages on creation of the `UdpServer`
- Unit tests are added for this service.

## What is not done yet and will follow in different PRs

- Getting the real concentrator id - is it part of the [data frame](https://doc.sm.tc/station/tcproto.html#lorawan-data-frames)? -> #685 
- Integration tests #686  

## Open questions

- The implementation assumes a single instance of `UdpServer` per LNS. If that's not the case, we would create a singleton cache and use only that one.
-  What is an appropriate expiration policy for the cache entries? At the moment took 24 hours as we do in `LoraDevAddrCache` (in `LoRaADRInMemoryStore` we don't set a limit)
- Should we enable finer-grained lock? Again in `LoRaADRInMemoryStore` we lock on the whole `MemoryCache`
- The check imo should be after the request is validated (where we check also the frame counter)


spygi:
Closing this as it's superseded from #720

PR #688, opened 2021-11-02
Fix e2e ci

Mandur:
I forgot to remove a condition :(


spygi (approved):

spygi:
Just a question - build_and_test job is on another .yaml , does the reference still work?

PR #705, opened 2021-11-03
Feature/703 adjust lora primitives

danigian:
# PR for issue #703 

## What is being addressed

This PR aims to add some helper methods and a FramePort struct for upcoming PR for handling uplink/downlink frames in LNS.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
```diff
@@ -25,6 +25,8 @@ namespace LoRaWan
 
         public DevAddr(uint value) => this.value = value;
 
+        public byte[] AsByteArray() => BitConverter.GetBytes(this.value);
```
atifaziz, line 28: Use `BinaryPrimitives.WriteUInt32LittleEndian` instead like in the rest of the primitives and it's also explicitly clear about the endianness.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Hertz.cs
```diff
@@ -25,6 +25,8 @@ namespace LoRaWan
         public double Mega => this.value / 1e6;
         public double Giga => this.value / 1e9;
 
+        public static Hertz FromMega(double value) => new Hertz((ulong)(value * 1e6));
```
atifaziz, line 28: This will trim. Do you want to round instead? Also, I'd recommend making a `checked` conversion:

```suggestion
        public static Hertz FromMega(double value) => new Hertz(checked((ulong)(value * 1e6)));
```

atifaziz (changes requested):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Hertz.cs
danigian: I think trimming is fine. In the end we are going from a potential XXX.ZZZ value to a XXXZZZ000 value no?

atifaziz: I guess it should be fine for since we multiple by a million before.

atifaziz (approved):
LGTM

Issue #565, opened 2021-10-18
Exception handling Redesign


Mandur:
The current exception handling should be revisited:

- [ ] There should not be any Catch all exception block without rethrowing.
- [x] We should create a `LoRaProcessingException` base type that can be used when throwing "known" exceptions.
- [ ] Only child of this base exception should be caught without being rethrown.
- [x] RunAndQueueNext should catch loraprocessingException

PR #713, opened 2021-11-03
Feature/565 exception handling redesign (part 1)

bastbu:
# PR for issue #565

## What is being addressed

With this PR we address several instances of:
- Logging of exceptions too far down in the stack
- Not bubbling up exceptions
- Not logging unobserved tasks that errored

## How is this addressed

Each commit is a self-contained fix to a certain exception-related problem:

- Logging exception too far down in the stack was addressed by wrapping these exceptions in the newly introduced `LoRaProcessingException` type which contains more information about the error.
- Not bubbling up exceptions was addressed by rethrowing (or wrapping) the exceptions
- Not logging unobserved tasks was addressed by adding a generic try/catch block and logging the exceptions as part of the `when` clause. This ensures that the stack is not unwound at the time of logging and that any scopes that we set are included in the log message for when we start to use structured logging)

There will be follow-up PRs that address the same issues for the sake of having smaller, better reviewable units.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
```diff
@@ -59,34 +58,25 @@ public async Task<string> GetRouterConfigMessageAsync(StationEui stationEui, Can
...
             var queryResult = await this.loRaDeviceApiService.SearchByDevEUIAsync(stationEui.ToString());
+            if (queryResult.Count != 1)
+            {
+                throw new LoRaProcessingException($"The configuration request of station '{stationEui}' did not match any configuration in IoT Hub. If you expect this connection request to succeed, make sure to provision the Basics Station in the device registry.",
+                                                  LoRaProcessingErrorCode.InvalidDeviceConfiguration);
+            }
+
+            var info = queryResult.Single();
+            using var client = this.loRaDeviceFactory.CreateDeviceClient(info.DevEUI, info.PrimaryKey);
+            var twin = await client.GetTwinAsync();
```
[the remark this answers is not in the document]
bastbu: Yes, I pushed the error handling of these exceptions to `LoRaDeviceClient.GetTwinAsync` and wrapped it in a `LoRaProcessingException`

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExceptionFilterUtility.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.NetworkServer
```
Mandur, line 4: Nit should this utility class be shared with other project? If yes, probably in LORaWan?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -210,28 +210,12 @@ public LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClientConnectionMana
...
-            {
-                Logger.Log(this.DevEUI, $"Error while communication with IoT Hub during device initialization. {ex.Message}", LogLevel.Error);
-                throw;
-            }
-            catch (IotHubException ex)
-            {
-                Logger.Log(this.DevEUI, $"An error occured in IoT Hub during device initialization. {ex.Message}", LogLevel.Error);
-                throw;
-            }
+            var twin = await this.connectionManager.GetClient(this)?.GetTwinAsync();
```
Mandur, line 213: I guess the interception of this error is now in the main loop right?

Mandur (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
bastbu: Exactly. This error will only be intercepted on the top level, in order to not log exceptions multiple times.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExceptionFilterUtility.cs
bastbu: Valid point. I think we should do that as soon as it's actually used from somewhere else. If we don't need it, I would suggest to keep the scope of such utilities as small as possible.

PR #715, opened 2021-11-04
Optimize `WebSocketTextChannel` for single reader

atifaziz:
# PR adds to story #676

## What is being addressed

The `WebSocketTextChannel` will only have a single reader at any given time and its internal channel can be optimised for the case.

## How is this addressed

Create the internal channel with the single-reader optimisation options.


p-schuler (approved):
Fantastic! ;)

PR #718, opened 2021-11-04
Feature/645 add custom console logger

p-schuler:
# PR for issue #645 

## What is being addressed

We need a custom logger to tailer the console output to our needs. 

## How is this addressed
Adding an implementation for ILogger that logs to the console. Features that it supports:

- Log to the StdOut and StdErr
- Configuration through IOptionsMonitor allowing to update the configuration and change log levels on the fly
- Scopes are supported to remove all the Logger.Log(DevEUI...) calls we have right now. The DevEUI can simply put into the scope. This will also enable the DevEUI to be available for query in other systems - like app insights


LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<ILoggerProvider, LoRaConsoleLoggerProvider>());
+
+            LoggerProviderOptions.RegisterProviderOptions<LoRaConsoleLoggerConfiguration, LoRaConsoleLoggerProvider>(builder.Services);
+
+            return builder;
+        }
+
+        public static ILoggingBuilder AddLoRaConsoleLogger(this ILoggingBuilder builder,
+                                                            Action<LoRaConsoleLoggerConfiguration> configure)
```
bastbu, line 158: [nit] alignment

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,166 @@
...
+        private string AddScopeInformation(string message)
+        {
+            if (this.provider.ScopeProvider is { } scopeProvider)
+            {
+                scopeProvider.ForEachScope<object>((activeScope, _) =>
+                {
+                    if (activeScope is IDictionary<string, object> activeScopeDictionary &&
+                        activeScopeDictionary.TryGetValue(ILoggerExtensions.DevEUIKey, out var obj) &&
+                        obj is string devEUI &&
+                        !message.StartsWith(devEUI, StringComparison.OrdinalIgnoreCase))
```
bastbu, line 125: Should we test this path here as well?

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan
+{
+    using Microsoft.Extensions.Logging;
+    using System;
```
danigian, line 7: using System should appear before anything else

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,168 @@
...
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Logging.Configuration;
+    using Microsoft.Extensions.Options;
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+
+    public class LoRaConsoleLoggerConfiguration
+    {
+        public LogLevel LogLevel { get; set; }
```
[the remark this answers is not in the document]
p-schuler: We can't. The way the configuration works, is requiring to be able to set the properties on an instance that is managed by configuration and DI.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,166 @@
...
+        public EventId EventId { get; set; }
+        public bool UseScopes { get; set; } = true;
+    }
+
+    public class LoRaConsoleLoggerProvider : ILoggerProvider
```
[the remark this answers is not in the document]
p-schuler: yup, required for testing though.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,166 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan
+{
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Logging.Configuration;
+    using Microsoft.Extensions.Options;
+    using System;
```
danigian, line 11: using System.* should appear before anything else

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+    using System.Collections.Generic;
+
+    public static class ILoggerExtensions
+    {
+        public const string DevEUIKey = "DevEUI";
+
+        public static IDisposable BeginDeviceScope(this ILogger logger, string devEUI) =>
```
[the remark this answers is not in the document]
p-schuler: Not ready unfortunately. The tostring gives us the wrong representation and we can't compare the existing Logger devEUI usage.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -35,7 +36,10 @@ public void ConfigureServices(IServiceCollection services)
 
             _ = services.AddLogging(loggingBuilder =>
                         {
-                            _ = loggingBuilder.SetMinimumLevel((LogLevel)int.Parse(NetworkServerConfiguration.LogLevel, CultureInfo.InvariantCulture));
+                            _ = loggingBuilder.ClearProviders();
+                            var logLevel = (LogLevel)int.Parse(NetworkServerConfiguration.LogLevel, CultureInfo.InvariantCulture);
```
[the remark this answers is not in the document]
p-schuler: Good suggestion. I let you do this as part of the observability work. :)

Tests/Unit/LoggerTests/LoRaConsoleLoggerTests.cs
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Unit.LoggerTests
+{
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Options;
+    using Moq;
+    using System;
```
danigian, line 9: using System.* should appear before anything else

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
danigian, line 14: I think it could be also interesting to have a "StationScope" here. What do you think?

danigian (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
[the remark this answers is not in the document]
danigian: Do we have an issue created for this?
@p-schuler

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
p-schuler: Absolutely. Added.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,166 @@
...
+
+    public static class LoRaConsoleLoggerExtensions
+    {
+        public static ILoggingBuilder AddLoRaConsoleLogger(this ILoggingBuilder builder)
+        {
+            _ = builder ?? throw new ArgumentNullException(nameof(builder));
+
+            builder.AddConfiguration();
+
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<ILoggerProvider, LoRaConsoleLoggerProvider>());
```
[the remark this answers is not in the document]
p-schuler: From the docs: 

Use TryAddEnumerable(IServiceCollection, ServiceDescriptor) when registering a service implementation of a service type that supports multiple registrations of the same service type. Using Add(IServiceCollection, ServiceDescriptor) is not idempotent and can add duplicate ServiceDescriptor instances if called twice. Using TryAddEnumerable(IServiceCollection, ServiceDescriptor) will prevent registration of multiple implementation types.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
p-schuler: Absolutely. As described in the comment, we can extend this. We have to be a bit careful not overloading every entry, on the console though. We will have the scopes include much more and that can be used in any way we want to. 

This is about having the base to replace the current logger .

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.csproj
```diff
@@ -1,9 +1,12 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
atifaziz, line 1: Revert unrelated change.

Tests/Unit/LoRaWan.Tests.Unit.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
atifaziz, line 1: Revert unrelated change.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
```
atifaziz, line 3: Would be good to add `#nullable enable` for new code.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
```diff
@@ -0,0 +1,166 @@
...
+        public IDisposable BeginScope<TState>(TState state) =>
+            this.provider.ScopeProvider is { } scopeProvider ? scopeProvider.Push(state) : default;
+
+        public bool IsEnabled(LogLevel logLevel) => logLevel >= this.provider.LogLevel;
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel)) return;
+            _ = formatter ?? throw new ArgumentNullException(nameof(formatter));
```
atifaziz, line 87: Argument checks should come first:

```suggestion
            _ = formatter ?? throw new ArgumentNullException(nameof(formatter));

            if (!IsEnabled(logLevel))
                return;
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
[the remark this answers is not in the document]
atifaziz: > The tostring gives us the wrong representation and we can't compare the existing Logger devEUI usage.

We should be adding support for things, such as formats and representations, as needed by features.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
p-schuler: It's ordered alphabetical - if we want to enforce the ordering differently, we should setup the tooling to do so. I was thinking to re-order and have system on top, however, that is limiting the tooling support also for VS, where we can simply cleanup the usings with a single click and they get ordered alphabetically. For productivity, I keep this.

Tests/Unit/LoggerTests/LoRaConsoleLoggerTests.cs
p-schuler: see above

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
p-schuler: Yeap - I'll add this to the template header in the editorconfig for consistency in the future.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoRaConsoleLogger.cs
p-schuler: As discussed: this was pure perf motivated. However, we can use the DebugLogger?.Log design to overcome putting things on the stack and then this becomes a non-issue.

Tests/Unit/LoRaWan.Tests.Unit.csproj
p-schuler: As discussed, I can't do much about this in VS. If we want to standardize on the encoding, we should do this in editorconfig. Manual fiddling with this, is not what we should do. Agreed with @atifaziz that he is looking into it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/ILoggerExtensions.cs
Mandur: I will take care of this Once this damn branch works

p-schuler: actually, found the setting that we were looking for and added it to the .editorconfig. With that we have VS support and they get ordered System first alphabetically.

Issue #389, opened 2021-09-28
[Story] Up (Class A and C) and downstream (Class A) support


roel4ez:
Pre-requisites:
- [x] #651
- [x] #703
- [x] #706
- [x] #708
- [x] #710
- [x] #725

PR #720, opened 2021-11-05
Concentrator deduplication service based on the UpstreamDataFrame

spygi:
# PR for issue #674

## What is being addressed

Multiple LBSs can be connected to one LNS. In this scenario, LNS will receive the same leaf device message multiple times. The added service with this PR ensures that we send upstream messages that are really unique, based on the fields of the `UpstreamDataFrame` (currently under review in #706 ).

## How is this addressed

- New service that is unit-tested but not yet integrated in the `LnsProtocolMessageProcessor` (pending work tracked in #389 )
- The service uses an `InMemoryCache` where keys are based on the received message and the value is the station EUI that last processed the same message.
  - The entries in this cache have a sliding expiration of 1 minute.
- For now, the simplest path is taken with no connectivity checks. A connectivity check will be added in the next PR, as we need to have first a proper mechanism to get open sockets from the `WebSocketWriterRegistry` -> #724


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/UpstreamDataFrame.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public UpstreamDataFrame(DevAddr devAddr, ushort frameCounter, string fRMPayload, Mic mic)
+        {
+            DevAddr = devAddr;
+            FrameCounter = frameCounter;
+            FRMPayload = fRMPayload;
+            Mic = mic;
+        }
+
+        public virtual DevAddr DevAddr { get; }
```
spygi, line 16: These properties were marked as virtual so that we could setup a mock on them. Is there a way to avoid this? It's also marked as an issue from CodeQL scanning.

Mandur: An interface would work, I don't know other way. That's an interesting question

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,69 @@
...
+                return false;
+            }
+
+            this.logger.LogInformation($"Duplicate message detected from DevAddr: {updf.DevAddr}, dropping.");
+            return true;
+        }
+
+        internal static string CreateCacheKey(UpstreamDataFrame updf)
+        {
+            using var sha256 = SHA256.Create();
```
Mandur, line 55: Do you know how performant is this creation? Would it make sense to keep a reference on the class to avoid recreating it on every request?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public bool ShouldDrop(UpstreamDataFrame updf, StationEui stationEui)
+        {
+            if (updf == null) throw new ArgumentNullException(nameof(updf));
+
+            var key = CreateCacheKey(updf);
+            StationEui previousStation;
+
+            lock (Cache)
```
Mandur, line 34: I was looking at AddOrGetExisting to replace this bit but that looks tricky. I am not a big fan of locking here, as this will be called very often. I was wondering maybe we could always add it and if the return is null get the value? (as memache is supposed to be thread safe)
@atifaziz @p-schuler

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/UpstreamDataFrame.cs
spygi: Would work yes but not sure if an interface is a good choice for a DTO class like this.

atifaziz: If it's a pure data class that has no behavioural component then it doesn't need to be mocked at all therefore don't need an interface or make them virtual.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public bool ShouldDrop(UpstreamDataFrame updf, StationEui stationEui)
+        {
+            if (updf == null) throw new ArgumentNullException(nameof(updf));
+
+            var key = CreateCacheKey(updf);
+            StationEui previousStation;
+
+            lock (this.Cache)
```
atifaziz, line 36: `MemoryCache` is thread-safe so seems odd to me to have to lock on it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            this.Logger.LogInformation($"Duplicate message detected from DevAddr: {updf.DevAddr}, dropping.");
+            return true;
+        }
+
+        internal static string CreateCacheKey(UpstreamDataFrame updf)
+        {
+            using var sha256 = SHA256.Create();
+            var key = sha256.ComputeHash(Encoding.UTF8.GetBytes(string.Join("", updf.DevAddr, updf.FrameCounter, updf.FRMPayload, updf.Mic)));
```
atifaziz, line 58: The `string.Join` here will cause lots of allocations, including conversions that could be sensitive to culture. Can we not take advantage of the `Write` method on each primitive to write out the bytes?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
atifaziz: Just bear in mind, `SHA256` is _not_ thread-safe. If you want to create it once and reuse it, safest is to make it a thread-static via `[ThreadStatic]`. See [C# Cautionary Tale — The dangers of SHA256 reuse](https://wanderingdeveloper.medium.com/c-cautionary-tail-the-dangers-of-sha256-reuse-2b5bb9c6fde9) for background, details and solution.

Tests/Unit/NetworkServerTests/ConcentratorDeduplicationTest.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace LoRaWan.Tests.Unit.NetworkServerTests
+{
+    using LoRaWan.NetworkServer;
+    using LoRaWan.NetworkServer.BasicsStation;
+    using Microsoft.Extensions.Logging.Abstractions;
+    using Moq;
+    using System;
+    using System.Threading.Tasks;
+    using Xunit;
```
atifaziz, line 12: Sort so `System.*` namespaces are listed first:

```suggestion
    using System;
    using System.Threading.Tasks;
    using LoRaWan.NetworkServer;
    using LoRaWan.NetworkServer.BasicsStation;
    using Microsoft.Extensions.Logging.Abstractions;
    using Moq;
    using Xunit;
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IConcentratorDeduplication.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.NetworkServer
+{
+    using LoRaWan.NetworkServer.BasicsStation;
+
+    public interface IConcentratorDeduplication
```
atifaziz, line 8: Is this interface used anywhere?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace LoRaWan.NetworkServer
+{
+    using LoRaWan.NetworkServer.BasicsStation;
+    using Microsoft.Extensions.Caching.Memory;
+    using Microsoft.Extensions.Logging;
+    using System;
+    using System.Security.Cryptography;
+    using System.Text;
+
+    public sealed class ConcentratorDeduplication : IConcentratorDeduplication, IDisposable
```
atifaziz, line 13: This does de-duplication of uplink data frames rather than _concentrators_ so isn't it better to call it `UplinkDataFrameDeduplicator`?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IConcentratorDeduplication.cs
spygi: Not yet, will be registered as a singleton after work done in #389 is finished.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/UpstreamDataFrame.cs
spygi: It is a pure data class.

The `virtual` was added for the test method `ConcentratorDeduplicationTest.CachedEntries_Should_Expire`: do you have a better idea how we can test this without making the properties virtual? Or do you think we should drop the test alltogether (imo, I think it adds value)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        internal readonly MemoryCache Cache;
+
+        private readonly ILogger<IConcentratorDeduplication> Logger;
```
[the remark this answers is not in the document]
spygi: I was following the [capitalization guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers) but looking at the rest of the codebase it seems we tend to use camelCase indeed - will amend.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,71 @@
...
+    using LoRaWan.NetworkServer.BasicsStation;
+    using Microsoft.Extensions.Caching.Memory;
+    using Microsoft.Extensions.Logging;
+    using System;
+    using System.Security.Cryptography;
+    using System.Text;
+
+    public sealed class ConcentratorDeduplication : IConcentratorDeduplication, IDisposable
+    {
+        internal readonly MemoryCache Cache;
```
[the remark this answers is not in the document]
spygi: It does. It was done for testing purposes (to check what is inside the `Cache`) - do you have a better idea or shall we not care about the internal state when unit testing?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        internal static string CreateCacheKey(UpstreamDataFrame updf)
+        {
+            using var sha256 = SHA256.Create();
+            var key = sha256.ComputeHash(Encoding.UTF8.GetBytes(string.Join("", updf.DevAddr, updf.FrameCounter, updf.FRMPayload, updf.Mic)));
+
+            return BitConverter.ToString(key);
+        }
+
+        private void AddToCache(string key, StationEui stationEui)
```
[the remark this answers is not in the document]
spygi: It was kept as a separate method as it will be used in a couple of places in the follow-up PR that checks connectivity #724 (though could be argued that we can add it then)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Hm, I see your point. The reasoning was that the criterion for the deduplication is from which concentrator we got it from. This is important in order to differentiate with the other deduplication service that is based on the LNS used to send a message upstream (and utilizes the Azure Function). I think using only `UplinkDataFrame` in the name will not provide enough semantics.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Absolutely, if possible we should avoid locking. 

Note that `AddOrGetExisting` is part of `System.Runtime.Caching` namespace while we are using the recommended [according to the docs](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-5.0#systemruntimecachingmemorycache] `Microsoft.Extensions.Caching.Memory.MemoryCache` 

For the `Microsoft.Extensions.Caching.Memory.MemoryCache` there is no mention in [the docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycache?view=dotnet-plat-ext-5.0) of it being thread-safe. Looking at [the source](https://source.dot.net/#Microsoft.Extensions.Caching.Memory/MemoryCache.cs,99099576db12fef5) there is also no locking primitives used (other than some `Interlocked` usages for size counters) so not sure we can avoid locks if we stick with this MemoryCache. 

Checking quickly also the source for the [other MemoryCache](https://source.dot.net/#System.Runtime.Caching/System/Runtime/Caching/MemoryCache.cs,85751dcd91d94103) it's also not clear to me how it's supposed to be "thread-safe" (in quotes because I think the term is overloaded and different people mean different things with it). 

/cc @atifaziz because he had a similar comment, to keep the discussion in a single place.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -0,0 +1,71 @@
...
+    public sealed class ConcentratorDeduplication : IConcentratorDeduplication, IDisposable
+    {
+        internal readonly MemoryCache Cache;
+
+        private readonly ILogger<IConcentratorDeduplication> Logger;
+        private readonly TimeSpan CacheEntryExpiration;
+
+        public ConcentratorDeduplication(
+            ILogger<IConcentratorDeduplication> logger,
+            int cacheEntryExpirationInMilliSeconds = 60_000)
```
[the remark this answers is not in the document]
spygi: Agree about the order of logger. 

About `TimeSpan` - it can't be used as an optional parameter because there is no compile time constant (afaik). I need to provide this parameter only from the tests in order to test the expiration policy (in a reasonable time).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: I tagged you in the other thread from Mik to keep the discussion in one place. I believe the [specific implementation](https://source.dot.net/#Microsoft.Extensions.Caching.Memory/MemoryCache.cs,99099576db12fef5) we use is not thread-safe.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
Mandur: Pfioue looks liek I was wrong, thank you Atif for the hint :)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
atifaziz: The guidelines only talk about _public_ members:

> ✔️ DO use PascalCasing for all **public** member, type, and namespace names consisting of multiple words.
>
> ✔️ DO use camelCasing for parameter names.

_Conventionally_ however, _private fields_ have always used camel-casing in 3 variations:

1. `camelCase`
2. `_camelCase`
3. `m_camelCase`

In this project, 1 seems to be the dominant convention.

Also have a look at [Names of Fields](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields):

> The field-naming guidelines apply to static public and protected fields. Internal and private fields are not covered by guidelines, and public or protected instance fields are not allowed by the member design guidelines.
> 
> ✔️ DO use PascalCasing in field names.
> 
> ✔️ DO name fields using a noun, noun phrase, or adjective.
> 
> ❌ DO NOT use a prefix for field names.
> 
> For example, do not use "g_" or "s_" to indicate static fields.

Note again, no guideline is offered for _private_ fields.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/UpstreamDataFrame.cs
atifaziz: Do you mean `CacheKey_Should_Consider_All_Required_Fields`?

spygi: Sorry 🤦‍♂️ yes, you are right

atifaziz: If you did indeed mean `CacheKey_Should_Consider_All_Required_Fields` then I think the test should test the generated cache key instead of testing that the properties of the object were simply accessed. Access to properties does not guarantee that they are part of the key. Consequently, you can make the properties non-virtual.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
atifaziz: Unfortunately, It's thread-safe by virtue of using [`ConcurrentDictionary<,>`](https://github.com/dotnet/runtime/blob/d40f560efbf4f85ec6d59892a6c4bafa39f66d19/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs#L26) for the entries, which is probably why you don't see direct use of any locking primitives.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
spygi: I will resolve this to clear the converstation but let me know if you disagree.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
spygi: Done in 168e2451b7956aa6c4fb48cdd30c7fc7601fa782

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
atifaziz: The traditional way to deal with this is to use constructor overloading and chaining:

```c#
public ConcentratorDeduplication(ILogger<IConcentratorDeduplication> logger) :
    this(logger, TimeSpan.FromMinutes(1)) { }

public ConcentratorDeduplication(ILogger<IConcentratorDeduplication> logger,
                                 TimeSpan cacheEntryExpiration)
{
    this.Cache = new MemoryCache(new MemoryCacheOptions());
    this.Logger = logger;
    this.CacheEntryExpiration = cacheEntryExpiration;
}
```

Another option, if you wish to stick with optional arguments, is to make `TimeSpan` nullable and then use `null` as the default value. In the implementation body of the constructor, you can then default a specific value in case of `null`. Below is an example of what I mean:

```c#
public ConcentratorDeduplication(ILogger<IConcentratorDeduplication> logger,
                                 TimeSpan? cacheEntryExpiration = null)
{
    this.Cache = new MemoryCache(new MemoryCacheOptions());
    this.Logger = logger;
    this.CacheEntryExpiration = cacheEntryExpiration ?? TimeSpan.FromMinutes(1);
}
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/UpstreamDataFrame.cs
spygi: Agree that it's not proof that they are part of the key. [Changed the test to test for an expected result instead and made the properties non-virtual again](47dbfc79f9c0a74442924160e7b9178d75616bc6 ).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: I worked on 2 variants: 
- [One is using the `Write` methods](0bc624bb9b8170bd143b63fdc20630b93f3ac4d8) when we have a primitive (e.g. the payload is a simple `string`). This implementation avoids new allocations by using `AsSpan()` (if my understanding is right) and also uses minimum amount of memory by allocating a buffer with the required length (again if my pointer arithmetics are correct). 
- [The second implementation uses a stringbuilder](bad5c28b9c7a5ebc8a56cd149af6d4f218a821e9) which allocates more memory (due to resizing) but is a lot simpler to understand and less error-prone. 

Imo we should go with the second implementation unless you think this is a critical path we must optimize heavily. Wdyt?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
spygi: Thanks very much for the snippets - I have seen these patterns but it hadn't clicked why we need those. Went with the second variant in df55a3bcea66a1a83f2d54939f71415da3595a7a

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IConcentratorDeduplication.cs
spygi: I will resolve this but feel free to reopen, I am interested in the answer.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Resolving this but feel free to reopen if you want to.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: I was debating whether the proposed solution was worth the complexity but at the end [changed to use the ThreadStatic field](c5885a51693f47307a8df5e4999cf66e9c91c75c) .

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Removed the locking with 227773b30e564a620e3a8868aa2efad23710b28a. If there are any resources, would like to read more about it this cache and thread-safety (not the one part of `System.Runtime`).

I noticed another occurrence of locking on this memory cache in `LoRaADRInMemoryStore` (possibly in other places too)

Tests/Unit/NetworkServerTests/ConcentratorDeduplicationTest.cs
```diff
@@ -0,0 +1,95 @@
...
+            var stationEui = new StationEui();
+
+            // act
+            _ = sut.ShouldDrop(defaultUpdf, stationEui);
+
+            // assert
+            await Task.Delay(delayInMilliseconds);
+            var key = ConcentratorDeduplication.CreateCacheKey(defaultUpdf);
+            Assert.Equal(expectedResult, sut.Cache.TryGetValue(key, out var _));
+        }
```
spygi, line 91: Leaning towards deleting this test that depends on timing. Any alternative ideas on how we can unit test cache expiration?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: The solution with ThreadStatic backing static field was marked (correctly imo) from CodeQL. 

Using simply a static ThreadStatic member (without a property that checks if it's initialized), doesn't work either from tests since static members are initialized exactly once from the compiler (subsequent ConcentrationDeduplication objects will not have this member initialized)

Reverted to the original code that creates a new instance of Sha256 to avoid this complexity.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
atifaziz: > * [The second implementation uses a stringbuilder](bad5c28b9c7a5ebc8a56cd149af6d4f218a821e9) which allocates more memory (due to resizing) but is a lot simpler to understand and less error-prone.

The link was failing for me, but I think you meant to refer to commit bad5c28b9c7a5ebc8a56cd149af6d4f218a821e9?

spygi: As discussed offline will go with the first option (modified).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: To sum-up our discussion with @atifaziz : we agree that the desired semantics of a deduplication is "whoever is first to read or write wins". Given that, we do not need pessimistic locking and it was correctly (based on our current knowledge) removed. The question of whether the class is or not thread-safe does not affect the solution.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
spygi: We decided to mark private and use an instance provided from the DI/tests instead.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Reverted back to using ThreadStatic as discussed with Atif.

PR #721, opened 2021-11-05
Add docker compose and bake

Mandur:
# PR for issue #585

## What is being addressed

CI should build containers for the lora network server.

## How is this addressed

We build the container for ARM and AMD64 as part of a new job in the CI.
I also built during the tests a docker-compose.yaml that could be useful for a release pipeline. Therefore, I also put it on the pr. (Docker-compose build will build all the containers of the solution, but that doesnt' work on GH action)

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/Dockerfile.arm32v7
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/Dockerfile.arm32v7
@@ -1,23 +1,34 @@
-FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build-env
+FROM mcr.microsoft.com/dotnet/sdk:5.0-bullseye-slim AS build-env
 
 WORKDIR /build
-
-WORKDIR /build/LoRaEngine/modules/LoRaWanNetworkSrvModule/
 COPY Directory.Build.props ./
 COPY AssemblyInfo.cs ./
 COPY .editorconfig ./
-COPY ./LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger ./Logger
-COPY ./LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools ./LoraTools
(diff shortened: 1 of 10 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/Dockerfile.arm32v7
```diff
@@ -1,23 +1,34 @@
...
 
-RUN dotnet publish -c Release -o out
+RUN dotnet publish "LoRaWanNetworkSrvModule.csproj" -c Release -o out --no-restore
 
-FROM mcr.microsoft.com/dotnet/aspnet:3.1-buster-slim-arm32v7
+FROM mcr.microsoft.com/dotnet/aspnet:3.1-bullseye-slim-arm32v7
 WORKDIR /app
 COPY --from=build-env /build/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWanNetworkSrvModule/out/* ./
+
+RUN useradd -ms /bin/bash moduleuser
```
danigian, line 31: What is this for?

danigian (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/Dockerfile.arm32v7
Mandur: I don't know thought it was already there. Most likely a spillover from another I will fix thanks for the catch

Mandur: Interestingly this is a spillover from the amd64 dockerfiles. I don't think it's needed, and I would remove this everywhere. I found some ref, I don't think it's needed. WDYT?
https://docs.microsoft.com/en-us/azure/iot-edge/tutorial-machine-learning-edge-06-custom-modules?view=iotedge-2020-11#add-directory-to-the-module

danigian: As discussed, would make sense to keep it for LNS module, but not for BasicStation one imho

PR #731, opened 2021-11-05
Introducing IoT Central as Device Registry Manager

kbeaugrand:
# PR for supporting Azure IoT Central for IoT device management

## What is being addressed

I don't know if this kind of feature was expected. But I was interressted to test this Starter Kit with IoT central for demonstration of the possibilities.

The objective was to adapt this project to enable the same deployment for Azure IoT Central for managing devices.

## How is this addressed

I created a dedicated interface for RegistryManger in the Facade with 2 implementations : 
- One for IoT Hub 
- One for IoT Central (based on REST APIs that are existing)

Since IoT Central leverage on IoT Hub and IoT DPS, The facade is able to registry dynamically the device during the LoRa Join step. But since the IoT Hub is assigned by the DPS, I had to provide as an outpout the IoT Hub Hostname that is assigned to the device (For IoT Hub implementation, I provided the IoT Hub that is used by the registry manager).

The network server catch the Assigned IoT Hub from the Join information to create the DeviceClient instance.

-> The API Version should be changed by doesn't create hard changes since this is only a new field taht is provided by the Facade.

## Unit Tests

I added some unit test to validate and verify the behavior of IoT Hub implementation and IoT Central's one.
I adpated existing UnitTests to fullfill the requirements.

**However I did not create or adapt any Integration Tests**

## What is missing

I sucessfully tested the device enrollement and device connectivity throught the Edge Gateway. 
It miss some E2E tests and automatic provisionning and deployment of the environment to accelerate the usage.

It might miss some Unit Test also as I see in test coverage that increase the line missing coverage...

Please let me know If I can do more ?


kbeaugrand:
Hi @spygi, 

We discussed today about the possibility to publish LoRa Device DTDL to get an automated import of the device model once the device is connected and so on... 

Unfortunatly, I still didn't find a possibility to upload the views of the devices. 
But, I foudn the public repository of Microsoft Azure Plug And Play models that could be use to add the LoRa Device interface at least : [https://github.com/Azure/iot-plugandplay-models](https://github.com/Azure/iot-plugandplay-models)

Maybe it could be interesting to discuss internally with the team to know how it's possible to publish views and others... 

More info at: [https://docs.microsoft.com/en-us/azure/iot-develop/concepts-model-repository](https://docs.microsoft.com/en-us/azure/iot-develop/concepts-model-repository)

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var results = new List<IDeviceTwin>();
+            if (!this.HasMoreResults)
+            {
+                return results;
+            }
+
+            HttpResponseMessage result = null;
+
+            result = await this.client.GetAsync(new Uri(this.client.BaseAddress,
```
Mandur, line 40: I am seeing here that we load the whole devices from IoT Central and then iterate locally to find the matchin one. have you considered using a [query ](https://docs.microsoft.com/en-us/rest/api/iotcentral/1.1-previewdataplane/query/run) 
(e.g. SELECT DevAddr FROM dtmi:modelDefinition:ru5kifjr:ns8lsni8q8 WHERE $id='<id>'" ) to find the correct one immediately and resolve this in IoT central?

kbeaugrand: Hi @Mandur, 

You're right, this is a better option. 
However, I didn't found a way on IoT Central APIs to query the registry.
Please find more info at [https://docs.microsoft.com/en-us/rest/api/iotcentral/](https://docs.microsoft.com/en-us/rest/api/iotcentral/)

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/IoTCentralDeviceRegistryManager.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (deviceRequest.StatusCode == HttpStatusCode.NotFound)
+            {
+                return null;
+            }
+
+            var deviceResponse = await deviceRequest.Content.ReadAsAsync<Device>();
+
+            var attestation = this.provisioningHelper.ProvisionDevice(deviceId, out var assignedIoTHubHostname);
+
+            return new IoTCentralDevice(deviceResponse, attestation, assignedIoTHubHostname);
```
Mandur, line 49: If attestation is null (device could not be provisioned) you probably want to return null here as well

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceProvisioningHelper.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            using var transportHandler = new ProvisioningTransportHandlerHttp();
+            using var securityProvider = new SecurityProviderSymmetricKey(deviceId, primaryKey: attestation.SymmetricKey.PrimaryKey, secondaryKey: attestation.SymmetricKey.SecondaryKey);
+            var provisioningClient = ProvisioningDeviceClient.Create(
+              this.deviceProvisioningEndpoint,
+              this.provisioningScopeId,
+              securityProvider,
+              transportHandler);
+
+            var registerResult = provisioningClient.RegisterAsync().Result;
```
Mandur, line 41: We should use async here and rename the method as Async

kbeaugrand: This is not as simple as that, this method returns the assigned iot hub as an output parameter.
This is not compatible with async method...

Mandur: What about creating the IoT CentralDevice object in there instead of returning a null value? I am not a big fan of blocking the thread here

kbeaugrand:
> What about creating the IoT CentralDevice object in there instead of returning a null value? I am not a big fan of blocking the thread here 

I Agree, I'll do that

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceProvisioningHelper.cs
kbeaugrand: I implemented your suggestion.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/IoTCentralDeviceRegistryManager.cs
kbeaugrand: I implemented your suggestion

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
kbeaugrand: Hi @Mandur, 

The Azure support team responded that the IoT Central API does not support querying devices in multiple models. 

I created an open question at [https://docs.microsoft.com/en-us/answers/questions/633800/azure-iot-central-query-across-multiple-device-tem.html](https://docs.microsoft.com/en-us/answers/questions/633800/azure-iot-central-query-across-multiple-device-tem.html). 

I hope they can take a look on that and move this question from backlog to their next sprints...

kbeaugrand:
Hi all, 

I put here a little documentation about starting with this feature (create a device template, import components), to get everything working for testing purpose.

## Create a new Device Template

On left bar, select the "**Device templates**" menu:
image

then click on "**New**" button:
image

On "**Select type**" screen, click on "**IoT Device**" custom template: 
image

Then click to "**Next: Customize...**" button:
image

Type your device template name:
image

Then click to "**Next: Review**" button: 
image

Validate the creation by clicking on "**Create**" button: 
image

Select the "Custom model" device template creation mode: 
image

On the device template edit screen, select the "**Add component**" menu item:
image

For the component creation, select the "Import interface" option: 
image

Select the [LoraDevice.dtdl.json](https://raw.githubusercontent.com/Azure/iotedge-lorawan-starterkit/feature/iot-central/Template/LoraDevice.dtdl.json) from this pull request:

Once imported, the device template will be populated with the "LoRa Device" interface as below: 
image

Then, you can publish the device template by clicking on "**Publish**" button: 
image

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
kbeaugrand: Hi @Mandur, 

I pushed an update to use query API rather than looping on devices...

However, since query APIs doesn't giv e the possibility to search accross multiple templates, I realized a loop on device templates that uses the LoRa Device component...

The result will be as many query as device templates that fit the previous request.

I think we can enhance this part with caching the device templates results. During my tests I saw that querying the device templates could take more than 10 secs rather that querying the devices (by /query) runs in less than 1 sec...

Mandur:
@Ellerbach As discussed could you help us reviewing this contribution from an IoT Central PoV?

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
```diff
@@ -54,8 +49,9 @@ public CreateEdgeDevice(RegistryManager registryManager)
 
             // Get function facade key
             var base64Auth = Convert.ToBase64String(Encoding.Default.GetBytes($"{publishingUserName}:{publishingPassword}"));
-            var apiUrl = new Uri($"https://{Environment.GetEnvironmentVariable("WEBSITE_CONTENTSHARE")}.scm.azurewebsites.net/api");
-            var siteUrl = new Uri($"https://{Environment.GetEnvironmentVariable("WEBSITE_CONTENTSHARE")}.azurewebsites.net");
+            var apiUrl = new Uri($"https://{GetEnvironmentVariable("FACADE_HOST_NAME")}.scm.azurewebsites.net/api");
```
Ellerbach, line 52: is that wise to change the environment variable? Isn't this going to create a breaking change?

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -42,17 +53,12 @@ public override void Configure(IFunctionsHostBuilder builder)
...
-#pragma warning disable CA2000 // Dispose objects before losing scope
-            // Object is handled by DI container.
-            _ = builder.Services.AddSingleton(RegistryManager.CreateFromConnectionString(iotHubConnectionString));
-#pragma warning restore CA2000 // Dispose objects before losing scope
             builder.Services.AddAzureClients(builder =>
-            {
-                _ = builder.AddBlobServiceClient(configHandler.StorageConnectionString)
-                           .WithName(WebJobsStorageClientName);
-            });
+                       {
```
Ellerbach, line 57: Seems that something is odd with the indentation, maybe just related to the GitHub rendering. I guess a ctrl + k + d from Visual Studio will just fix those:

image

LoRaEngine/LoraKeysManagerFacade/DeviceRegistryMode.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
+{
+    using System.Runtime.Serialization;
+
+    internal enum DeviceRegistryMode
+    {
+        IoTHub,
+        IoTCentral
```
Ellerbach, line 11: More than IoT Central, it's DPS that uses the mechanism. So what is specific to IoT Central that you have to do here and not just related to DPS?
And this does apply to the URL and other elements like this one. I may point some out as well.

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -25,12 +31,17 @@ public override void Configure(IFunctionsHostBuilder builder)
...
-
             var iotHubConnectionString = configHandler.IoTHubConnectionString;
-            if (iotHubConnectionString == null)
+            var iotCentralEndpoint = configHandler.IoTCentralEndpoint;
+
+            if (configHandler.DeviceRegistryMode == DeviceRegistryMode.IoTHub && iotHubConnectionString == null)
             {
                 throw new InvalidOperationException($"Missing {ConfigHandler.IoTHubConnectionStringKey} in settings");
             }
+            else if (configHandler.DeviceRegistryMode == DeviceRegistryMode.IoTCentral && iotCentralEndpoint == null)
```
Ellerbach, line 41: related to a previous conversation as IoT Central is using DPS, what is IoT Central related that cannot be just DPS?

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -67,21 +73,60 @@ public override void Configure(IFunctionsHostBuilder builder)
...
             internal const string StorageConnectionStringKey = "AzureWebJobsStorage";
 
             internal static ConfigHandler Create(IFunctionsHostBuilder builder)
             {
                 var tempProvider = builder.Services.BuildServiceProvider();
                 var config = tempProvider.GetRequiredService<IConfiguration>();
 
-                var iotHubConnectionString = config.GetConnectionString(IoTHubConnectionStringKey);
-                if (!string.IsNullOrEmpty(iotHubConnectionString))
+                var redisConnectionString = config.GetConnectionString(RedisConnectionStringKey);
```
Ellerbach, line 128: why do you renamed this variable?

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -95,6 +140,16 @@ internal static ConfigHandler Create(IFunctionsHostBuilder builder)
 
             internal abstract string IoTHubConnectionString { get; }
 
+            internal abstract string IoTCentralToken { get; }
```
Ellerbach, line 143: All those are more DPS related than IoT Central central related. So I would rename all to be DeviceProvisioningService rather than IoTCentral. But again, I may miss something specific somewhere.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/Device.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade.IoTCentralImp.Definitions
```
Ellerbach, line 4: On top of the previous comments about IoT Central vs DPS, I would avoid abbreviations. So IotCentral should be enough or DeviceProvisioningService. In this case Imp (or Implementation) doesn't really bring any value in the namespace. And I would challenge as well Definitions. What real value does it bring?

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/Device.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public class Device
```
Ellerbach, line 6: missing /// comments on anything public. This applies to any other public element in the rest of the code as well
note: you can do in the interface as well

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwin.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public DeviceTwin(string deviceName, DesiredProperties properties)
+        {
+            this.DeviceId = deviceName;
+            this.properties = properties;
+        }
+
+        public string DeviceId { get; }
+
+        public string GetDevAddr()
```
Ellerbach, line 22: you can simplify a bit all those single line returns:

```csharp
public string GetDevAddr() => this.properties.DevAddr;
```

I makes it more concise and easy then to read.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/IoTCentralDeviceRegistryManager.cs
```diff
@@ -0,0 +1,113 @@
...
+    using System.Linq;
+    using System.Net;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+    using LoraKeysManagerFacade.IoTCentralImp.Definitions;
+    using Newtonsoft.Json.Linq;
+
+    internal class IoTCentralDeviceRegistryManager : IDeviceRegistryManager
+    {
+        private const string API_VERSION = "api-version=1.1-preview";
```
Ellerbach, line 17: Looking at the code, you don't have any other way to override this constant. It would be nice to check if an environment variable exists for this one and override it. Otherwise, if the API is changing , go for a release and API are compatible, you'd have to fully rebuild all. 
Note: I would most likely not leave this comment if it would be a non preview API

Template/EdgeLoraGateway.dtdl.json
```diff
@@ -0,0 +1,92 @@
+[
```
Ellerbach, line 1: good thoughts on the dtdl side!

Ellerbach (commented):
Overall, looks good. Main question is on the naming and what's associated with it. IoT Central uses DPS. So I have a bit hard time to decorrelate what is IoT Central specific and what is DPS related.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
Ellerbach: As IoT Central is using DPS, most likely DPS API should help getting devices, adding, listing them: https://docs.microsoft.com/en-us/rest/api/iot-dps/

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
atifaziz: I would also read it once into a variable and use the variable in the construction of the `Uri` instances so the environment variable name and reading doesn't have to be duplicated.

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
```diff
@@ -75,83 +71,17 @@ public CreateEdgeDevice(RegistryManager registryManager)
...
-                    _ = await this.registryManager.AddDeviceAsync(abpDevice);
-                    var abpTwin = new Twin();
-                    abpTwin.Properties.Desired = new TwinCollection(@"{AppSKey:'2B7E151628AED2A6ABF7158809CF4F3C',NwkSKey:'3B7E151628AED2A6ABF7158809CF4F3C',GatewayID:'',DevAddr:'0228B1B1',SensorDecoder:'DecoderValueSensor'}");
-                    var abpRemoteTwin = await this.registryManager.GetTwinAsync(AbpDeviceId);
-                    _ = await this.registryManager.UpdateTwinAsync(AbpDeviceId, abpTwin, abpRemoteTwin.ETag);
-                }
+                await this.registryManager.CreateEdgeDeviceAsync(
+                    deviceName,
+                    deployEndDevice,
+                    new Uri($"{siteUrl}/api/"),
```
atifaziz, line 79: The better way to combine URLs is to use a [`Uri` constructor that accepts a base and relative part](https://docs.microsoft.com/en-us/dotnet/api/system.uri.-ctor#System_Uri__ctor_System_Uri_System_String_):

```suggestion
                    new Uri(siteUrl, "/api/"),
```

LoRaEngine/LoraKeysManagerFacade/DeviceFeedConstants.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
+{
+    /// <summary>
+    /// These constants are use when user selected the DEPLOY_DEVICE option while created a Edge Gateway
+    /// This is for demo purpose only and should not be used for productive workloads.
+    /// </summary>
+    internal class DeviceFeedConstants
```
atifaziz, line 10: This should be a `static` class since it's never expected to be instantiated:

```suggestion
    internal static class DeviceFeedConstants
```

I wonder also why not give it a clear name, like `DemoDeviceIds?` I mean I would make their _demo_ purpose stand out.

LoRaEngine/LoraKeysManagerFacade/DeviceFeedConstants.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public const string AbpDeviceId = "46AAC86800430028";
```
atifaziz, line 12: I would like to see primitives like `DevEui` being used in new code rather than degenerating to strings.

LoRaEngine/LoraKeysManagerFacade/DeviceFeedConstants.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
```
atifaziz, line 4: Enable nullable context in new source files:

```suggestion
#nulleable enable

namespace LoraKeysManagerFacade
```

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -67,21 +73,60 @@ public override void Configure(IFunctionsHostBuilder builder)
                     .AddSingleton<IFunctionBundlerExecutionItem, ADRExecutionItem>()
                     .AddSingleton<IFunctionBundlerExecutionItem, PreferredGatewayExecutionItem>()
                     .AddSingleton<LoRaDevAddrCache>();
+
+
+            if (configHandler.DeviceRegistryMode == DeviceRegistryMode.IoTHub)
+            {
+                var regex = new Regex("HostName=(.*[.]azure[-]devices[.]net);");
```
atifaziz, line 80: Improvements: 
```suggestion
                var regex = new Regex(@"\bHostName=(.+\.azure-devices\.net);", RegexOptions.IgnoreCase);
```

- Make the C# string _verbatim_ with `@`
- Require `HostName` match on a word boundary.
- `*` should be `+` so it does not match empty
- Escape `.` with `\.`
- `-` doesn't need to be escaped
- Use `RegexOptions.IgnoreCase | RegexOptions.CultureInvariant` for case-insensitive matches in the invariant culture.

atifaziz: Also don't you think it'll be simpler and safer to use [`IotHubConnectionStringBuilder.Create`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.client.iothubconnectionstringbuilder.create?view=azure-dotnet#Microsoft_Azure_Devices_Client_IotHubConnectionStringBuilder_Create_System_String_) to parse the string and then access [its `HostName` property](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.client.iothubconnectionstringbuilder.hostname?view=azure-dotnet#Microsoft_Azure_Devices_Client_IotHubConnectionStringBuilder_HostName)?

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -126,7 +191,18 @@ internal LocalConfigHandler()
 
                 internal override string RedisConnectionString => this.config.GetValue<string>(RedisConnectionStringKey);
 
-                internal override string IoTHubConnectionString => this.config.GetValue<string>(IoTHubConnectionStringKey);
+                internal override string IoTHubConnectionString => this.config.GetValue<string>(IoTHubConnectionStringKey, string.Empty);
```
atifaziz, line 194: What's the reason for defaulting an empty string in these new settings? Seems it makes it unclear what's optional and what belongs together as a unit of settings.

LoRaEngine/LoraKeysManagerFacade/IDevice.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
```
atifaziz, line 4: Enable nullable context in new source files:

```suggestion
#nullable enable

namespace LoraKeysManagerFacade
```

LoRaEngine/LoraKeysManagerFacade/IDeviceRegistryManager.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
```
atifaziz, line 4: Enable nullable context in new source files:

```suggestion
#nullable enable

namespace LoraKeysManagerFacade
```

LoRaEngine/LoraKeysManagerFacade/IDeviceTwin.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
```
atifaziz, line 4: Enable nullable context in new source files:

```suggestion
#nullable enable

namespace LoraKeysManagerFacade
```

LoRaEngine/LoraKeysManagerFacade/IDeviceTwin.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    using System;
+
+    public interface IDeviceTwin
```
atifaziz, line 8: I'd like to see `DevEui`, `StationEui` and `DevAddr` being used here rather than degenerating to a `string`.

Also curious to know why `DeviceId` is a property and the others modeled as methods?

LoRaEngine/LoraKeysManagerFacade/IDeviceTwin.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        string DeviceId { get; }
+
+        string GetGatewayID();
+
+        string GetNwkSKey();
+
+        string GetDevAddr();
+
+        DateTime GetLastUpdated();
```
atifaziz, line 18: Should this be in local time always? Why not just go with `DateTimeOffset` to have the least ambiguity?

LoRaEngine/LoraKeysManagerFacade/IRegistryPageResult.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
```
atifaziz, line 4: Enable nullable context in new source files:

```suggestion
#nullable enable

namespace LoraKeysManagerFacade
```

LoRaEngine/LoraKeysManagerFacade/IRegistryPageResult.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    public interface IRegistryPageResult<TResult>
+        where TResult : class
```
atifaziz, line 10: Is this really needed?

```suggestion
```

LoRaEngine/LoraKeysManagerFacade/IRegistryPageResult.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        Task<IEnumerable<TResult>> GetNextPageAsync();
```
atifaziz, line 12: Async operations should take a `CancellationToken`.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/Device.cs
atifaziz, line 6: Seal:

```suggestion
    public sealed class Device
```

Also, can we make this class immutable?

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/DeviceCollectionResult.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade.IoTCentralImp.Definitions
+{
+    using System.Collections.Generic;
+
+    public class DeviceCollectionResult
```
atifaziz, line 8: This doesn't appear to be used anywhere? If so, let's remove it.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/Device.cs
atifaziz: Ditto for all other types being introduced in this namespace.

LoRaEngine/LoraKeysManagerFacade/IRegistryPageResult.cs
atifaziz: Ditto elsewhere.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/DeviceTemplateInfo.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade.IoTCentralImp.Definitions
+{
+
+    internal class DeviceTemplateInfo
```
atifaziz, line 7: Is there a reason for this being `internal` and other types being `public` in this namespace?

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/IAttestation.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade.IoTCentralImp.Definitions
+{
+    internal interface IAttestation
```
atifaziz, line 6: Is there a reason for this being `internal` and other types being `public` in this namespace?

Also, this abstraction isn't used anywhere except on its one implementation `SymmetricKeyAttestation`. Should we remove it in case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)?

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/IAttestation.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        string Type { get; set; }
```
atifaziz, line 8: Should this be an `enum` type?

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
kbeaugrand: Hi @Ellbach, 

Thank you for your suggestion. I'll check if querying the DPS by Rest API will do the stuff. But in my mind, it seems that the DPS will not handle device properties (obviously reported properties), so we won't be able to search the device by its DevAddr...

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DesiredProperties.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class DesiredProperties
+    {
+        public string GatewayID { get; set; }
+
+        public string DevAddr { get; set; }
+
+        public string NwkSKey { get; set; }
+
+        [JsonExtensionData]
+        public Dictionary<string, JToken> AdditionalData { get; } = new Dictionary<string, JToken>();
```
atifaziz, line 19: nit: this can be shortened to:

```suggestion
        public Dictionary<string, JToken> AdditionalData { get; } = new();
```

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwin.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            return new List<DateTime>()
+            {
+                metadata?["NwkSKey"]?["lastUpdateTime"]?.ToObject<DateTime>() ?? DateTime.MinValue,
+                metadata?["DevAddr"]?["lastUpdateTime"]?.ToObject<DateTime>() ?? DateTime.MinValue,
+                metadata?["GatewayID"]?["lastUpdateTime"]?.ToObject<DateTime>() ?? DateTime.MinValue
+            }
+            .OrderByDescending(x => x)
+            .First();
```
atifaziz, line 46: The duplication of each time lookup across keys can be reduced as follow:

```suggestion
            var updateTimes =
                from x in new[] { "NwkSKey", "DevAddr", "GatewayID" }
                select metadata?[x]?["lastUpdateTime"]?.ToObject<DateTime>() ?? DateTime.MinValue into dt
                orderby dt descending
                select dt;

            return updateTimes.First();
```

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
```diff
@@ -0,0 +1,65 @@
...
+        private readonly IEnumerator<DeviceTemplateInfo> deviceTemplatesEnumerator;
+
+        public DeviceTwinPageResult(HttpClient client, IEnumerable<DeviceTemplateInfo> deviceTemplateInfos, string apiVersion, Func<DeviceTemplateInfo, string> query)
+        {
+            this.client = client;
+            this.deviceTemplateInfos = deviceTemplateInfos;
+            this.apiVersion = apiVersion;
+            this.query = query;
+
+            deviceTemplatesEnumerator = deviceTemplateInfos.GetEnumerator();
```
atifaziz, line 29: This enumerator is never disposed.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
```diff
@@ -0,0 +1,65 @@
...
+    using System.Threading.Tasks;
+    using LoraKeysManagerFacade.IoTCentralImp.Definitions;
+    using Newtonsoft.Json.Linq;
+
+    internal class DeviceTwinPageResult : IRegistryPageResult<IDeviceTwin>
+    {
+        private readonly HttpClient client;
+        private readonly string apiVersion;
+        private readonly Func<DeviceTemplateInfo, string> query;
+        private readonly IEnumerable<DeviceTemplateInfo> deviceTemplateInfos;
```
atifaziz, line 19: `deviceTemplateInfos` never seems to be used.

```suggestion
```

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
```diff
@@ -0,0 +1,65 @@
...
+            var results = new List<IDeviceTwin>();
+
+            if (!this.HasMoreResults)
+            {
+                return results;
+            }
+
+            HttpResponseMessage result = null;
+
+            result = await this.client.PostAsJsonAsync(new Uri(this.client.BaseAddress, $"/api/query?{this.apiVersion}"), new
```
atifaziz, line 46: Not a fan of embedding strings in URLs without proper encoding or escaping. Can we improve this?

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
```diff
@@ -0,0 +1,65 @@
...
+            HasMoreResults = deviceTemplatesEnumerator.MoveNext();
+        }
+
+        public bool HasMoreResults { get; private set; }
+
+        public async Task<IEnumerable<IDeviceTwin>> GetNextPageAsync()
+        {
+            var results = new List<IDeviceTwin>();
+
+            if (!this.HasMoreResults)
```
atifaziz, line 39: Property access doesn't need a `this` qualifier:

```suggestion
            if (!HasMoreResults)
```

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                return results;
+            }
+
+            HttpResponseMessage result = null;
+
+            result = await this.client.PostAsJsonAsync(new Uri(this.client.BaseAddress, $"/api/query?{this.apiVersion}"), new
+            {
+                query = this.query(deviceTemplatesEnumerator.Current)
+            });
```
atifaziz, line 49: Why not combine to?

```suggestion
            var result = await this.client.PostAsJsonAsync(new Uri(this.client.BaseAddress, $"/api/query?{this.apiVersion}"), new
            {
                query = this.query(deviceTemplatesEnumerator.Current)
            });
```

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace LoraKeysManagerFacade.IoTCentralImp
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+    using LoraKeysManagerFacade.IoTCentralImp.Definitions;
+    using Newtonsoft.Json.Linq;
```
atifaziz, line 12: Can we avoid use of Json.NET in new code and replace with System.Text.Json?

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/IoTCentralDevice.cs
```diff
@@ -0,0 +1,26 @@
...
+            this.attestationObject = attestationObject;
+            this.AssignedIoTHub = assignedIoTHubHostname;
+        }
+
+        public string PrimaryKey => this.attestationObject.SymmetricKey.PrimaryKey;
+
+        public string DeviceId => this.deviceObject.Id;
+
+        public string AssignedIoTHub { get; }
+    }
```
atifaziz, line 25: Avoid keeping whole objects alive in fields if we only use/need some of their properties:

```suggestion
    internal sealed class IoTCentralDevice : IDevice
    {
        public IoTCentralDevice(Device deviceObject, SymmetricKeyAttestation attestationObject, string assignedIoTHubHostname)
        {
            DeviceId = deviceObject.Id;
            PrimaryKey = attestationObject.SymmetricKey.PrimaryKey;
            AssignedIoTHub = assignedIoTHubHostname;
        }

        public string PrimaryKey { get; }
        public string DeviceId { get; }
        public string AssignedIoTHub { get; }
    }
```

I would even go as far as sending just the needed values in the constructor.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/IoTCentralDeviceRegistryManager.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        private readonly IDeviceProvisioningHelper provisioningHelper;
+        private readonly HttpClient client;
```
atifaziz, line 20: This client is not disposed.

LoRaEngine/LoraKeysManagerFacade/IRegistryPageResult.cs
atifaziz: Also consider returning a `Task<ICollection<TResult>>` to require the sequence of results to be materialised.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
Ellerbach: Correct, DPS is for the association with the Hub. And looking again at the code, if the device is provisioned "just" with DPS, the code will be similar (once allocated) to what it is with an associated IoT Hub.
Reported properties will show up in the twins, they are just twins. Specific twins but twins.

LoRaEngine/LoraKeysManagerFacade/IDevice.cs
Ellerbach: @Mandur you should think of enabling for the all repository. I wanted to make the comment as well but was not courageous enough to propose just for a file ;-) You'll have "some" work to enable it everywhere but quite some benefits as well. It's a great way to catch non checked nulls thru the static code analyse made by Roslyn.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
Ellerbach: the device will report the twin. Here is an example with a device associated to a modelId, for P&P, so a typical IoT Central device. It does report the properties properly in the twins:
image

The writable properties are a bit different but you can access them thru PropertyName.value instead of just PropertyName .

Ellerbach: So most likely, you can simplify quite a lot the way you query the device properties and reuse the non P&P way by adjusting it. That's the only thing specific to P&P and to IoT Central.
DPS will give you the associated hub to query

kbeaugrand: Hi @Ellerbach,

I agree, bu we cannot wuery the IoT Hub for two reasons : 

- We may hav multiple IoT Hub Instances involded behind the IoT Central
- We don't have any key to connect to the IoT Hub and query it...

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
kbeaugrand: I think we could do another PR dedicated to this topic. 

There is also a question around this: we use these urls to configure the deployment manifest, but we already have this information in the current request (hostname, code) since we are called by this information. 
This would simplify the configuration and the code. 

What do you think about it?

LoRaEngine/LoraKeysManagerFacade/DeviceFeedConstants.cs
kbeaugrand: Hi @Atif, 

Thank you for this suggestion. 
I didn't know this annotation. could you explain me what does it provide and when its interresting to use ?

LoRaEngine/LoraKeysManagerFacade/IDeviceTwin.cs
kbeaugrand: Hi, 

Since original values are strings. How should I proceed to convert it to DevEui,... ?

For the second point I only kept the code as existing (since it was extensions methods). I'll change that to properties and accessors.

LoRaEngine/LoraKeysManagerFacade/IDeviceTwin.cs
kbeaugrand: Before this PR, it used Local DateTime ...

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/DeviceCollectionResult.cs
kbeaugrand: it's used by IEnumerable<Device>.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/DeviceTemplateInfo.cs
kbeaugrand: fixed

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/IAttestation.cs
kbeaugrand: Interface removed

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/Definitions/IAttestation.cs
kbeaugrand: This interface was removed

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/IoTCentralDeviceRegistryManager.cs
kbeaugrand: The client is managed by DI.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
Ellerbach: > * We may hav multiple IoT Hub Instances involded behind the IoT Central

That would be ok as you'll get it thru the DPS instance

> * We don't have any key to connect to the IoT Hub and query it...

But that's the real bummer, you're right. Is there a way to pull the keys and have that done somehow in a more transparent way? 

what bother me with your implementation is that it's quite complex while al the elements behind are almost the same as in the original implementation, just that they are split over different hubs and properties are reported a bit differently.

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
Ellerbach: in this case, the first / is not needed and should be removed.

LoRaEngine/LoraKeysManagerFacade/DeviceFeedConstants.cs
Ellerbach: They're anyway transformed as string later, so what will be the gain to have them in something else that string?

LoRaEngine/LoraKeysManagerFacade/DeviceRegistryMode.cs
Ellerbach: So if there is no better way to have the IoT Central API, then I'm fine with IoT Central here.

LoRaEngine/LoraKeysManagerFacade/IoTCentralImp/DeviceTwinPageResult.cs
kbeaugrand: > But that's the real bummer, you're right. Is there a way to pull the keys and have that done somehow in a more transparent way ?

I didn't find any way to get the keys.. ;(

> what bother me with your implementation is that it's quite complex while al the elements behind are almost the same as in the original implementation, just that they are split over different hubs and properties are reported a bit differently.

I completly agree with you. To be honest, when starting to connect my devices with IoT Central I guest that it was as simple as creating a switch, adding a configuration and handle the IoT Central with some APIs to get informations about IoT Hub... 
But sincerely, while it's not ideal, since this is a SaaS/AaaS solution, it's quite normal to get only access to the high level.

Ellerbach: I looked at some of the docs and some of the options. It seems that the operations lists https://docs.microsoft.com/en-us/rest/api/iot-dps/operations/list and then some of the operations allows to get the keys directly of the concerned IoT Hubs. Especially this one: Microsoft.Devices/iotHubs/listkeys/Action
So if that's the case, it can drastically simplify quite some elements.

kbeaugrand: Hi @Ellerbach, 

That's true when working directly with DPS and IoT Hub. 
However, In case of IoT Central, as for IoT Hub, we haven't the keys to the DPS anyway ...

kbeaugrand:
Hi, 

@Mandur , @atifaziz , @Ellerbach , I rebased this branche to get all working without any merge issue. 
We are still facing the difficulty arround the IoT Central API taht doesn't give us a way to queyr multiple templates at a time. 

Do you think that we could merge it anyway, since it seems that it doesn't break the IoT Hub configuration and work together to continue what is missing for IoT Central with future PRs ?

kbeaugrand:
Hi As discussed, 

I close this PR, I'll provide some new PRs by pieces to provide this feature...

Issue #732, opened 2021-11-08
[Story] DwellTime support for AS923 ADR


MaggieSalak:
## Description

DwellTime is a parameter specific to AS923 region which can be either 0 or 1. We need to dive into this and find out if it needs to be set to 1 or not and calculate dependent values accordingly (e.g. max payload sizes are different in case it's set to 1).

roel4ez:
@mandur can we pick this up? Is this the only story remaining to close out AS923 support?

Issue #413, opened 2021-09-28
[Story] Support Basic frequency plan (DR, Freq, TxPow, RX1DROffset) for AS923


roel4ez:
- [x] ADR
- [ ] Implement region AS923
- [ ] Support for different offset values - pass necessary data in constructor and calculate the offset in the AS923 region setup
- [ ] Unit tests

PR #733, opened 2021-11-08
Feature/413 Support for AS923 frequency plan

MaggieSalak:
# PR for issue #413 

## What is being addressed

Support for region AS923 is introduced.

## How is this addressed

- New region `RegionAS923` was added, with support for basic frequency plan, RX1 and RX2 receive windows, TXPowertoMaxEIRP, RX1 DR offsets.
- Unit test data added in class `RegionAS923TestData` to cover the new region.
- Unit tests for overloads using Rxpk added in class `RegionAS923TestWithRxpk` to cover the obsolete Rxpk logic (which is still in use) - can be deleted once new LBS implementation is completed.
- Minor refactoring of existing region methods, mostly renaming and reordering of the existing methods.

## Notes

- Support for `DwellTime` parameter is tracked in a separate story #732. The current implementation allows passing of the parameter via the region AS923 constructor (with default value 0) and configures the region depending on the parameter value, but the parameter is currently never set to 1.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionAS923.cs
```diff
@@ -0,0 +1,143 @@
...
+
+    public class RegionAS923 : Region
+    {
+        private readonly Hertz frequencyOffset;
+        private readonly bool useDwellTimeLimit;
+
+        public RegionAS923(Hertz frequencyChannel0, Hertz frequencyChannel1, int dwellTime = 0)
+            : base(LoRaRegionType.AS923)
+        {
+            frequencyOffset = new Hertz(frequencyChannel0.AsUInt64 - 923_200_000);
```
Mandur, line 20: nit; maybe 923_200_000 and 923_400_000 could be in a const

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionAS923.cs
```diff
@@ -0,0 +1,143 @@
...
+            if (frequencyChannel1.AsUInt64 - 923_400_000 != frequencyOffset.AsUInt64)
+            {
+                throw new ArgumentException($"Provided channel frequencies {frequencyChannel0}, {frequencyChannel1} for Region {LoRaRegion} are inconsistent.");
```
Mandur, line 23: I tthink you should rather throw a ConfigurationErrorsException here

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionAS923.cs
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            useDwellTimeLimit = dwellTime == 1;
```
Mandur, line 26: if I undertand correctly the value of dwelltime can be either 1 or 0, if the value is above we should throw a ConfigurationErrorsException

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs
```diff
@@ -124,17 +123,28 @@ public static Region CN470
...
+        private static Region as923;
+
+        public static Region AS923
+        {
+            get
+            {
+                if (as923 == null)
+                {
+
+                    // Temporarily using these default parameter values until we can pass the configured frequencies to the region
```
Mandur, line 142: IS this coming in the next PR? 😀

Mandur (approved):
Very cool

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionManager.cs
MaggieSalak: I discussed with @danigian and we added it in this PR in the `LnsStationConfiguration` as part of 7f02093d2ac949886017a1bc4d0172b05fbffd79. The default values have been removed from `RegionManager`.

PR #734, opened 2021-11-08
LNS - Loading server certificate from file

danigian:
# PR for issue #639 

## What is being addressed

This PR is introducing two NetworkStationConfiguration variables for allowing to import a .pfx to be used as server-side certificate.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServer.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServer.cs
@@ -25,22 +25,15 @@ namespace LoRaWan.NetworkServer.BasicsStation
         {
             if (configuration is null) throw new ArgumentNullException(nameof(configuration));
 
-            var endpoints = new List<string>()
-            {
-                FormattableString.Invariant($"http://0.0.0.0:{Port}")
-            };
-
-            if (!string.IsNullOrEmpty(configuration.LnsServerPfxPath))
-            {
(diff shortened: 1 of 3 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServer.cs
```diff
@@ -19,18 +21,42 @@ public static class BasicsStationNetworkServer
...
+
+            var endpoints = new List<string>()
+            {
+                FormattableString.Invariant($"http://0.0.0.0:{Port}")
+            };
+
+            if (!string.IsNullOrEmpty(configuration.LnsServerPfxPath))
+            {
+                endpoints.Add(FormattableString.Invariant($"https://0.0.0.0:{SecurePort}"));
+            }
```
Mandur, line 36: If I understand the current logic we always refer to the http standard endpoint even if we have a https configure, woudln't that make sense to only enable it if we don't have a certificate?

```suggestion
            var endpoints = new List<string>();
            if (!string.IsNullOrEmpty(configuration.LnsServerPfxPath))
            {
                endpoints.Add(FormattableString.Invariant($"https://0.0.0.0:{SecurePort}"));
            }
            else
             {
                FormattableString.Invariant($"http://0.0.0.0:{Port}")
            };
```

danigian: Thanks for the suggestion. Applied in slightly different form

Mandur (approved):

PR #736, opened 2021-11-08
Introducing TC_TRUST_PATH in BasicsStation Docker Image

danigian:
# PR for issue #735 

## What is being addressed

This PR is introducing the TC_TRUST_PATH environment variable for startup script of BasicsStation docker image.

When properly setting the variable a file including the trust root certificate will be copied to tc.trust

Usage example:
- Copy the "ca.crt" on your device in "/home/pi/basicstation"
- Add a binding in HostConfiguration for mapping the above mentioned folder to a "/mappedDrive"
        "Binds": [
            "/home/pi/basicstation/:/mappedDrive/"
        ]
- Set the TC_TRUST_PATH to be "/mappedDrive/ca.crt"

When the BasicsStation container will start, the provided certificate will be copied over and used as tc.trust (as per official BasicsStation documentation)

Warning: in case of mismatch between certificate CN and the hostname configured in tc.uri, it's possible to set the "TLS_SNI" environment variable to false, in order to skip the check for SNI (defaulted to true)


LoRaEngine/modules/LoRaBasicsStationModule/start_basestation_amd64.sh
```diff
@@ -15,6 +15,12 @@ if [[ -z "$SPI_DEV" ]] || [[ $SPI_DEV == '$LBS_SPI_DEV' ]]; then
     SPI_DEV = 0
 fi
 
+if [[ -z "$TC_TRUST_PATH" ]]; then
```
Mandur, line 18: Couldn't we default to check if a cert is certs on a path (e,.g. /etc/certs/lora) and check if we have a cert there?

Mandur (approved):

LoRaEngine/modules/LoRaBasicsStationModule/start_basestation_amd64.sh
danigian: Added a DEFAULT_TC_TRUST_PATH to be "/var/lorastarterkit/certs/tc.trust" (chose /var, as I was thinking about compatibility with EFLOW)

PR #743, opened 2021-11-08
Fix CA error occurring in VS2022

bastbu:
## What is being addressed

This PR fixes a CA error that is raised when building the solution with the VS2022 current release.


atifaziz (changes requested):
Seems this does not need a global suppression at the module scope so let's move it to `Program.cs` as a regular directive, as follows:

```c#
// Similar to bug https://github.com/dotnet/roslyn-analyzers/issues/5184
#pragma warning disable CA1812
```

atifaziz (approved):

PR #746, opened 2021-11-08
Fix dockerfiles

Mandur:
Dockerfiles are wrong and break the build


Mandur:
Successfull E2E run @danigian https://github.com/Azure/iotedge-lorawan-starterkit/runs/4149589686?check_suite_focus=true

danigian (approved):

Issue #657, opened 2021-10-28
If the Datarate offset is invalid we reply on the upstreamDR


Mandur:
In the [current logic](https://github.com/Azure/iotedge-lorawan-starterkit/blob/bcd6a2b6782ee889c3d36997941b88b8eccc0d64/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs#L268) we reply on the same upstream datarate if the datarate offset is not valid. Such a case should never occur, and an error should happen.
If we want anyway to send back something this should be region dependent.

PR #750, opened 2021-11-09
return null and report error in case RX1 DR offset is invalid

MaggieSalak:
# PR for issue #657 

## What is being addressed

In case provided RX1 data rate offset was out of range (greater than the size of the RX1 DR offset table), we were responding using the same data rate as the uplink. Instead this should be treated as an error.

## How is this addressed

If the offset is invalid, we log an error and return null. All usages of `GetDownstreamDataRate()` have been updated to handle this situation. Unit tests were updated accordingly.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -152,15 +152,16 @@ public string GetDownstreamDataRate(Rxpk upstreamChannel, int rx1DrOffset = 0)
...
                 {
                     return DRtoConfiguration[(ushort)RX1DROffsetTable[GetDRFromFreqAndChan(upstreamChannel.Datr)][rx1DrOffset]].configuration;
                 }
                 else
                 {
-                    Logger.Log($"rx1 Dr Offset was not set to a valid value {rx1DrOffset}, defaulting to {upstreamChannel.Datr} datarate", LogLevel.Error);
-                    return upstreamChannel.Datr;
+                    Logger.Log($"RX1 data rate offset was set to an invalid value {rx1DrOffset}; " +
+                        $"maximum allowed offset is {RX1DROffsetTable[0].Count - 1}", LogLevel.Error);
+                    return null;
```
Mandur, line 164: I think we could throw a LoRa ProcessingException https://github.com/Azure/iotedge-lorawan-starterkit/pull/713 @bastbu

MaggieSalak: Sounds good. What do you think about the case where `IsValidUpstreamRxpk(upstreamChannel)` is false (where we currently return null)?

Mandur: I think this should be changed to void, if IsValidUpstreamRxpk(upstreamChannel) is false we should also throw

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -177,15 +177,15 @@ public string GetDownstreamDataRate(Rxpk upstreamChannel, int rx1DrOffset = 0)
...
                 if (rx1DrOffset <= RX1DROffsetTable[0].Count - 1)
                 {
                     return (ushort)RX1DROffsetTable[dataRate][rx1DrOffset];
                 }
                 else
                 {
-                    Logger.Log($"rx1 Dr Offset was not set to a valid value {rx1DrOffset}, defaulting to {dataRate} datarate", LogLevel.Error);
-                    return dataRate;
+                    throw new LoRaProcessingException($"RX1 data rate offset was set to an invalid value {rx1DrOffset}; " +
+                           $"maximum allowed offset is {RX1DROffsetTable[0].Count - 1}", LoRaProcessingErrorCode.InvalidDeviceConfiguration);
```
Mandur, line 188: I would suggest here to add another errorcode, like "InvalidUpstreamDataRate"

Mandur (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
MaggieSalak: I agree, changed the method to return the correct data rate or throw `LoRaProcessingException`.

Issue #697, opened 2021-11-03
LBS device should be equivalent to LoRa device in IoT Hub


bastbu:
## Expected Behavior

The LBS devices in IoT Hub should have names equivalent to LoRa devices, which means that LBS device names should not need to have dashes.

## Current Behavior

Currently LBS device names need to include dashes, e.g. `DC-A6-32-FF-FE-B3-2F-C6`.

roel4ez:
This needs a bigger discussion. Let's agree on this before starting @danigian @bastbu

PR #757, opened 2021-11-09
Bug/697 lbs device no hyphens

bastbu:
# PR for issue #697

## What is being addressed

This PR makes sure that when fetching the station device twin, we use the same device ID format as we do for LoRa devices.

## How is this addressed

Uses the primitives and a specific formatting to fetch the device by ID.

Depends on #752

### DIFF
--- a/Docs/Operations/station-configuration.md
+++ b/Docs/Operations/station-configuration.md
@@ -4,17 +4,17 @@ Following the LoRaWAN  Network Server (LNS) specification, each Basics Station (
 
 In the following we describe how to register an LBS in IoT Hub and how to store its configuration.
 
-1. Create an IoT Hub device that has a name equal to the LBS EUI in hex-representation, e.g. `DC-A6-32-FF-FE-B3-2F-C6`.
+1. Create an IoT Hub device that has a name equal to the LBS EUI in hex-representation, e.g. `DCA632FFFEB32FC6`.
 2. The LBS configuration needs to be stored as a desired twin property of the newly created LBS device. Make sure to store the configuration under `properties.desired.routerConfig`.
    1. The configuration follows the `router_config` format from the LNS protocol as closely as possible. However, since device twins encode numbers as 32-bit values and given some configuration properties (such as EUIs) are 64-bit numbers, there are some minor differences.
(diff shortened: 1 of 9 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
```diff
@@ -174,7 +181,7 @@ public override async Task<SearchDevicesResult> SearchByDevEUIAsync(string devEU
                     return new SearchDevicesResult();
                 }
 
-                Logger.Log(devEUI, $"error calling get device by devEUI api: {response.ReasonPhrase}, status: {response.StatusCode}, check the azure function log", LogLevel.Error);
+                Logger.Log(eui, $"error calling get device by devEUI api: {response.ReasonPhrase}, status: {response.StatusCode}, check the azure function log", LogLevel.Error);
```
spygi, line 184:
```suggestion
                Logger.Log(eui, $"error calling get device/station by EUI api: {response.ReasonPhrase}, status: {response.StatusCode}, check the azure function log", LogLevel.Error);
```

spygi (approved):
LGTM!

Docs/Operations/station-configuration.md
```diff
@@ -4,17 +4,17 @@ Following the LoRaWAN  Network Server (LNS) specification, each Basics Station (
 
 In the following we describe how to register an LBS in IoT Hub and how to store its configuration.
 
-1. Create an IoT Hub device that has a name equal to the LBS EUI in hex-representation, e.g. `DC-A6-32-FF-FE-B3-2F-C6`.
+1. Create an IoT Hub device that has a name equal to the LBS EUI in hex-representation, e.g. `DCA632FFFEB32FC6`.
```
Mandur, line 7:
```suggestion
1. Create an IoT Hub device that has a name equal to the LBS EUI in hex-representation, e.g. `DCA632FFFEB32FC6` (case sensitive).
```

Mandur: Not sure it's needed but I discovered it yesterday ;)

bastbu: It shouldn't be case sensitive any longer. Can you still reproduce with the latest version of dev?

PR #758, opened 2021-11-09
Checks previous station connectivity

spygi:
# PR for issue #724

## What is being addressed

An add-on on #674 , here we check further if the previous station is still "reachable" at the time of deduplication. 

## How is this addressed

A method is added on the socket registry to check for connectivity. The check is based solely on what the underlying socket reports as it's status (see `WebSocketTextChannel.IsClosed`) i.e. we don't actually try to send something to the socket. Note that this check is still best-effort as there is no guarantee on what will happen to the socket when we actually try to reach the LBS downstream.

In order to keep methods small and doing a single thing, I opted against extending the `WebSocketWriterRegistry.TryGetHandle` to return both the handle and the socketwriter.

Unit tests covering this case were added. At the moment the service is still not integrated with `LnsProtocolMessageProcessor`: this will happen in the next PR together with integration tests.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
```diff
@@ -84,6 +84,16 @@ public bool TryGetHandle(TKey key, [NotNullWhen(true)] out IWebSocketWriterHandl
...
 
+        /// <summary>
+        /// Checks whether the socket writer associated with the given key
+        /// is flagged as open. Actual connectivity is not checked.
+        /// </summary>
+        /// <remarks>
+        /// Implementation does not throw when key is not present in the underlying dictionary.
+        /// </remarks>
+        public bool IsSocketWriterOpen(TKey key)
+            => TryGetWriter(key, out var socketWriter) && !socketWriter.IsClosed;
```
spygi, line 95: if the socket is closed we could also deregister it here, thoughts?

Tests/Unit/NetworkServer/BasicsStation/DownstreamSenderTests.cs
```diff
@@ -7,7 +7,6 @@ namespace LoRaWan.Tests.Unit.NetworkServer.BasicsStation
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
-    using LoRaTools.LoRaPhysical;
```
spygi, line 10: unrelated change, import not used.

danigian:
As discussed with Atif, I will hand-over the review of this PR to him.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
atifaziz: The `WebSocketWriterRegistry` technically doesn't care about that. That's the job of the user of this type to call `Deregister` when the socket is closed (or if `Prune` is being called on a regular basis, which we don't at the moment). The one user of this _does do that_:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/b50bf4fcb63e4ca4c0067a13ecf9a68ce67b9870/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs#L163

I do feel this method should therefore be renamed and implemented such that it distinguishes between missing an entry, closed and open:

```c#
public bool? IsClosed(TKey key)
{
    lock (this.sockets)
        return this.sockets.TryGetValue(key, out var writer) ? writer.Object.IsClosed : null;
}
```

where the return value would be:

- `null` &rArr; unknown/unregistered
- `true` &rArr; closed
- `false` &rArr; open

spygi: My comment might have been confusing, I didn't mean the `Deregister()` method here (as I understand this is for users of the registry). I rather meant `Prune()`. But we don't have to do that and eventually the closed sockets will be pruned.

About returning null here if the socket doesn't exist, do you have a use-case where that distinguish would be helpful in the current story/context? Because if not, I wouldn't do that yet.

spygi: Let me know if we should hop on a call for this.

spygi: Discussed in a call with Atif and our preferred way would be to create a private method that returns both objects and re-used from 2 places.

atifaziz: As agreed on the call, let's either: 

1. Consolidate the duplication between `TryGetHandle` and `TryGetWriter`
2. Use `TryGetHandle`, make `IsSocketWriterOpen` private and expose `IsOpen` on `IWebSocketWriterHandle<>`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -74,6 +87,7 @@ private void AddToCache(string key, StationEui stationEui)
                 SlidingExpiration = DefaultExpiration
             });
 
-        public void Dispose() => this.cache.Dispose();
+        public void Dispose()
+            => this.cache.Dispose();
```
atifaziz, line 91: nit: Revert as this change seems unrelated?

```suggestion
        public void Dispose() => this.cache.Dispose();
```

atifaziz (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
spygi: Done in e1bc1e1c9ba961a124531e8297a7476025a07a51

Issue #610, opened 2021-10-21
[bug] Namespaces of tests should follow better conventions


spygi:
Reported by @atifaziz :

## Current Behavior
Tests are currently organised as follows:
- Common
- Unit
  - Facade
  - NetworkServer
  - LoRaWan 
- Integration
- E2E

## Expected Behavior
For the class `LoRaDevice` in `LoRaWan.NetworkServer` namespace, most expect tests to be found in (FQN) `LoRaWan.Tests.Unit.NetworkServer.LoRaDeviceTests`

Another suggestion is to drop "tests" from the subfolder structure and namespace: 
> I find repeating Tests twice in the namespace and then again in the class name doesn't add much value, e.g.:
LoRaWan.Tests.Unit.NetworkServer**Tests**.LoRaDeviceTest -> LoRaWan.Tests.Unit.NetworkServer.LoRaDeviceTest**s**

that's usually more conventional.

PR #759, opened 2021-11-09
Bug/610 fix test namespaces

spygi:
# PR for issue #610

## What is being addressed

Adjusting namespaces and folder structure to what is more common in the .NET community.

## How is this addressed

- Namespaces of test classes match classes under test prefixed with `LoRaWan.Tests.Unit`
- Folders are used only to organize tests in buckets therefore they comply with the namespace but loosely i.e. we don't nest fully to avoid having too many folders with very few test classes in them. Folders anyway don't influence test execution or project boundaries. 
- The suffix "tests" was removed from the folder names.

Some examples:
- `LoRaTools.CommonAPI.ApiVersion` is tested in `LoRaWan.Tests.Unit.LoRaTools.CommonAPI.ApiVersionTest` in folder  `Tests/Unit/LoRaTools/` 
- `LoraKeysManagerFacade.DeviceGetter` is tested in `LoRaWan.Tests.Unit.LoraKeysManagerFacade.DeviceGetterTest` in folder `Tests/Unit/LoraKeysManagerFacade/`


Tests/Unit/NetworkServer/ChangeTrackingPropertyTest.cs
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace LoRaWan.Tests.Unit.NetworkServerTests
+namespace LoRaWan.Tests.Unit.NetworkServer
 {
-    using LoRaTools.Regions;
+    using global::LoRaTools.Regions;
     using LoRaWan.NetworkServer;
     using Xunit;
```
bastbu, line 9: comment for a different file: `ConcentratorDeduplicationTest` should also have a namespace without `Test`.

Tests/Unit/LoRaTools/LoRaCloudToDeviceMessageTest.cs
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace LoRaWan.Tests.Unit.LoRaWanTests
+namespace LoRaWan.Tests.LoRaTools.CommonAPI
```
bastbu, line 4: The namespace here doesn't match the convention to always start with `LoRaWan.Tests.Unit`.

Tests/Unit/CollectionExtensionsTests.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace LoRaWan.Tests.Unit.LoRaWanTests
+namespace LoRaTools.Tests.Unit.Utils
```
bastbu, line 4: [nit] Why have a separate namespace just for this class? Should we consider to parent it under `LoRaTools.Tests.Unit`? That would make it easier to navigate in the test explorer

Tests/Unit/LoRaWan/LoRaLibraryTest.cs
```diff
@@ -3,16 +3,16 @@
 
 #pragma warning disable CA1303 // Do not pass literals as localized parameters
 
-namespace LoRaWan.Tests.Unit.LoRaWanTests
+namespace LoRaWan.Tests.Unit
```
bastbu, line 6: A general observation about the tests under `LoRaWan`: why remove the entire `LoRaWanTests` here instead of just the `Tests` suffix? Would it not help discoverability and be a less invasive change if we were having them under a namespace `LoRaWan.Tests.Unit.LoRaWan`?

Tests/Unit/LoraKeysManagerFacade/PreferredGatewayTest.cs
```diff
@@ -91,7 +91,6 @@ public async Task When_Calling_After_Delay_Should_Return_First_Gateway()
 
             var req2 = new FunctionBundlerRequest() { GatewayId = "gateway2", ClientFCntUp = staleFcntUp, Rssi = -90 };
             var pipeline2 = new FunctionBundlerPipelineExecuter(new IFunctionBundlerExecutionItem[] { this.preferredGatewayExecutionItem }, devEUI, req2);
-            var res2 = await this.preferredGatewayExecutionItem.ExecuteAsync(pipeline2);
```
bastbu, line 94: Not so sure if this is a safe change - should we consider discarding instead? The operation itself might have a side effect

Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
```diff
@@ -15,8 +15,7 @@ namespace LoRaWan.Tests.Unit.NetworkServerTests.JsonHandlers
     using Xunit;
     using static Bandwidth;
     using static SpreadingFactor;
-    using static NetworkServer.BasicsStation.RouterConfigStationFlags;
```
bastbu, line 18: why remove the static imports here? does it not work with `global::` imports?

Tests/Unit/NetworkServer/LoRaCloudToDeviceMessageWrapperTest.cs
```diff
@@ -1,11 +1,12 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace LoRaWan.Tests.Unit.NetworkServerTests
+namespace LoRaWan.Tests.Unit.NetworkServer
 {
     using System;
     using System.Linq;
     using System.Text;
+    using global::LoRaTools;
     using LoRaTools;
```
bastbu, line 10: [nit] unused import

bastbu (approved):

Tests/Unit/LoRaWan/LoRaLibraryTest.cs
spygi: I thought `LoRaWan.Tests.Unit.LoRaWan` doesn't really make sense because it has 2 times `LoRaWan` in there (complies with the requirement to not have repetition there). It is arguable though what's the best approach here, I could also see yours.

Tests/Unit/NetworkServer/LoRaCloudToDeviceMessageWrapperTest.cs
spygi: Removed LoRaTools with 8fd7eb42de7b8974bc25bc5426b455fa768baf28

Tests/Unit/LoraKeysManagerFacade/PreferredGatewayTest.cs
spygi: Ignored with 8fd7eb42de7b8974bc25bc5426b455fa768baf28

Tests/Unit/CollectionExtensionsTests.cs
spygi: Actually `CollectionExtensions` is part of `LoRaTools.Utils`, so the test should be in namespace `LoRaWan.Tests.Unit.LoRaTools.Utils` and located in the `Tests/Unit/LoRaTools/` folder, changed accordingly.

Tests/Unit/LoRaTools/LoRaCloudToDeviceMessageTest.cs
spygi: Fixed in 8fd7eb42de7b8974bc25bc5426b455fa768baf28

Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
spygi: No, I still get a warning with `global::`

Tests/Unit/NetworkServer/ChangeTrackingPropertyTest.cs
spygi: Fixed in c46892ab76eb7832c0648cca09cd6148ba74f00b

Issue #751, opened 2021-11-09
Reduce TXPower for CI


Mandur:
We need to be able to set the TXPower for:

- [x] Leaf device from the CI
- [x] Update documentation on local E2E setup

PR #785, opened 2021-11-11
Feature/751 Allow setting custom TxPower for end devices in E2E tests

MaggieSalak:
# PR for issue #751 

## What is being addressed

Currently TXPower value in used by end devices in E2E tests is hardcoded to 14. We need to be able to provide custom TXPower value in the CI and when running the tests locally.

## How is this addressed

- new environment variable added in the E2E CI for the TXPower value
- new configuration setting for tests added to store the TXPower value
- E2E tests updated to use the TXPower value from the configuration instead of a hardcoded value

The change was verified locally and in the CI by running the E2E tests and making sure (from the logs) that TXPower is set to the requested value, e.g. for TxPower = 2:
09:12:46.413 [SERIAL] +POWER: 2


MaggieSalak:
The change was verified locally and in the CI by running the E2E tests and making sure (from the logs) that TXPower is set to the requested value, e.g.
`09:12:46.413 [SERIAL] +POWER: 2`

.github/workflows/e2e-ci.yaml
```diff
@@ -11,6 +11,9 @@ on: # rebuild any PRs and main branch changes
       TestsToRun:
         default: '[\"SensorDecodingTest\",\"OTAAJoinTest\",\"ABPTest\",\"OTAATest\",\"MacTest\",\"ClassCTest\",\"C2DMessageTest\",\"MultiGatewayTest\"]'
         description: 'tests to run'
+      TxPower:
+        description: 'TXPower value to use in E2E tests'
+        default: '14'
```
Mandur, line 16: could we put the default to something lower, like 6?

Mandur (approved):

.github/workflows/e2e-ci.yaml
MaggieSalak: Done in c7aec0970a423076b6830852326ec72924e399f5

PR #788, opened 2021-11-11
Remove logging to IoT Hub that seems unused

atifaziz:
# PR that addresses story #456

## What is being addressed

Logging to IoT is unused (and potentially untested).

## How is this addressed

Removed the implementation and corresponding settings (nothing found in the documentation).

---
PS If needed, this may be brought back as part of story #740.

Issue #456, opened 2021-10-06
[Story] Migrate static Logger methods to DI


roel4ez:
## Acceptance criteria

- [ ] The static `Logger` is removed
- [ ] All logging goes via `ILogger` and DI
- [ ] Logging sinks are migrated
- [ ] Catch all exceptions block ([in `UdpLogger.InitializeUdpLogger`](https://github.com/Azure/iotedge-lorawan-starterkit/blob/d4ffe15adb68fe8904097e08207ce23133a084b4/LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs#L246)) are removed
- [ ] Docs are up to date

bastbu:
@atifaziz #740 is a possible task of this story tracked as a separate issue. Please close that issue as well if you resolve this.

Issue #740, opened 2021-11-08
[Story] Create `ILoggerProvider` for IoT Hub logger


bastbu:
## Description

We should add support to export log statements as IoT Hub messages. For that we will need to create an implementation similar to our `LoRaConsoleLogger`, which exports logs as IoT Hub messages.

## Acceptance criteria

- [ ] Logger that exports logs as IoT Hub messages
- [ ] Logger is disabled by default
- [ ] Logger can be enabled via configuration

PR #789, opened 2021-11-11
Convert UDP logging to (non-static) TCP logging

atifaziz:
This PR is for part of issue #456.

It primarily replaces UDP logging with TCP logging to have all the reliability of the latter and render tests more stable. UDP logging has the problem that data can be out of order and transmission errors are not retried. There are also a number of bugs addressed such as receiving partial data only.

The bulk of the `Logger` implementation has been simplified and moved out into separate non-static classes, namely: `ILogSink`, `CompositeLogSink`, `ConsoleLogSink`, `TcpLogSink`.

`ILogSink` is a simple new abstraction:

```c#
public interface ILogSink
{
    LogLevel LogLevel { get; }
    void Log(LogLevel logLevel, string message);
}
```

The rest of the classes are an implementation of this. Having implementations removes the if-else branches that present in the original `Logger` code. Based on configuration, `Logger` now instantiates either `ConsoleLogSink` or `TcpLogSink` or both via a `CompositeLogSink`. Eventually, these will be replaced by standard .NET logging infrastructure in a future PR, which should be straightforward since none of them has static members for logging.

In the test project, the `UdpLogListener` has been replaced by `TcpLogListener`.

The `TcpLogSink` starts a connection when the first message is logged. It also maintains a queue of logged messages that's read in a loop and where each message is delivered to the log listener. If an error occurs, the connection is closed and a new one is established. Messages accumulated during the re-connection time and attempts are then delivered. If too many messages accumulate during this time, then the oldest ones are automatically dropped to make room for new ones. The default queue capacity is 1,000 messages, connection retries is 6 attempts with 10 seconds of wait between each attempt. After the retries are exhausted, the message is dropped.

The `TcpLogSink` also re-normalises all line endings in the logged messages to use `CR` + `LF`, independent of the platform.

---
~~PS This PR includes #788 that's pending merge.~~


bastbu:
Can we run a full CI for this PR?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
```diff
@@ -0,0 +1,295 @@
...
+                                var buffer = buffers.Rent(size);
+                                var bi = encoding.GetBytes(line, buffer);
+                                buffer[bi++] = 13; // CR
+                                buffer[bi++] = 10; // LF
+                                Debug.Assert(bi == size);
+
+                                try
+                                {
+                                    await client.GetStream().WriteAsync(buffer.AsMemory(..bi), cancellationToken);
+                                    break; // don't retry on success
```
bastbu, line 234: I think there's a bug in this `break` statement. With that we break out of the inner loop, but not of the outer loop. As a result we will send the log statement many times

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
```diff
@@ -0,0 +1,295 @@
...
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Primitives;
+
+    public interface ILogSink
+    {
+        public LogLevel LogLevel { get; }
+        public void Log(LogLevel logLevel, string message);
+    }
+
+    public class LogSink : ILogSink
```
bastbu, line 29: What's the idea behind this `LogSink`? As far as I can see, it's only used once.

Tests/Common/SimpleTcpListener.cs
```diff
@@ -0,0 +1,69 @@
...
+            var listener = this.listener;
+            if (listener is null || listener != Interlocked.CompareExchange(ref this.listener, null, listener))
+                return;
+            listener.Stop();
+        }
+
+        private static async Task ListenAsync(TcpListener listener, Func<Context, NetworkStream, Task> processor)
+        {
+            while (true)
+                _ = OnAcceptAsync(await listener.AcceptTcpClientAsync().ConfigureAwait(false));
```
bastbu, line 56: Why do we not await this? If an error happens in there it goes unobserved.

Tests/Common/SimpleTcpListener.cs
```diff
@@ -0,0 +1,69 @@
...
+            Start(port, null, processor);
+
+        public static SimpleTcpListener Start(int port, int? backlog, Func<Context, NetworkStream, Task> processor)
+        {
+            var listener = TcpListener.Create(port);
+            if (backlog is { } someBackLog)
+                listener.Start(someBackLog);
+            else
+                listener.Start();
+            _ = ListenAsync(listener, processor);
```
bastbu, line 41: Do we want to expose this task to ensure that we can observe if any errors happen while listening?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        public static TcpLogSink Start(IPEndPoint serverEndPoint, LogLevel logLevel,
+                                       int? maxRetryAttempts = null,
+                                       TimeSpan? retryDelay = null,
+                                       int? backlogCapacity = null,
+                                       Func<string, string>? formatter = null,
+                                       ILogger<TcpLogSink>? logger = null)
+        {
+            var sink = new TcpLogSink(serverEndPoint, logLevel, maxRetryAttempts, retryDelay, backlogCapacity, formatter, logger);
+            _ = Task.Run(() => sink.SendAllLogMessagesAsync(sink.cancellationTokenSource.Token));
```
bastbu, line 153: What's the reason why sending log messages should happen only on a best effort basis? How do we observe/debug if something goes wrong with sending logs on the `TcpLogSink`?

Tests/Common/SimpleTcpListener.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace LoRaWan.Tests.Common
+{
+    using System;
+    using System.Net;
+    using System.Net.Sockets;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public sealed class SimpleTcpListener : IDisposable
```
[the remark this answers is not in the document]
atifaziz: Clarity, simplicity and separation of concerns.

Tests/Common/TcpLogListener.cs
```diff
@@ -0,0 +1,69 @@
...
+    using System.Collections.Generic;
+    using System.Net;
+    using System.Text;
+    using System.Threading.Tasks;
+
+    public sealed class TcpLogListener : IDisposable
+    {
+        private static readonly char[] NewLineChars = { '\r', '\n' };
+
+        public static TcpLogListener Start(int port, bool dontLogToConsole = false)
```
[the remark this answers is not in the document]
atifaziz: Yes. It's good practice to make the default for a Boolean value to be `false`. It eliminates guessing. Do you think `disableConsoleLogging` would be a less confusing name?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -75,27 +75,21 @@ public class NetworkServerConfiguration
...
         public string LogLevel { get; set; } = "4";
 
-        /// <summary>
-        /// Gets or sets a value indicating whether logging to IoT Hub is enabled.
-        /// Default is false.
-        /// </summary>
-        public bool LogToHub { get; set; }
-
         /// <summary>
         /// Gets or sets a value indicating whether logging to udp is enabled (used for integration tests mainly).
```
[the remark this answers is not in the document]
atifaziz: Thanks for spotting. Fixed with 82effa311a8eda054d5ee21d9402ee2780639dca.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerConfiguration.cs
```diff
@@ -21,19 +17,15 @@ public class LoggerConfiguration
         // Default: 4 (Error)
         public LogLevel LogLevel { get; set; } = LogLevel.Error;
 
-        // Gets/sets if logging to IoT Hub is enabled
-        // Default: false
-        public bool LogToHub { get; set; }
-
         // Gets/sets if logging to udp is enabled (used for integration tests mainly)
```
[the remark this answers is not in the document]
atifaziz: Thanks for spotting. Fixed with 82effa311a8eda054d5ee21d9402ee2780639dca.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/Logger.cs
```diff
@@ -6,254 +6,108 @@ namespace LoRaWan
...
+        private static ILogSink sink = new ConsoleLogSink(LogLevel.Error);
 
-        private static LoggerConfiguration configuration = new LoggerConfiguration();
-        private static volatile UdpClient udpClient;
-        private static IPEndPoint udpEndpoint;
-        private static volatile bool isInitializeUdpLoggerRunning;
-        private static Timer retryUdpLogInitializationTimer;
-
-        public static void Init(LoggerConfiguration loggerConfiguration)
+        public static void Init(LoggerConfiguration configuration,
```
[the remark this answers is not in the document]
atifaziz: Yes, there will be a follow-up. This PR doesn't address #456 in its entirety otherwise it would have been very large. This PR prepares the ground work since `Logger` is now very thin and also implements more robust logging over TCP.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
atifaziz: Very good catch. I refactored this a bit late with c61b03b0281667f2e925b43753b37634c3850b5d and inverted the loops but this is now fixed with 06f0af3e945ad25543b32a567b34da48aa63f6f9.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
atifaziz: It was used in two places but now I've switched to using for all subclasses with 39842d79de31986f47054d474d82468e8231cc7a. The idea behind it is that it's the base implementation.

Tests/Common/TcpLogListener.cs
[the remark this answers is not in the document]
bastbu: I see, thanks for explaining. Knowing that, I don't mind either.

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
```diff
@@ -103,16 +103,10 @@ public async Task AssertNetworkServerModuleLogStartsWithAsync(string logMessageS
...
-                searchResult = await SearchUdpLogs(predicate, options);
-            else
-                searchResult = await SearchIoTHubLogs(predicate, options);
-
-            return searchResult;
-        }
+        public async Task<SearchLogResult> SearchNetworkServerModuleAsync(Func<string, bool> predicate, SearchLogOptions options = null) =>
+            this.tcpLogListener != null
+                ? await SearchUdpLogs(predicate, options)
+                : await SearchIoTHubLogs(predicate, options);
```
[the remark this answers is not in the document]
atifaziz: Technically, we should have removed IoT Hub log searching in #788, but removing it and making TCP logging optional doesn't make sense at all because then none of the tests would pass. This is something I thought could be revisited more properly as part of #740 when we want to add back IoT Hub logging.

/cc @Mandur

Tests/Common/SimpleTcpListener.cs
atifaziz: Because we don't want the processing of one client connection to block the accepting of another. That said, there is a chance that `AcceptTcpClientAsync` or `OnAcceptAsync` (up to its first await) could have thrown so I've revised this for logging and resilience here with 12db0c112a6485b89f02e35f602e04323a64de4a. I've also renamed `OnAcceptAsync` to `OnProcessAsync` with d208e5748a3b84cbd9ec4ba0b8db90c6d30321a4 to the accepting and processing more distinct.

Tests/Common/SimpleTcpListener.cs
atifaziz: This was also addressed by 12db0c112a6485b89f02e35f602e04323a64de4a.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
atifaziz: `SendAllLogMessagesAsync` logs emits debug, warning and error messages when things go wrong.

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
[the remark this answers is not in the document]
p-schuler: I'm confused. We remove the IoT Hub logger here, because we say it's not used, but then we talk about bringing it back as a logging provider in #740 ?

LoRaEngine/example.env
```diff
@@ -29,8 +29,8 @@ PKT_FWD_SPI_DEV=2
 ##################################
 NET_SRV_VERSION=1.0.7
 NET_SRV_LOG_LEVEL=Error
-NET_SRV_LOGTO_UDP=false
-NET_SRV_LOG_TO_UDP_ADDRESS=127.0.0.1
+NET_SRV_LOG_TO_TCP=false
```
p-schuler, line 32: everyone needs to update their .env files. This is a breaking change for everyone using the kit today and needs to be release noted.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
```diff
@@ -0,0 +1,291 @@
...
+        public LogLevel LogLevel { get; }
+
+        public virtual void Log(LogLevel logLevel, string message)
+        {
+            if (LogLevel > logLevel)
+                return;
+            CoreLog(logLevel, message);
+        }
+
+        protected virtual void CoreLog(LogLevel logLevel, string message) { /* Nop */ }
```
p-schuler, line 42: shouldn't this be an abstract?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
```diff
@@ -0,0 +1,291 @@
...
+    using System.Net;
+    using System.Net.Sockets;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Channels;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Primitives;
+
+    public interface ILogSink
```
p-schuler, line 23: We seem to add quite a bit of code for something we won't use in the future. When we move to the ILogger, this seems to be throw away code. Why bother for this intermediate step?

bastbu (approved):

LoRaEngine/example.env
atifaziz: @Mandur Are we tracking release notes somewhere for the next version? Something to add to the [upgrade page](https://azure.github.io/iotedge-lorawan-starterkit/dev/user-guide/upgrade/) for vNext?

/cc @roel4ez

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LogSink.cs
atifaziz: The idea was to allow an instance of `LogSink` to be used as a null/NOP logger, which I had originally but went away with 1cee7f002352918ba5efc70998869fb8cd25d322. I can make it abstract now.

atifaziz: Made abstract with dd3c84fc26e6103f8684a392a4315de6dd0b9588.

atifaziz:
> Can we run a full CI for this PR?

Sure, pending merge of #809.

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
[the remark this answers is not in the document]
Mandur: We remove it from the static logger, but logging to iot hub is a desired feature which is desirable, therefore we will need it as part of the ILogger. make sense?

LoRaEngine/example.env
Mandur: We do not have such feature. that is a good idea thought ;)

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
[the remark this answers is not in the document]
atifaziz: > I'm confused. We remove the IoT Hub logger here, because we say it's not used, but then we talk about bringing it back as a logging provider in #740 ?

So one of the story's acceptance criteria was “Logging sinks are migrated”. Since #740 was going to revisit the IoT Hub logging, it was decided to migrate the minimum set of sinks that were necessary for now and on which the tests depend the most.

LoRaEngine/example.env
atifaziz: Tracking in #825.

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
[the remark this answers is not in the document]
p-schuler: I think we want to keep feature parity with what we had previously. We do not know, what existing users of the kit actually use, hence we need to get IoT hub logging back in using ILogger sinks.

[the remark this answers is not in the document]
p-schuler: Since #740 is tracking this, I resolve this comment.

LoRaEngine/example.env
p-schuler: We should definitely prep release notes and keep track of changes we are making.

p-schuler (approved):

Issue #792, opened 2021-11-12
Not implemented, but supported, LNS protocol message type should not throw exceptions


danigian:
## Expected Behavior
Not implemented, but supported, LNS protocol message type should not throw exceptions since we can't control the behavior of the clients connecting to our network server.

Currently exceptions are thrown and this is an issue

PR #793, opened 2021-11-12
Bugfix/792 Not implemented, but spec supported, LNS protocol message type should not throw exceptions

danigian:
# PR for issue #792 

## What is being addressed

Not implemented, but supported, LNS protocol message type should not throw exceptions since we can't control the behavior of the clients connecting to our network server.

## How is this addressed

- [x] Adding remaining types to LnsMessageType enum
- [x] Adjusting LnsMessageTypeParser and ToBasicStationString extension
- [x] Adding case for logging warning
- [x] Adding unit test


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/LnsMessageType.cs
```diff
@@ -14,28 +14,30 @@ internal enum LnsMessageType
...
-
-        */
+        RemoteShell             // rmtsh
     }
 
     internal static class LnsMessageTypeExtensions
     {
         internal static string ToBasicStationString(this LnsMessageType lnsMessageType) => lnsMessageType switch
         {
 #pragma warning disable format
```
bastbu, line 29: We should remove the suppression due to reformatting of existing code

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -232,6 +232,13 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
                     break;
                 case var messageType and (LnsMessageType.DownlinkMessage or LnsMessageType.RouterConfig):
                     throw new NotSupportedException($"'{messageType}' is not a valid message type for this endpoint and is only valid for 'downstream' messages.");
+                case var messageType and (LnsMessageType.ProprietaryDataFrame
+                                          or LnsMessageType.MulticastSchedule
+                                          or LnsMessageType.TimeSync
+                                          or LnsMessageType.RunCommand
+                                          or LnsMessageType.RemoteShell):
+                    this.logger.LogWarning($"'{messageType}' ({messageType.ToBasicStationString()}) is not handled in current LoRaWan Network Server implementation.");
```
bastbu, line 240: How often do we expect this to happen? Is it really worth logging a warning here, or is it expected and maybe a `Debug` statement would suffice to avoid flooding the logs?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -232,6 +232,13 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
...
+                    break;
```
bastbu, line 241: Do we want a unit test to cover this (the fact that we tolerate these types even though we don't handle them)?

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
danigian: Frequency should not be high. I think it's important to keep it as a warning to make sure that anyone expecting these messages to be handled, actually realizes that's not the case

Issue #798, opened 2021-11-12
`When_Takes_Too_Long_Receiving_First_C2D_Should_Abandon_Message` test fails sometimes


atifaziz:
## Expected Behavior

Expected unit tests to pass on each run.

## Current Behavior

Instead the `CloudToDeviceMessageTests.When_Takes_Too_Long_Receiving_First_C2D_Should_Abandon_Message` (integration) test fails with the following details:

```
  Failed LoRaWan.Tests.Integration.CloudToDeviceMessageTests.When_Takes_Too_Long_Receiving_First_C2D_Should_Abandon_Message [2 s]
  Error Message:
   Moq.MockException : Mock<ILoRaDeviceClient:196>:
This mock failed verification due to the following:

   ILoRaDeviceClient x => x.UpdateReportedPropertiesAsync(It.IsAny<TwinCollection>()):
   This setup was not matched.

   ILoRaDeviceClient x => x.ReceiveAsync(It.IsAny<TimeSpan>()):
   This setup was not matched.

   ILoRaDeviceClient x => x.AbandonAsync(Message):
   This setup was not matched.
  Stack Trace:
     at Moq.Mock.Verify(Func`2 predicate, HashSet`1 verifiedMocks) in C:\projects\moq4\src\Moq\Mock.cs:line 319
   at Moq.Mock.VerifyAll() in C:\projects\moq4\src\Moq\Mock.cs:line 283
   at LoRaWan.Tests.Integration.CloudToDeviceMessageTests.When_Takes_Too_Long_Receiving_First_C2D_Should_Abandon_Message() in /home/runner/work/iotedge-lorawan-starterkit/iotedge-lorawan-starterkit/Tests/Integration/CloudToDeviceMessageTests.cs:line 1001
--- End of stack trace from previous location where exception was thrown ---
```

## Steps to Reproduce

The test failure is sporadic and [occurs during some CI builds](https://github.com/Azure/iotedge-lorawan-starterkit/runs/4189305422?check_suite_focus=true#step:11:1881) and goes away with re-runs.

## Logs

- 📎 [`logs_3610.zip`](https://github.com/Azure/iotedge-lorawan-starterkit/files/7528105/logs_3610.zip)

PR #804, opened 2021-11-15
Bug/798 Make tests using real timer more deterministic

bastbu:
# PR for issue #798

## What is being addressed

With this PR we introduce deterministic timing for two tests that were so far using real timing.

## How is this addressed

We refactor the existing `TestLoRaOperationTimeWatcher` to accept enumerables instead of constant values as input for the elapsed time calls. By using deterministic `GetElapsedTime` calls, the intermittent integration test failure that occurred due to timing issues can be prevented.

Making the tests more deterministic also means that `When_Takes_Too_Long_Getting_FcntDown_Should_Abandon_Message` and `When_Takes_Too_Long_Receiving_First_C2D_Should_Abandon_Message` tested the exact same path, which is why we remove one test in favor of a deterministic test that always runs with the same "`elapsedTimes`".

### DIFF
--- a/Tests/Common/LoRaEnumerable.cs
+++ b/Tests/Common/LoRaEnumerable.cs
@@ -9,7 +9,8 @@ namespace LoRaWan.Tests.Common
     {
         public static IEnumerable<T> Repeat<T>(T value)
         {
-            yield return value;
+            while (true)
+                yield return value;
         }
     }
 }
--- a/Tests/Common/LoRaEnumerable.cs
+++ b/Tests/Common/LoRaEnumerable.cs
@@ -7,7 +7,7 @@ namespace LoRaWan.Tests.Common
 
     internal static class LoRaEnumerable
     {
-        public static IEnumerable<T> Repeat<T>(T value)
+        public static IEnumerable<T> RepeatInfinite<T>(T value)
         {
             while (true)
                 yield return value;
--- a/Tests/Common/LoRaEnumerable.cs
+++ b/Tests/Common/LoRaEnumerable.cs
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Common
+{
+    using System.Collections.Generic;
+
+    internal static class LoRaEnumerable
+    {
+        public static IEnumerable<T> Repeat<T>(T value)
+        {
+            yield return value;
+        }
+    }
+}
--- a/Tests/Integration/CloudToDeviceMessageTests.cs
+++ b/Tests/Integration/CloudToDeviceMessageTests.cs
@@ -991,13 +991,10 @@ namespace LoRaWan.Tests.Integration
 
             var payload = simulatedDevice.CreateUnconfirmedDataUpMessage("1234", fcnt: PayloadFcnt);
             var rxpk = payload.SerializeUplink(simulatedDevice.AppSKey, simulatedDevice.NwkSKey).Rxpk[0];
-            using var request = WaitableLoRaRequest.Create(rxpk, new[]
-            {
-                TimeSpan.FromMilliseconds(100),
-                TimeSpan.FromMilliseconds(100),
-                TimeSpan.FromSeconds(3),
-                TimeSpan.FromSeconds(3)
-            }, PacketForwarder);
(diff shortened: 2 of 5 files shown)


Tests/Common/LoRaEnumerable.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Common
+{
+    using System.Collections.Generic;
+
+    internal static class LoRaEnumerable
+    {
+        public static IEnumerable<T> Repeat<T>(T value)
```
p-schuler, line 10: I would suggest to make this more obvious that it is repeating over and over again without ending. RepeatAlways?

Tests/Integration/CloudToDeviceMessageTests.cs
```diff
@@ -1059,7 +991,13 @@ public async Task When_Takes_Too_Long_Getting_FcntDown_Should_Abandon_Message()
 
             var payload = simulatedDevice.CreateUnconfirmedDataUpMessage("1234", fcnt: PayloadFcnt);
             var rxpk = payload.SerializeUplink(simulatedDevice.AppSKey, simulatedDevice.NwkSKey).Rxpk[0];
-            using var request = CreateWaitableRequest(rxpk, useRealTimer: true);
+            using var request = WaitableLoRaRequest.Create(rxpk, new[]
```
p-schuler, line 994: As much as I like the idea, this does create a hard dependency on the calls to determine the elapsed time. If we add one more call to validate the elapsed time, we will have to change all the tests using this. We have only one at the moment, but eventually, we may want to have more (I know there are more tests trying to emulate exactly this). Not sure a helper would really make things better to have a single method where we can identify them?

Tests/Integration/CloudToDeviceMessageTests.cs
```diff
@@ -1076,7 +1014,7 @@ public async Task When_Takes_Too_Long_Getting_FcntDown_Should_Abandon_Message()
             // 3. Device FcntDown did change
             var devices = loRaDeviceRegistry.InternalGetCachedDevicesForDevAddr(simulatedDevice.DevAddr);
             Assert.True(devices.TryGetValue(simulatedDevice.DevEUI, out var loRaDevice));
-            Assert.Equal(InitialDeviceFcntDown + 1, loRaDevice.FCntDown);
+            Assert.Equal(InitialDeviceFcntDown + Constants.MaxFcntUnsavedDelta, loRaDevice.FCntDown);
```
p-schuler, line 1017: I don't get how we can end up with + 10 here?

bastbu: It's not something that I changed, it came from the existing test. It comes from `SingleGatewayFrameCounterUpdateStrategy.Initialize`.

Tests/Integration/CloudToDeviceMessageTests.cs
bastbu: My thinking was that the only way to truly enforce this would be to make the `WaitableLoRaRequest.Create` that takes an enumerable private and introduce different methods/configurations on `WaitableLoRaRequest`. I decided that with only one test it was not worth it, but given that we can already anticipate that more tests will reuse this exact configuration. I went ahead and implemented it with 9165ee7ea701761457affad4c05195639bed31c2

Tests/Common/LoRaEnumerable.cs
bastbu: 3d6d917fc233e5b26070b75cce4521611ba418c5

Tests/Integration/CloudToDeviceMessageTests.cs
p-schuler: did you debug this? How does it get bumped up? Let me know if you want to step through it.

Tests/Integration/CloudToDeviceMessageTests.cs
p-schuler: nice. much better than what I had in mind.

p-schuler (approved):
Let's figure out the reason for the increased frame counter before getting this in. Otherwise LGTM

Tests/Integration/CloudToDeviceMessageTests.cs
bastbu: As per our discussion, will resolve this now

PR #806, opened 2021-11-15
Remove UdpServer

MaggieSalak:
# PR for issue #770 

## What is being addressed

`UdpServer` class is no longer needed in the LoRaWan.NetworkServer project.

## How is this addressed

- `UdpServer` class was removed
- `UseBasicsStation` configuration property was removed


Mandur:
I would suggest to also remove Use_basic_station from the ci files and deployments

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -1,413 +0,0 @@
...
-            this.RndKeysGenerator.GetBytes(token);
-            return Task.FromResult(token);
-        }
-
-        // Creates a new instance of UdpServer
-        public static UdpServer Create()
-        {
-            var configuration = NetworkServerConfiguration.CreateFromEnvironmentVariables();
-
-            var loRaDeviceAPIService = new LoRaDeviceAPIService(configuration, new ServiceFacadeHttpClientProvider(configuration, ApiVersion.LatestVersion));
```
bastbu, line 56: The `LoRaDeviceAPIService` seems to be unused now, can we remove it as well (same for `ServiceFacadeHttpClientProvider`)?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -1,413 +0,0 @@
...
-            var frameCounterStrategyProvider = new LoRaDeviceFrameCounterUpdateStrategyProvider(configuration, loRaDeviceAPIService);
-            var deduplicationStrategyFactory = new DeduplicationStrategyFactory(loRaDeviceAPIService);
```
bastbu, line 58: This class seems to be unused as well now (or only used from tests). Can we make sure that all the orphaned classes are removed as well? Or would we otherwise need to bring them back later?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
Mandur: LoRaDeviceAPIService is used in https://github.com/Azure/iotedge-lorawan-starterkit/blob/b152c4b6703c41fefbe4c6820ae5b181b5d9bd96/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs#L36

bastbu: Yup, stupid mistake of mine, I was searching for constructor occurrences, not for occurrences of the type

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -1,413 +0,0 @@
...
-                        break;
-                }
-            }
-        }
-
-        private void DispatchMessages(byte[] buffer, DateTime startTimeProcessing)
-        {
-            try
-            {
-                var messageRxpks = Rxpk.CreateRxpk(buffer);
```
bastbu, line 187: This method seems to be unused now, should we move it to the test code entirely/remove it?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
```diff
@@ -1,413 +0,0 @@
...
-                    throw new LoRaProcessingException("Module twin 'FacadeServerUrl' property does not exist.", LoRaProcessingErrorCode.InvalidModuleConfiguration);
-                }
-
-                try
-                {
-                    this.loRaDeviceAPIService.SetAuthCode(moduleTwinCollection["FacadeAuthCode"].Value as string);
-                }
-                catch (ArgumentOutOfRangeException)
-                {
-                    throw new LoRaProcessingException("Module twin 'FacadeAuthCode' does not exist.", LoRaProcessingErrorCode.InvalidModuleConfiguration);
```
bastbu, line 263: `LoRaprocessingErorrCode.InvalidModuleConfiguration` became unused, should we remove it?

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
Mandur: ServiceFacadeHttpClientProvider is used in https://github.com/Azure/iotedge-lorawan-starterkit/blob/b152c4b6703c41fefbe4c6820ae5b181b5d9bd96/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs#L58

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
Mandur: Imho we should rename/migrate the create RXPK and rxpk format with an associated cleanup. I will create a follow up work item for this

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
MaggieSalak: Good point, removed

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/UdpServer.cs
MaggieSalak: Thanks, in that case I'll resolve this comment since we'll tackle it in another PR.

PR #818, opened 2021-11-16
Feature/685 integrate deduplication

spygi:
# PR for issue #685

## What is being addressed

Deduplication service is integrated with the Lns processor. Tests are added covering the integration of these 2 components using web sockets.

## How is this addressed

A test server is being used against which we open multiple web sockets which are coming from the same or different concentrators. 
In order to test the integration between LNS and the deduplication components we need to mock out only the `IMessageDispatcher` and `IBasicsStationConfigurationService`. For this we are changing the adding of services to `TryAdd` which doesn't throw when the service is already registered.
We also need to re-add the lock on the cache in order to have deterministic results on the tests.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -135,7 +138,7 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
                 : throw new InvalidOperationException($"{BasicsStationNetworkServer.RouterIdPathParameterName} was not present on path.");
 
             var channel = new WebSocketTextChannel(socket, sendTimeout: TimeSpan.FromSeconds(3));
-            var handle = socketWriterRegistry.Register(stationEui, channel);
+            var handle = this.socketWriterRegistry.Register(stationEui, channel);
```
spygi, line 141: unrelated.

Tests/Integration/LoRaWan.Tests.Integration.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
spygi, line 1: Required in order to be able to debug tests (running them works fine even without it).

danigian (approved):

PR #826, opened 2021-11-16
Provisioning tool update for adding BasicsStation

danigian:
# PR for issue #850

## What is being addressed

Adding a way to easily "add" (no update) 'Concentrator' type of devices through provisioning tool.


Mandur (approved):
Would be cool to have a little doc on how to use the new feature, or tu use a config other than the default one

danigian:
> Would be cool to have a little doc on how to use the new feature, or tu use a config other than the default one

we should have something in the backlog for non default configs I would say.
Regarding docs, do we have a page I am missing? I thought it was just fine to update the built-in tool option docs

Issue #828, opened 2021-11-17
`LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests` may not be not cleaning up


atifaziz:
## Expected Behavior

Expected test to pass on each invocation.

## Current Behavior

Depending on the test runner in use (e.g. when using ReSharper's runner), the test can run indefinitely due to improper clean-up.

## Additional Information

For example, ReSharper's test running displays the following warning when `LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests` tests are run:

![image](https://user-images.githubusercontent.com/20511/142151272-504855e1-591c-463d-a583-e5de5d643f74.png)

PR #833, opened 2021-11-17
Ensures IHost is stopped and disposed.

spygi:
# PR for issue #828

## What is being addressed

The `IHost` for the `LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests` was not properly stopped and disposed, causing the Resharper runner to hang.

## How is this addressed

By implementing the `IAsyncLifetime` interface. 
ReSharper no longer complains when running these tests and execution completes.

### DIFF
--- a/Tests/Integration/LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.cs
+++ b/Tests/Integration/LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.cs
@@ -22,9 +22,12 @@ namespace LoRaWan.Tests.Integration
     /// <summary>
     /// These tests test the integration between <see cref="LnsProtocolMessageProcessor"/> and <see cref="ConcentratorDeduplication"/>.
     /// </summary>
-    public sealed class LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests : IDisposable
+    public sealed class LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests
+        : IAsyncLifetime
     {
         private IHost testHost;
+
+        public Task InitializeAsync() => Task.CompletedTask;
 
         private async Task CreateTestHostAsync(Mock<IMessageDispatcher> messageDispatcherMock)
         {
@@ -109,9 +112,9 @@ namespace LoRaWan.Tests.Integration
             Assert.Equal(expected, dispatcherCounter);
         }
 
-        public void Dispose()
+        public async Task DisposeAsync()
         {
-            _ = this.testHost?.StopAsync();
+            await this.testHost?.StopAsync();
             this.testHost?.Dispose();
         }
     }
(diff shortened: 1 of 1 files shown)


Tests/Integration/LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.cs
```diff
@@ -22,11 +22,13 @@ namespace LoRaWan.Tests.Integration
     /// <summary>
     /// These tests test the integration between <see cref="LnsProtocolMessageProcessor"/> and <see cref="ConcentratorDeduplication"/>.
     /// </summary>
-    public sealed class LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests
+    public sealed class LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests : IDisposable
     {
-        private static async Task<TestServer> TestServerAsyncFactory(Mock<IMessageDispatcher> messageDispatcherMock)
+        private IHost testHost;
+
+        private async Task CreateTestHostAsync(Mock<IMessageDispatcher> messageDispatcherMock)
```
atifaziz, line 29: Wonder why you changed the return type here? I think it's unrelated and would even give fewer changes in the PR.

Tests/Integration/LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.cs
```diff
@@ -108,5 +108,11 @@ public async Task When_Same_Message_Comes_Multiple_Times_Result_Depends_On_Which
             var expected = isSameStation ? 2 : 1;
             Assert.Equal(expected, dispatcherCounter);
         }
+
+        public void Dispose()
+        {
+            _ = this.testHost?.StopAsync();
```
atifaziz, line 114: Since this is not awaited, I can't imagine how this is working unless `StopAsync` is (incidentally) always completing synchronously before being disposed. I believe the right fix here is to implement `IAsyncLifetime` and stop with an await in `DisposeAsync`.

atifaziz (changes requested):

Tests/Integration/LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.cs
spygi: I think it's a bit clearer this way because what gets created from the method is also stored in the state to get stopped/disposed later on. I consider this a nit and will resolve - let me know if you feel otherwise.

Tests/Integration/LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.cs
spygi: 🤦‍♂️I should have known better here, thanks for the catch.

spygi: Btw not sure why R# complains here, not sure how a NullPointerReference can even occur here
```
var obj = new A();
await obj?.DisposeAsync();

class A
{
	public async Task DisposeAsync()
	{
		await Task.Delay(1);
	}
}
```

atifaziz (approved):

Issue #808, opened 2021-11-15
Ensure that test logs can be correlated with test case


bastbu:
## Description

By using `ITestOutputHelper` in xUnit we can make sure that logs can be correlated with the respective test. This makes it easier for us to debug failing tests, since the logs for the tests can be analyzed as a whole.

## Acceptance criteria

- [ ] `ITestOutputHelper` is used for logging in integration tests

bastbu:
https://xunit.net/docs/capturing-output

bastbu:
A quick spike with the below logger shows that using the `ITestOutputHelper` correlates the standard output with the test case when running in VS. On the console (if using `dotnet test --logger "console;verbosity=detailed"` the statements captured by `ITestOutputHelper` are not automatically prefixed/correlatable with a test case.

This logger:

```csharp
private class TestOutputLogger<T> : ILogger<T>
{
    private const LogLevel TestLogLevel = LogLevel.Debug;
    
    private readonly ITestOutputHelper testOutputHelper;
    
    public TestOutputLogger(ITestOutputHelper testOutputHelper) =>
        this.testOutputHelper = testOutputHelper;
    
    public IDisposable BeginScope<TState>(TState state) => NullDisposable.Instance;
    
    public bool IsEnabled(LogLevel logLevel) => logLevel >= TestLogLevel;
    
    public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
    {
        if (!IsEnabled(logLevel)) return;
        var message = formatter(state, exception);
        this.testOutputHelper.WriteLine(message);
    }
}
```

made sure that the test output was correlated in the test explorer:

![image](https://user-images.githubusercontent.com/22341213/148348217-662ed292-5543-43be-9f78-980ca12bebd8.png)

On the console the output was not automatically prefixed with the test case:

```
FunctionBundler request finished preparing.
FunctionBundler request finished preparing.
FunctionBundler request: {"GatewayId":"foo","ClientFCntUp":1,"ClientFCntDown":0,"Rssi":2.3,"AdrRequest":{"DataRate":2,"RequiredSnr":-15,"PerformADRCalculation":false,"FCntUp":1,"FCntDown":0,"MinTxPowerIndex":7,"GatewayId":"foo","ClearCache":false},"FunctionItems":5}
```

Issue #829, opened 2021-11-17
Logging statements execute a lot of work if log level is not enabled


bastbu:
## Expected Behavior

Expensive log statements, such as `logger.LogDebug($"cloud to device message: {((frmPayload?.Length ?? 0) == 0 ? "empty" : ConversionHelper.ByteArrayToString(frmPayload))}, id: {cloudToDeviceMessage.MessageId ?? "undefined"}, fport: {fport ?? 0}, confirmed: {requiresDeviceAcknowlegement}, cidType: {macCommandType}, macCommand: {macCommands.Count > 0}");` should only be executed if the log level is enabled.

## Current Behavior

Many expensive log statement operations (string interpolations, conversions, etc) are always executed.

Mandur:
@atifaziz looking if needed in .net 6

atifaziz:
Based on an initial look, it seems like that logging did not get any update with regards to optimised and optional string interpolation. Either we can add this support or go with a lightweight level-logging wrapper that short-circuits any work if the right logging level is not in effect. If we want to use structure logging instead, then there are other options still.

PR #842, opened 2021-11-17
Feature/456 logger di

bastbu:
# PR for issue #456

## What is being addressed

With this PR we convert most static logging calls with calls that rely on an instance of `ILogger`.

## How is this addressed

Several strategies helped achieve this goal:
- We inject an `ILogger`/`ILoggerFactory` wherever possible to enable logging
- We migrate the existing `ILogSink` infrastructure for the TCP logging and integrate it with `ILoggerProvider`
- We create internal constructors that inject `NullLogger<T>.Instance` for tests - be aware that this removes the console output of the tests. A follow-up story to address this is #808. Since currently we cannot correlate the log outputs with the test case, we might as well remove them (which greatly simplifies the number of files touched in this PR)
- We rely on log scopes to prepend the device address/EUI/station EUI information to requests.
- We remove some dead code that was relying on the static logger

We will merge this PR into `feature/456-non-static-logger`, and subsequently merge the two PRs together to `dev` after we had a successful CI run.

With this PR we do not address converting to structured logging for all log statements.


spygi:
As discussed with @danigian I will take up this review to save him from some time

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/FunctionBundler/FunctionBundlerPreferredGatewayExecutionItem.cs
```diff
@@ -15,8 +14,6 @@ public void Prepare(FunctionBundlerExecutionContext context, FunctionBundlerRequ
             if (request is null) throw new ArgumentNullException(nameof(request));
 
             request.FunctionItems |= FunctionBundlerItemType.PreferredGateway;
-
-            Logger.Log(context.LoRaDevice.DevEUI, "FunctionBundler ADR request finished preparing.", LogLevel.Debug);
```
spygi, line 19: Is there a reason why this got removed?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -206,6 +209,13 @@ public LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClientConnectionMana
             ClassType = LoRaDeviceClassType.A;
         }
 
+        /// <summary>
+        /// Use constructor for test code only.
+        /// </summary>
+        internal LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClientConnectionManager connectionManager)
```
spygi, line 215: This was added in order to avoid refactoring the tests to use the other constructor right? Seems fair to me.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
```diff
@@ -94,11 +96,11 @@ public async ValueTask<DecodePayloadResult> DecodeMessageAsync(string devEUI, by
                 var url = new Uri($"{toCall}{queryStringParamSeparator}devEUI={devEUIEncoded}&fport={fport}&payload={payloadEncoded}");
 
                 // Call SensorDecoderModule
-                return await CallSensorDecoderModule(devEUI, url);
```
spygi, line 97: thanks for the cleanup 👍

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ReceivedLoRaCloudToDeviceMessage.cs
```diff
@@ -36,10 +35,9 @@ public byte[] GetPayload()
                 {
                     return Convert.FromBase64String(RawPayload);
                 }
-                catch (FormatException ex)
+                catch (FormatException)
                 {
                     // Invalid base64 string, return empty payload
-                    Logger.Log($"Payload '{RawPayload}' is not a valid base64 value: {ex.Message}", LogLevel.Error);
```
spygi, line 42: Logging removed (not sure why) - should we use here as well the `ExceptionFilterUtility` ?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /// <summary>
+        /// Prefixes the message with predefined scope values.
+        /// Right now we only support DevEUI and DevAddr, but could be extended to
+        /// include others (StationEUI, Gateway, etc).
+        /// If the message is already prefixed with the DevEUI/DevAddr, we don't
+        /// add it again.
+        /// </summary>
+        /// <param name="message">The already formatted message</param>
+        /// <returns></returns>
```
spygi, line 23: Can we update the docs? Specifically I think we support stationEUI as a scope right?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -293,22 +292,7 @@ public DownlinkPktFwdMessage Serialize(string appSKey, string nwkSKey, string da
...
-
-                if (devEUI.Length != 0)
-                {
-                    Logger.Log(devEUI, $"{(LoRaMessageType)Mhdr.Span[0]} {jsonMsg}", LogLevel.Debug);
-                }
-                else
-                {
-                    Logger.Log(ConversionHelper.ByteArrayToString(DevAddr.Span.ToArray()), $"{(LoRaMessageType)Mhdr.Span[0]} {jsonMsg}", LogLevel.Debug);
-                }
-            }
```
spygi, line 309: Intentionally removed?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/PhysicalPayload.cs
```diff
@@ -59,7 +58,7 @@ public PhysicalPayload(byte[] input, bool server = false)
                 // TX_ACK That packet type is used by the gateway to send a feedback to the to inform if a downlink request has been accepted or rejected by the gateway.
                 if (Identifier == PhysicalIdentifier.TxAck)
                 {
-                    Logger.Log($"Tx ack received from gateway", LogLevel.Debug);
+                    logger?.LogDebug("Tx ack received from gateway");
```
spygi, line 61: nit: We could avoid the safe navigation by avoid the null default parameter or/and adding an explicit check like we do already for the `input`. Constructor has only 3 usages.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -133,14 +133,6 @@ public virtual bool TryGetJoinChannelIndex(double frequency, out int channelInde
         [Obsolete("#655 - This Rxpk based implementation will go away as soon as the complete LNS implementation is done.")]
         public abstract bool TryGetDownstreamChannelFrequency(Rxpk upstreamChannel, out double frequency, DeviceJoinInfo deviceJoinInfo = null);
 
-        /// <summary>
-        /// Implements logic to get the correct downstream transmission frequency for the given region based on the upstream channel frequency.
-        /// </summary>
-        /// <param name="upstreamFrequency">Frequency of the upstream message.</param>
-        /// <param name="dataRate">Ustream data rate.</param>
-        /// <param name="deviceJoinInfo">Join info for the device, if applicable.</param>
-        public abstract bool TryGetDownstreamChannelFrequency(double upstreamFrequency, ushort dataRate, out double downstreamFrequency, DeviceJoinInfo deviceJoinInfo = null);
```
spygi, line 142: Additional cleanup, thanks 🙌

Tests/Unit/Logger/LoRaConsoleLoggerTests.cs
```diff
@@ -69,6 +69,51 @@ public void When_Message_Already_Prefixed_Scope_Should_Not_Be_Added()
...
         }
 
+        [Theory]
+        [InlineData("EUI", null, null, "EUI foo")]
+        [InlineData(null, "ADDR", null, "ADDR foo")]
+        [InlineData(null, null, 1, "0000000000000001 foo")]
+        [InlineData("EUI", null, 1, "EUI foo")]
+        [InlineData("EUI", "ADDR", null, "EUI foo")]
+        [InlineData(null, "ADDR", 1, "ADDR foo")]
+        public void When_DevEUI_And_DevAddr_Scope_Set_DevEUI_Scope_Is_Set(string devEuiScope, string devAddrScope, int? stationEuiScope, string expected)
```
spygi, line 79: [nit]Test name could be improved e.g. `When_Multiple_Scopes_Set_DevEUI_Preferred_Over_DevAddr_Preferred_Over_StationEUI`

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/TcpLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+    using System.Collections.Concurrent;
+    using System.Net;
+    using System.Net.Sockets;
+    using LoRaWan;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Logging.Configuration;
+
+    internal sealed class TcpLoggerProvider : ILoggerProvider
```
spygi, line 18: This type is only used in `TcpLoggerExtensions`, can we make it nested there and also make it private to limit its visibility?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -48,6 +49,14 @@ public void ConfigureServices(IServiceCollection services)
                             _ = loggingBuilder.SetMinimumLevel(logLevel);
                             _ = loggingBuilder.AddLoRaConsoleLogger(c => c.LogLevel = logLevel);
 
+                            if (NetworkServerConfiguration.LogToTcp)
+                            {
+                                _ = loggingBuilder.AddTcpLogger(new TcpLoggerConfiguration(TcpLoggerConfiguration.InitLogLevel(NetworkServerConfiguration.LogLevel),
```
spygi, line 54: Am I right that we add this in production only for E2E tests? There is no other way around it, right? (i.e. our tests check if the logs appear to assert the outcome of the test)

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerHelper.cs
spygi: Can we add a note about the order of scopes defined being important?

spygi (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            return message;
+
+            static string? GetScopeIfNotInMessage(IDictionary<string, object> dict, string key, string message) =>
+                dict.TryGetValue(key, out var o) &&
+                o is string result &&
+                !message.StartsWith(result, StringComparison.OrdinalIgnoreCase)
+                ? result
+                : null;
```
spygi, line 60: Can we improve readability here with some spacing?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+                };
+#pragma warning restore format
+            }
+
+            return message;
+
+            static string? GetScopeIfNotInMessage(IDictionary<string, object> dict, string key, string message) =>
+                dict.TryGetValue(key, out var o) &&
+                o is string result &&
+                !message.StartsWith(result, StringComparison.OrdinalIgnoreCase)
```
spygi, line 58: Do you test if the message already contains the scope? Maybe it was meant to happen in `LoRaConsoleLoggerTests.DevAddr_Scope_Set_In_Message` test but the message initially does not contain a scope.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -221,10 +223,14 @@ internal static class DownlinkMessageBuilder
...
             // todo: check the device twin preference if using confirmed or unconfirmed down
-            Logger.Log(loRaDevice.DevEUI, $"sending a downstream message with ID {ConversionHelper.ByteArrayToString(rndToken)}", LogLevel.Information);
-            return new DownlinkMessageBuilderResponse(
-                ackLoRaMessage.Serialize(loRaDevice.AppSKey, loRaDevice.NwkSKey, datr, freq, tmst, loRaDevice.DevEUI, lnsRxDelay, rxpk.Rfch, rxpk.Time, request.StationEui),
-                isMessageTooLong);
+            logger.LogInformation($"sending a downstream message with ID {ConversionHelper.ByteArrayToString(rndToken)}");
+
+            var downlinkPktFwdMessage = ackLoRaMessage.Serialize(loRaDevice.AppSKey, loRaDevice.NwkSKey, datr, freq, tmst, loRaDevice.DevEUI, lnsRxDelay, rxpk.Rfch, rxpk.Time, request.StationEui);
+
+            if (logger.IsEnabled(LogLevel.Debug))
```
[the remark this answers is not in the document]
bastbu: Strictly speaking it's an unrelated change - I just saw a potentially expensive operation and added the guard here. Tracked other occurrences in #829

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
bastbu: Exatly, otherwise we would have had a larger diff. We probably need to rethink this with #808

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
bastbu: I moved this log statement to the two call-sites and refactored the logic a bit.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
bastbu: Yeah, I didn't want to migrate to the DI logger for dead code and decided to remove it instead :)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
bastbu: Yes, that is correct. I did not challenge this approach as this has always been the case, it's just more visible now since we moved the logger to DI. I also don't think there's a way around this - I'm not aware of a less invasive tool we have other than conditional registration of services based on the configuration.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/FunctionBundler/FunctionBundlerPreferredGatewayExecutionItem.cs
bastbu: I moved the log statement to the call site, where we already have a logger injected.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ReceivedLoRaCloudToDeviceMessage.cs
bastbu: Valid point. This is one of the few logging statements that I actively removed - I thought it wasn't worth the hassle to inject a logger from all call sites into this method. Tagging @Mandur to see whether I should bring this back or whether it's ok to resolve.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerHelper.cs
bastbu: Good catch, updated with 8a39f0caaa8319ef0fefa1f28a69221c7bb5b174

LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
```diff
@@ -24,12 +24,14 @@ private sealed class RedisLockWrapper : IDisposable
             private readonly string lockKey;
             private readonly string owner;
             private readonly IDatabase redisCache;
+            private readonly ILogger logger;
```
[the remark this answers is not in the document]
bastbu: Removed the logger entirely with ee4654cf68e8656e0f0d03578041ee8d0bba8a35

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/PhysicalPayload.cs
bastbu: I don't think it's worth it - when I followed the path of usages I saw that it is only used from test code (or code that ended up being used in test code). It's related to the comment here: https://github.com/Azure/iotedge-lorawan-starterkit/pull/806#discussion_r750156094, apparently we have an issue tracking to move this code to the test projects.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/TcpLogger.cs
bastbu: I tried to be consistent with the `LoRaConsoleLoggerProvider`. It doesn't seem right to me to hide a class that is required for custom loggers within extensions - open to discuss this if we should pick it up again.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerHelper.cs
bastbu: Definitely, came up with something that reads better in bf43bc9af85d96fc47f2d786110385a471ef835c

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/LoggerHelper.cs
bastbu: There was an existing test for this: `When_Message_Already_Prefixed_Scope_Should_Not_Be_Added`

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
spygi: Fair points. Let's leave this open, curious to hear what the other reviewers have in mind (you can resolve if you want to merge).

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/TcpLogger.cs
bastbu: Addressed with f8ebe5e28f671cebf0f3f306a48ebc7d474391fb

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ReceivedLoRaCloudToDeviceMessage.cs
Mandur: Hmmm imho this looks like an important log as it reacts on malformed input by the user. I was wondering about throwing catch it locally to abandon the message and rethrow, catching on the global filter to log the exception.

**I would anyway write an issue and tackle it in a different work item to be able to calmly do that**

bastbu: Created #857 for this.

Issue #830, opened 2021-11-17
Add deduplication for join requests


spygi:
## Description

This is a follow-up to #638 

## Acceptance criteria

- Join requests are deduplicated
- Code is reused between the join requests deduplication and upstream data frames.

PR #847, opened 2021-11-18
Feature/830 dedup join requests

spygi:
# PR for issue #830 

## What is being addressed

Adds deduplication of JoinRequests, similarly to what was done previously for the data frames.

## How is this addressed

- Generics are being used for type safety and to avoid duplicating code.
  - We register at once the concrete implementations in the startup class.
- Unit and integration tests are added.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -29,18 +31,18 @@ internal sealed class ConcentratorDeduplication : IConcentratorDeduplication, ID
...
-            ILogger<IConcentratorDeduplication> logger)
+            ILogger<IConcentratorDeduplication<T>> logger)
         {
             this.cache = cache;
             this.socketRegistry = socketRegistry;
             this.logger = logger;
         }
 
-        public bool ShouldDrop(UpstreamDataFrame updf, StationEui stationEui)
+        public bool ShouldDrop(T frame, StationEui stationEui)
```
MaggieSalak, line 41: [nit] Shall we use `message` here as well like you do in `CreateCacheKey()`?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -71,7 +73,15 @@ public bool ShouldDrop(UpstreamDataFrame updf, StationEui stationEui)
             return false;
         }
 
-        internal static string CreateCacheKey(UpstreamDataFrame updf)
+        internal static string CreateCacheKey(T message)
+            => message switch
+            {
+                UpstreamDataFrame asDataFrame => CreateCacheKeyCore(asDataFrame),
+                JoinRequestFrame asJoinRequestFrame => CreateCacheKeyCore(asJoinRequestFrame),
+                _ => throw new ArgumentException($"{message} not of proper type")
```
MaggieSalak, line 81: [nit] "{message} with invalid type"

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -86,6 +96,23 @@ internal static string CreateCacheKey(UpstreamDataFrame updf)
             return BitConverter.ToString(key);
         }
 
+        private static string CreateCacheKeyCore(JoinRequestFrame joinReq)
+        {
+            var totalBufferLength = JoinEui.Size + DevEui.Size + DevNonce.Size;
+            Span<byte> buffer = stackalloc byte[totalBufferLength];
```
MaggieSalak, line 102: So in this case it's guaranteed that the length will not exceed 128?

MaggieSalak (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Correct, because all of the fields we are using (JoinEui, DevEui and DevNonce) have a fixed size: 8 + 8 + 2 bytes in total. Unless we change these primitives drastically of course. 

In the `CreateCacheKeyCore(UpstreamDataFrame)` we include also the payload in the buffer which has a dynamic size, that's why if it's smaller than 128 bytes we use the stack, otherwise the heap.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Done d451eb10cceaad924c92e99b4ae88d5141b668f4

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Actually, I think frame is better suited here as `UpstreamDataFrame` and `JoinRequestFrame` both have frame in their names. Refactored to use frame everywhere (except for log messages where we should be user-friendlier) in d451eb10cceaad924c92e99b4ae88d5141b668f4

Issue #675, opened 2021-11-01
Current Basic station implementation only support number for setting log level


Mandur:
When deploying if we set Log_level to "debug" it crashes with the following message:

```
Unhandled exception. System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at System.Number.ParseInt32(ReadOnlySpan`1 value, NumberStyles styles, NumberFormatInfo info)
   at System.Int32.Parse(String s, IFormatProvider provider)
   at LoRaWan.NetworkServer.BasicsStation.BasicsStationNetworkServerStartup.<ConfigureServices>b__7_0(ILoggingBuilder loggingBuilder) in /build/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs:line 33
   at Microsoft.Extensions.DependencyInjection.LoggingServiceCollectionExtensions.AddLogging(IServiceCollection services, Action`1 configure)
   at LoRaWan.NetworkServer.BasicsStation.BasicsStationNetworkServerStartup.ConfigureServices(IServiceCollection services) in /build/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs:line 31
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.InvokeCore(Object instance, IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.<>c__DisplayClass9_0.<Invoke>g__Startup|0(IServiceCollection serviceCollection)
   at Microsoft.AspNetCore.Hosting.StartupLoader.ConfigureServicesDelegateBuilder`1.<>c__DisplayClass15_0.<BuildStartupServicesFilterPipeline>g__RunPipeline|0(IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.Invoke(Object instance, IServiceCollection services)
[...]

PR #848, opened 2021-11-18
Allow setting the log level in word form

MaggieSalak:
# PR for issue #675 

## What is being addressed

LNS was crashing if it was configured to use log level expressed in a word (e.g. "Debug"). Only number representation of the `LogLevel` enum was allowed. 

## How is this addressed

The logic in `BasicsStationNetworkServerStartup.ConfigureServices()` was adjusted to also allow word representation (case insensitive) of the log level to be provided. The numerical version is still supported.

## How was the fix verified

The bug was first reproduced by setting the log level in `LOG_LEVEL` variable of the LNS module to "Debug".
The fix was verified by deploying the updated LNS module and checking that the LNS doesn't crash and the log contains statements from the expected level.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
@@ -3,6 +3,7 @@
 
 namespace LoRaWan.NetworkServer.BasicsStation
 {
+    using System;
     using System.Globalization;
     using LoRaTools.ADR;
     using LoRaWan;
@@ -40,7 +41,10 @@ namespace LoRaWan.NetworkServer.BasicsStation
             _ = services.AddLogging(loggingBuilder =>
                         {
                             _ = loggingBuilder.ClearProviders();
-                            var logLevel = (LogLevel)int.Parse(NetworkServerConfiguration.LogLevel, CultureInfo.InvariantCulture);
+                            var logLevel = int.TryParse(NetworkServerConfiguration.LogLevel, out var logLevelNum)
+                                ? (LogLevel)logLevelNum
+                                : Enum.Parse<LogLevel>(NetworkServerConfiguration.LogLevel, true);
+
                             _ = loggingBuilder.SetMinimumLevel(logLevel);
                             _ = loggingBuilder.AddLoRaConsoleLogger(c => c.LogLevel = logLevel);
 
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
@@ -41,8 +41,8 @@ namespace LoRaWan.NetworkServer.BasicsStation
             _ = services.AddLogging(loggingBuilder =>
                         {
                             _ = loggingBuilder.ClearProviders();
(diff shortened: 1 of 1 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -40,7 +41,10 @@ public void ConfigureServices(IServiceCollection services)
             _ = services.AddLogging(loggingBuilder =>
                         {
                             _ = loggingBuilder.ClearProviders();
-                            var logLevel = (LogLevel)int.Parse(NetworkServerConfiguration.LogLevel, CultureInfo.InvariantCulture);
+                            var logLevel = int.TryParse(NetworkServerConfiguration.LogLevel, out var logLevelNum)
```
Mandur, line 44: Shoudl we use the overload TryParse (string? s, System.Globalization.NumberStyles style, IFormatProvider? provider, out int result); to be stull using invariant culture?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -40,7 +41,10 @@ public void ConfigureServices(IServiceCollection services)
...
+                                ? (LogLevel)logLevelNum
```
atifaziz, line 45: C# allows integers conversions to enums to always succeed so use [`Enum.IsDefined`](https://docs.microsoft.com/en-us/dotnet/api/system.enum.isdefined) to ensure the parsed integer is indeed a defined value of `LogLevel`.

```suggestion
                                ? (LogLevel)logLevelNum is var level && Enum.IsDefined(typeof(LogLevel), level) ? level : throw new InvalidCastException()
```

MaggieSalak: Done in b3c7f3921621aa11d8efa6036a95f8e43788fee6

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
MaggieSalak: Done in b3c7f3921621aa11d8efa6036a95f8e43788fee6

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -40,7 +41,10 @@ public void ConfigureServices(IServiceCollection services)
             _ = services.AddLogging(loggingBuilder =>
                         {
                             _ = loggingBuilder.ClearProviders();
-                            var logLevel = (LogLevel)int.Parse(NetworkServerConfiguration.LogLevel, CultureInfo.InvariantCulture);
+                            var logLevel = int.TryParse(NetworkServerConfiguration.LogLevel, NumberStyles.Integer, CultureInfo.InvariantCulture, out var logLevelNum)
+                                 ? (LogLevel)logLevelNum is var level && Enum.IsDefined(typeof(LogLevel), level) ? level : throw new InvalidCastException()
```
atifaziz, line 45: nit: fix indentation:

```suggestion
                                ? (LogLevel)logLevelNum is var level && Enum.IsDefined(typeof(LogLevel), level) ? level : throw new InvalidCastException()
```

atifaziz (approved):

Issue #391, opened 2021-09-28
[Story] ADR on CUPS protocol (how we manage certs, which ones do we need, ...)


roel4ez:
- [x] ADR document reviewed and available in repo
- [x] Implementation stories created on the backlog

Issue #394, opened 2021-09-28
CUPS Firmware upgrade


roel4ez:
- [x] #1189
- [x] #1245
- [x] #1247
- [x] #1248
- [x] #1249
- [x] #1250
- [x] #1251
- [x] #1252
- [x] #1345
- [x] #1299

roel4ez:
@mandur maybe it makes sense to take this story out and make it into its own feature inside #344

MaggieSalak:
I think we can close this story @danigian @roel4ez ?

PR #853, opened 2021-11-19
CUPS ADR

danigian:
# PR for issue #391 

## What is being addressed

Credentials management and CUPS Implementation ADR

### DIFF
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -74,7 +74,7 @@ For increased security the server should "require" client authentication and ver
 
 Therefore, in order for Basic Station to connect to the CUPS HTTPS endpoint the following files should be provided to the station:
 
-- cups.uri (i.e. "https://LoRaWanNetworkSrvModule:443")
+- cups.uri
 - cups.crt (including the certificate in PEM format)
 - cups.key (including the EC Private Key for the child certificate)
 - cups.trust (including the chain of trust certificate for the server-side certificate)
@@ -270,7 +270,7 @@ For generating both server and client certificates there are two options.
 
 In this starter kit, we are assuming that the same server certificate is used for both CUPS and LNS endpoints.
 
-Using instructions at ['Basics Station credentials management'](../../user-guide/station-credential-management/) page, copy the certificate so that the LoRaWan Network Server can retrieve it at its startup and use it within Kestrel.
+Using instructions at ['Basics Station credentials management'](../user-guide/station-credential-management.md) page, copy the certificate so that the LoRaWan Network Server can retrieve it at its startup and use it within Kestrel.
 
 For concentrator configuration, it is important to retrieve the chain of trust which is used for signing the CUPS/LNS server certificate.
 
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -145,7 +145,7 @@ LoRa Device Provisioning tool is not uploading the credential bundle files to th
     Concentrator->>CUPS Server: POST /update-info
     CUPS Server->>IoT Hub: Retrieve 'Concentrator' twin
     CUPS Server->>CUPS Server: Verifies 'Concentrator' client certificate
-    Note right of CUPS Server: Logic for client certificate validation to be defined. First version including a 'cupsClientThumbprint' to verify against.
+    Note right of CUPS Server: Logic for client certificate validation to be defined. First version including a 'cupsClientThumbprint' to verify against. The client certificate thumbprint should be reported as a property.
     alt client certificate not valid
         CUPS Server->>Concentrator: Return forbidden
     else client certificate valid
@@ -189,14 +189,14 @@ The following "desired" properties should be properly set:
     "tcCredentialUrl": "https://..."
 },
 // Following field to be used for first version client certificate validation
-"clientThumbprint": "Client certificate thumbprint"
+"clientThumbprint": [ "Client certificate thumbprint" ]
 ```
 
 - **'cupsCredCrc'**: should be computed as CRC32 checksum calculated over the concatenated credentials files `cups.{trust,cert,key}`
 - **'tcCredCrc'**: should be computed as CRC32 checksum calculated over the concatenated credentials files `tc.{trust,cert,key}`
 - **'cupsCredentialUrl'**: should point to the blob/secret containing the concatenated credentials `cups.{trust,cert,key}`
 - **'tcCredentialUrl'**: should point to the blob/secret containing the concatenated credentials `tc.{trust,cert,key}`
-- **'cupsClientThumbprint'**: should include the thumbprint of the client certificate used for authenticating against the CUPS server
+- **'clientThumbprint'**: should include the thumbprint of the client certificate used for authenticating against the CUPS server. It is an array for allowing an old thumbprint to be accepted until the rotation mechanism completes
 
 ### Facade Azure Function related changes
 
@@ -248,6 +248,18 @@ The CLI should be changed in order to:
   - upload, via Azure Blob Storage C# SDK, the credential files to the Blob Storage
   - update the Twin for the concentrator device accordingly
 - In case a device being created is not making use of CUPS, tool should provide a `--no-cups` option, allowing to just specify a `clientThumbprint` (for client certificate validation on LNS endpoint)
+- In case a device is already created and a credential rotation is required, the tool should:
+  - provide a "update" mechanism that:
+    - accept as input a cupsCredentials file (being the concatenation of `cups.{trust,cert,key}`)
+    - accept as input a tcCredentials file (being the concatenation of `tc.{trust,cert,key}`)
+    - accept as input a clientThumbprint string (for client certificate validation). New thumbprint is appended to the existing one.
+    - compute the CRC32 of above mentioned files
+    - create a device for the concentrator
+    - upload, via Azure Blob Storage C# SDK, the credential files to the Blob Storage
+    - update the Twin for the concentrator device accordingly
+  - provide a "revoke" mechanism that:
+    - accept as input an old clientThumbprint to be removed from the twin
+    - check that reported thumbprint property is not equal to the thumbprint that we want to remove (this means that the basic station has not rotated certificate yet)
 
 ## Appendix
 
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -154,7 +154,7 @@ LoRa Device Provisioning tool is not uploading the credential bundle files to th
         CUPS Server->>Concentrator: Return forbidden
     else client certificate valid
         alt no cups config available
-            CUPS Server->>Concentrator: Return no updates
+            CUPS Server->>CUPS Server: Throw as this is an indication of misconfiguration
         else cups config available
             alt different cupsUri
                 alt different cupsCredCrc
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -102,15 +102,15 @@ There is no need to copy/paste manually these files in the Basic Station; the CU
 
 There are multiple options of where to store the `cups.{trust,cert,key}` and `tc.{trust,cert,key}` bundle files.
 
-#### a. IoT Hub linked Azure Storage Account
+#### a. Azure Storage Account
 
 This is the easiest option for a new user of the Starter Kit, even though it's not the safest one.
 
-When deploying the Starter Kit, an Azure Storage Account is linked to the IoT Hub being created for the devices.
+When deploying the Starter Kit, an Azure Storage Account is provisioned for Azure Functions.
 
-LoRa Device Provisioning tool is capable of uploading the CLI provided bundle files and properly update the twin for the concentrator device being created in IoT Hub.
+LoRa Device Provisioning tool should be capable of uploading the CLI provided bundle files and properly update the twin for the concentrator device being created in IoT Hub.
 
-When using this option, the 'Facade' Azure Function will retrieve certificate files from such storage account by using a Managed Identity.
+When using this option, the 'Facade' Azure Function will retrieve certificate files from such storage account.
 
 #### b. Azure Key Vault secret
 
@@ -175,15 +175,7 @@ LoRa Device Provisioning tool is not uploading the credential bundle files to th
 
 ### IoT Hub related changes
 
-Two changes are expected for IoT Hub.
-
-First one is related to the ability to associate an Azure storage account with IoT Hub.
-
-One of the possible 'sinks' of where the credential files are going to be stored is the IoT Hub linked Azure Blob Storage.
-
-For this option to work, you must [associate an Azure Storage account to IoT Hub](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-csharp-csharp-file-upload#associate-an-azure-storage-account-to-iot-hub). Any LoRaWan Starter Kit IoT Hub, should have this option enabled by default, therefore some changes are required in the way the template is deployed.
-
-Second change is related to the concentrator device twin.
+Only change is related to the concentrator device twin.
 
 The following "desired" properties should be properly set:
 
@@ -230,10 +222,7 @@ Note right of LoRaWanNetworkSrvModule: Response should be a json including base6
 
 In addition to this, the Azure Function must be able to properly authenticate to target sink via Managed Identity (when using Azure Blob Storage or KeyVault).
 
-Considering that the default sink for the starter kit should be an Azure Storage Account linked to the IoT Hub, the template for the starter kit should be changed in such a way that:
-
-- a managed identity is created for the Azure Function
-- a proper [RBAC](https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#reader-and-data-access) built-in role assignment is created for allowing the Function to access the above mentioned Storage Account
+Considering that the default sink for the starter kit should be an Azure Storage Account, the template for the starter kit should be changed in such a way that a new Blob Container is created for uploading credential blob files.
 
 If using KeyVault secrets instead of Blob Storage, instructions on how to create a role assignment should be provided.  
 
@@ -256,7 +245,7 @@ The CLI should be changed in order to:
   - accept as input a clientThumbprint string (for client certificate validation)
   - compute the CRC32 of above mentioned files
   - create a device for the concentrator
-  - upload, via IoT Hub C# SDK, the credential files to the Blob Storage
+  - upload, via Azure Blob Storage C# SDK, the credential files to the Blob Storage
   - update the Twin for the concentrator device accordingly
 - In case a device being created is not making use of CUPS, tool should provide a `--no-cups` option, allowing to just specify a `clientThumbprint` (for client certificate validation on LNS endpoint)
 
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -70,7 +70,7 @@ In order to keep the starter kit simple, the same set of server and client crede
 
 The CUPS endpoint should support mutual TLS authentication between client (basic station) and server (network server IoT Edge module).
 
-For increased security the server should "require" client authentication and verify the provided credentials.
+If not manually disabled via flags, for increased security, the server has to "require" client authentication and verify the provided credentials.
 
 Therefore, in order for Basic Station to connect to the CUPS HTTPS endpoint the following files should be provided to the station:
 
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -48,7 +48,7 @@ LoRa Basics™ Station supports four different authentication modes:
 - TLS Server and Client Authentication
 - TLS Server Authentication and Client Token
 
-An authentication mode can be set by configuring some files in the device where Basics Station is going to be executed.
+An authentication mode can be set by configuring some files in the device where Basics Station is going to be executed. More documentation [here](https://doc.sm.tc/station/credentials.html)
 
 Basics Station can either use two different sets of credentials for achieving client authentication to the CUPS endpoint and to the LNS endpoint or re-use the same set of credentials for both.
 
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -66,6 +66,8 @@ Driving this decision is the willing of increasing the overall security of the s
 
 In order to keep the starter kit simple, the same set of server and client credentials are to be used for both CUPS and LNS protocol endpoints.
 
+For the same simplicity reason, credential bundles are going to be stored in an Azure Storage Account.
+
 ### Authentication to CUPS endpoint
 
 The CUPS endpoint should support mutual TLS authentication between client (basic station) and server (network server IoT Edge module).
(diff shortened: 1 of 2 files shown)


docs/adr/006_cups.md
```diff
@@ -0,0 +1,296 @@
...
+
+The aim of this document section is to take and record a decision for:
+
+- Which authentication mode needs to be used for connecting to CUPS endpoint
+- Which authentication mode needs to be used for connecting to LNS endpoint
+- Where are the certificates stored and how to retrieve those?
+
+### Decision(s)
+
+Out of the four authentication modes we should aim for "TLS Server and Client Authentication" for both CUPS and LNS endpoints.
```
Mandur, line 63: [Brainstorm] Not sure it's a good idea, but Like to discuss

What if we used the client token auth instead of certificate for client. We could generate a client token with the station's private key in IoT Hub (signature-string). We would then verify the token by simply trying to connect to the IoT Hub.

The advantage I see is No need to provision a client certificate on the LNS for validation as the authorization procedure is offloaded to the iot hub.

docs/adr/006_cups.md
```diff
@@ -0,0 +1,296 @@
...
+
+When deploying the Starter Kit, an Azure Storage Account is linked to the IoT Hub being created for the devices.
+
+LoRa Device Provisioning tool is capable of uploading the CLI provided bundle files and properly update the twin for the concentrator device being created in IoT Hub.
+
+When using this option, the 'Facade' Azure Function will retrieve certificate files from such storage account by using a Managed Identity.
+
+#### b. Azure Key Vault secret
+
+This option requires more manual intervention, as the LoRa Device Provisioning tool is not adapted for uploading the credential bundle files to an Azure Key Vault instance.
```
Mandur, line 117: Could we use the function to have additional automation there

docs/adr/006_cups.md
```diff
@@ -0,0 +1,296 @@
...
+        alt no cups config available
+            CUPS Server->>Concentrator: Return no updates
+        else cups config available
+            alt different cupsUri
+                alt different cupsCredCrc
+                    CUPS Server->>Facade Function: Retrieve CUPS Credential Blob
+                    Facade Function->>CUPS Server: CUPS Credential Blob
+                    CUPS Server->>Concentrator: Updated cupsUri and cupsCred
+                else equal cupsCredCrc
+                    CUPS Server->>Concentrator: Updated cupsUri
```
Mandur, line 161: Sorry to ask, this means you can point cups to another cups server?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,296 @@
...
+
+### IoT Hub related changes
+
+Two changes are expected for IoT Hub.
+
+First one is related to the ability to associate an Azure storage account with IoT Hub.
+
+One of the possible 'sinks' of where the credential files are going to be stored is the IoT Hub linked Azure Blob Storage.
+
+For this option to work, you must [associate an Azure Storage account to IoT Hub](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-csharp-csharp-file-upload#associate-an-azure-storage-account-to-iot-hub). Any LoRaWan Starter Kit IoT Hub, should have this option enabled by default, therefore some changes are required in the way the template is deployed.
```
Mandur, line 184: I am not sure I understand well this. Based on the doc it looks like this is necessary only if a device need to upload file. As far as i understand we don't need to upload files, as the upload would be carried by the provsiioning tool no`?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Facade Azure Function related changes
+
+The Azure Function should implement a new endpoint 'FetchConcentratorCredentials' which executes the following flow:
+
+```mermaid
+sequenceDiagram
+LoRaWanNetworkSrvModule->>Function: POST /fetchConcentratorCredentials
+Note right of LoRaWanNetworkSrvModule: Request includes 'stationEui' and 'credentialsType' (cups/tc/both)
+Function->>IoT Hub: Retrieve Concentrator device Twin
```
Mandur, line 217: An alternative would be to make the secret URL as part of the request as we have it locally

docs/adr/006_cups.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+In addition to this, the Azure Function must be able to properly authenticate to target sink via Managed Identity (when using Azure Blob Storage or KeyVault).
+
+Considering that the default sink for the starter kit should be an Azure Storage Account linked to the IoT Hub, the template for the starter kit should be changed in such a way that:
+
+- a managed identity is created for the Azure Function
+- a proper [RBAC](https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#reader-and-data-access) built-in role assignment is created for allowing the Function to access the above mentioned Storage Account
+
+If using KeyVault secrets instead of Blob Storage, instructions on how to create a role assignment should be provided.
```
Mandur, line 238: Imho in this case we need to provision this as part of the starterkit template

docs/adr/006_cups.md
danigian: Agree that there's no need to provision a client certificate, but there would be need to create a tool generating such token and anyways you would need to provision the token to the device, no?

docs/adr/006_cups.md
danigian: Can you expand more on the option of having the function as "provisioning proxy endpoint"?

Anyways I would tackle the revisitation of security around function endpoint in a different PR

docs/adr/006_cups.md
danigian: Yes, a cups server can point to another cups server (as far as i can read from official docs, not tested in real life)

docs/adr/006_cups.md
danigian: The idea would be that the provisioning tool is instantiating a device client connection to upload the certificates related to such device.

We can revisit this if you have other ideas

docs/adr/006_cups.md
danigian: It depends on what we want to give as "default" option for the Starter Kit. To be discussed if we want to keep it simple and just use blob storage or we want to use KeyVault 👍

docs/adr/006_cups.md
danigian: As we discussed, we should not make the function more complex than what it should be. Its purpose should be only the one of downloading the client certificates from Storage or Key Vault.

I will resolve this conversation, feel free to reopen if needed. @Mandur

docs/adr/006_cups.md
danigian: As discussed, I have removed any reference to the IoT Hub Storage Account link. Let's keep it simple and use the Azure Function Storage Account

Resolving this conversation, feel free to reopen if needed @Mandur

docs/adr/006_cups.md
danigian: As we discussed via Teams, let's keep it simple and let's start with Azure Blob Storage for now.
I will resolve this conversation, feel free to reopen if needed. @Mandur

docs/adr/006_cups.md
```diff
@@ -0,0 +1,285 @@
...
+Therefore, in order for Basic Station to connect to the LNS WSS endpoint the following files should be provided to the station:
+
+- tc.uri (i.e.: "wss://LoRaWanNetworkSrvModule:5001")
+- tc.crt (including the certificate in PEM format)
+- tc.key (including the EC Private Key for the child certificate)
+- tc.trust (including the chain of trust certificate for the server-side certificate)
+
+Same certificate files as above are expected to be used for the Starter Kit, even though you might want to consider increasing security of the solution by generating different certificates for this endpoint.
+
+There is no need to copy/paste manually these files in the Basic Station; the CUPS Protocol Implementation later described in this document will retrieve the needed credentials from a centralized repository and update the Basic Station configuration when it starts. This option is also providing an option to rotate certificates when these are near the expiration.
```
bastbu, line 99: How do we plan to invalidate the old credentials after rotation?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,285 @@
...
+- LoRa Basics™ Station is configured with a cups.uri, a cups.trust and mandatory cups.cert/key
+- The concentrator device twin should include changes as described in following section
+
+#### Flow
+
+```mermaid
+  sequenceDiagram
+    autonumber
+    Concentrator->>CUPS Server: POST /update-info
+    CUPS Server->>IoT Hub: Retrieve 'Concentrator' twin
```
bastbu, line 146: Do we need to protect against rogue LBSes (always reconnecting) to avoid going to the IoT hub many times?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,285 @@
...
+                alt different cupsCredCrc
+                    CUPS Server->>Facade Function: Retrieve CUPS Credential Blob
+                    Facade Function->>CUPS Server: CUPS Credential Blob
+                    CUPS Server->>Concentrator: Updated cupsUri and cupsCred
+                else equal cupsCredCrc
+                    CUPS Server->>Concentrator: Updated cupsUri
+                end
+            else equal cupsUri
+                alt different cupsCredCrc
+                    CUPS Server->>Facade Function: Retrieve CUPS Credential Blob
```
bastbu, line 165: Can we make the difference clearer why we have this twice in the flow diagram?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,285 @@
...
+
+The CLI should be changed in order to:
+
+- Specify if the device being created is making use of CUPS or not
+- In case a device being created is making use of CUPS, the tool should:
+  - accept as input a cupsCredentials file (being the concatenation of `cups.{trust,cert,key}`)
+  - accept as input a tcCredentials file (being the concatenation of `tc.{trust,cert,key}`)
+  - accept as input a clientThumbprint string (for client certificate validation)
+  - compute the CRC32 of above mentioned files
+  - create a device for the concentrator
```
bastbu, line 247: Maybe it would be worth mentioning certs rotation

docs/adr/006_cups.md
```diff
@@ -0,0 +1,296 @@
...
+- Defining the changes needed in 'Facade' Azure Function for handling updated credential retrieval
+- Defining the changes needed in LoRaWan Network Server for handling above mentioned flows
+- Defining the changes needed in LoRa Device Provisioning CLI for handling creation/update of concentrators with certificate support
+
+## Out-of-scope
+
+Out of scope for this document is:
+
+- Defining a flow diagram for firmware update process
+- Defining any change needed in IoT Hub / Azure Function / Network Server for handling firmware updates
```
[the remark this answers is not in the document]
danigian: #394 is the story that tackles firmware upgrade

docs/adr/006_cups.md
danigian: I have just added some details of changes needed in "LoRa Device Provisioning CLI".

The idea, to be documented is the following:

- IT employee creates new client certificates for a concentrator
- IT employee uploads new certificates to concentrator
- IT employee issues an "update" command with the Device Provisioning CLI
- After verifying that new credentials were downloaded (automatically after 1day or by unplugging-replugging device for forcing CUPS), a "revoke" command needs to be issued with the Device Provisioning CLI to revoke old client certificate thumbprint

Anything that you would add here? @bastbu @Mandur

docs/adr/006_cups.md
danigian: Yes, we should cache the 'Concentrator' twins, but I assume that's an implementation detail. Do you want me to specify it here?

@bastbu

docs/adr/006_cups.md
danigian: Unfortunately github makes your comment hard to understand 😢
What is 'this'? Retrieving CUPS Credential Blob?

docs/adr/006_cups.md
danigian: Added a section already. Would you mind having a look now? @bastbu

danigian (commented):
@Mandur, @bastbu the following implementation stories would be created out of this ADR:

- [ ] Adjust ARM Template to provision container for certificate bundle storage
- [ ] Implement logic for verifying client certificate in Network Server implementation
- [ ] Modify device provisioning CLI to add devices with client certificates
- [ ] Implement CUPS protocol endpoint
- [ ] Modify device provisioning CLI to support certificate rotation
- [ ] Provide clear documentation on supported authentication modes (none, server-only, server+client,server+client w/ cups)
- [ ] Add E2E test for verifying new device scenario
- [ ] Add E2E test for verifying credential rotation scenario

docs/adr/006_cups.md
bastbu: No need to, more for me to be aware about this

docs/adr/006_cups.md
bastbu: Yes. It would be helpful to clarify why would need to query it twice (`Retrieve CUPS Credential Blob` is in there twice)

docs/adr/006_cups.md
bastbu: I saw that I had the comment twice. Sorry!

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+To better understand how the CUPS Protocol should be implemented, it is needed to understand how Basics Station is handling authentication.
+
+LoRa Basics™ Station supports four different authentication modes:
+
+- No Authentication
+- TLS Server Authentication
+- TLS Server and Client Authentication
+- TLS Server Authentication and Client Token
+
+An authentication mode can be set by configuring some files in the device where Basics Station is going to be executed.
```
bastbu, line 51: [nit] maybe a link to the docs which files those are would be helpful for following the flow of the document

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+For increased security the server should "require" client authentication and verify the provided credentials.
+
+Therefore, in order for Basic Station to connect to the LNS WSS endpoint the following files should be provided to the station:
+
+- tc.uri (i.e.: "wss://LoRaWanNetworkSrvModule:5001")
+- tc.crt (including the certificate in PEM format)
+- tc.key (including the EC Private Key for the child certificate)
+- tc.trust (including the chain of trust certificate for the server-side certificate)
+
+Same certificate files as above are expected to be used for the Starter Kit, even though you might want to consider increasing security of the solution by generating different certificates for this endpoint.
```
bastbu, line 97: I would propose to capture a decision on this here (I think above you mentioned that we are not going to do this for now, which seemed reasonable to me). We can always update the ADR if we want to extend this in the future.

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+### Authentication to CUPS endpoint
+
+The CUPS endpoint should support mutual TLS authentication between client (basic station) and server (network server IoT Edge module).
+
+If not manually disabled via flags, for increased security, the server has to "require" client authentication and verify the provided credentials.
+
+Therefore, in order for Basic Station to connect to the CUPS HTTPS endpoint the following files should be provided to the station:
+
+- cups.uri
+- cups.crt (including the certificate in PEM format)
```
bastbu, line 78: Why not use DER format directly as described in the spec?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+- tc.key (including the EC Private Key for the child certificate)
+- tc.trust (including the chain of trust certificate for the server-side certificate)
+
+Same certificate files as above are expected to be used for the Starter Kit, even though you might want to consider increasing security of the solution by generating different certificates for this endpoint.
+
+There is no need to copy/paste manually these files in the Basic Station; the CUPS Protocol Implementation later described in this document will retrieve the needed credentials from a centralized repository and update the Basic Station configuration when it starts. This option is also providing an option to rotate certificates when these are near the expiration.
+
+### Where to store certificates
+
+There are multiple options of where to store the `cups.{trust,cert,key}` and `tc.{trust,cert,key}` bundle files.
```
bastbu, line 103: We should also decide here what we want to do and track that decision. From what I understood storage account would be the way to go initially, let's capture that here.

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+
+In addition to this, the Azure Function must be able to properly authenticate to target sink via Managed Identity (when using Azure Blob Storage or KeyVault).
+
+Considering that the default sink for the starter kit should be an Azure Storage Account, the template for the starter kit should be changed in such a way that a new Blob Container is created for uploading credential blob files.
+
+If using KeyVault secrets instead of Blob Storage, instructions on how to create a role assignment should be provided.  
+
+### LoRaWan Network Server changes
+
+- LoRaDeviceAPIServiceBase and its implementation (LoRaDeviceAPIService) should include a method for calling the 'FetchConcentratorCredentials' endpoint
```
bastbu, line 231: How do we provision the server certs of the LNS?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+- In case a device being created is making use of CUPS, the tool should:
+  - accept as input a cupsCredentials file (being the concatenation of `cups.{trust,cert,key}`)
+  - accept as input a tcCredentials file (being the concatenation of `tc.{trust,cert,key}`)
+  - accept as input a clientThumbprint string (for client certificate validation)
+  - compute the CRC32 of above mentioned files
+  - create a device for the concentrator
+  - upload, via Azure Blob Storage C# SDK, the credential files to the Blob Storage
+  - update the Twin for the concentrator device accordingly
+- In case a device being created is not making use of CUPS, tool should provide a `--no-cups` option, allowing to just specify a `clientThumbprint` (for client certificate validation on LNS endpoint)
+- In case a device is already created and a credential rotation is required, the tool should:
```
bastbu, line 251: How do we plan to support rotation of LNS certs? Since the CUPS endpoint will be part of the same problem, we could have a chicken-and-egg problem there if the client wants to connect to the HTTPS CUPS endpoint, no?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+    autonumber
+    Concentrator->>CUPS Server: POST /update-info
+    CUPS Server->>IoT Hub: Retrieve 'Concentrator' twin
+    CUPS Server->>CUPS Server: Verifies 'Concentrator' client certificate
+    Note right of CUPS Server: Logic for client certificate validation to be defined. First version including a 'cupsClientThumbprint' to verify against. The client certificate thumbprint should be reported as a property.
+    alt client certificate not valid
+        CUPS Server->>Concentrator: Return forbidden
+    else client certificate valid
+        alt no cups config available
+            CUPS Server->>Concentrator: Return no updates
```
bastbu, line 153: Is this an indication for malconfiguration? If the LBS tries to use CUPS and the configuration is not present, should we consider failing instead?

bastbu (approved):

docs/adr/006_cups.md
Mandur: As discussed orally, the above idea would provide an interresting simplification of required implementation. However. if we want to permit rotation of the tc.trust file, the gain would be very low.

docs/adr/006_cups.md
Mandur: I am not sure I follow the revoking part. As far as I understood as soon as we issue the update command the thumbprint in the desired properties would be updated to the new value?

Mandur: Also when we use the provisioning CLI to upload new certs I assumed we would be overwritting the old ones in the azure storage account, correct?

docs/adr/006_cups.md
```diff
@@ -0,0 +1,297 @@
...
+"cups": {
+    "cupsUri": "https://IP_or_DNS:PORT",
+    "tcUri": "wss://IP_or_DNS:PORT",
+    "cupsCredCrc": INT,
+    "tcCredCrc": INT,
+    "cupsCredentialUrl": "https://...",
+    "tcCredentialUrl": "https://..."
+},
+// Following field to be used for first version client certificate validation
+"clientThumbprint": [ "Client certificate thumbprint" ]
```
Mandur, line 192: Why not put this as part of the cups block?

Mandur: As discussed, it is not only for cups and could be used for the tc.uri

docs/adr/006_cups.md
Mandur: We discussed this and afaik the plan is to stay with a cert on the file system as introduced in #734

docs/adr/006_cups.md
Mandur: That's a good point, I was wondering of having the cups server directly on the function, but that would pose a series of other issues.

docs/adr/006_cups.md
Mandur: Ok I understood after the discussion, the thumprints are in list to allow updates

Mandur (approved):

docs/adr/006_cups.md
danigian: Where in the spec is it saying it's DER?

docs/adr/006_cups.md
danigian: You are right, good catch

docs/adr/006_cups.md
bastbu: Yeah I thought as much. I think "we won't do this" is also a fair answer, will resolve the comment.

docs/adr/006_cups.md
bastbu: https://lora-developers.semtech.com/build/software/lora-basics/lora-basics-for-gateways/?url=cupsproto.html

> For client authentication using X509 certificates, all credentials are encoded in DER format:
trust is the certificate of the trusted certificate authority (CA)
cert is the personal certificate for the gateway
key is the personal private key for the gateway

docs/adr/006_cups.md
danigian: Both of you are right, rotation of root certificate should not happen frequently, therefore we can consider this an edge scenario.

That said, the idea would be the following:
- Another LoRaNetworkServer instance should be executed in parallel (with different PORT ideally) with the updated Server Certificate.
- Provisioning CLI tool should update all the concentrator devices with new CUPS URI and the new certificate bundles (which are now including the updated *.trust). The update process would be the same as it was a simple client certificate rotation process.
- As soon as a device is connecting to the "old" CUPS endpoint, a new uri and credentials bundle will be provided to the station.
- The Station will restart and try to use the "new" CUPS endpoint.
- Ideally after 24hrs, it's possible to remove the "old" LoRaNetworkServer instance

If you agree, I could include this scenario in the ADR or in the upcoming documentation.

@bastbu, @Mandur, any thoughts?

docs/adr/006_cups.md
danigian: https://lora-developers.semtech.com/build/software/lora-basics/lora-basics-for-gateways/?url=credentials.html

Here we have a clear indication of the fact that the *.trust is in PEM-encoded format.

Also having a look at the script in the [examples from BasicStation repo](https://github.com/lorabasics/basicstation/blob/master/examples/cups/prep.sh) we can see that only the signing keys for firmware updates are output as DER.

As agreed, let's keep this PEM 👍

Issue #836, opened 2021-11-17
E2E testing with multiple concentrators


roel4ez:
## Acceptance Criteria

- [x] #838
- [x] change E2E tests to OTA devices (joins) - depends on #830 
- [x] Test with local setup
- [ ] enable E2E tests in CI

MaggieSalak:
Enabling the test in the CI is tracked in https://github.com/Azure/iotedge-lorawan-starterkit/issues/871

PR #855, opened 2021-11-19
Multi-concentrator E2E test

MaggieSalak:
# PR for issue #836 

## What is being addressed

E2E test for multi-concentrator scenario using an OTAA device.

## How is this addressed

- new OTAA device defined
- new E2E test added in a new test class

## How was it tested

The test was successfully run by @danigian on a setup with 2 concentrators. Enabling it in the E2E CI would require adding another concentrator to the current setup - cc @Mandur for additional input here.


Tests/E2E/MultiConcentratorTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            await ArduinoDevice.setDeviceModeAsync(LoRaArduinoSerial._device_mode_t.LWOTAA);
+            await ArduinoDevice.setIdAsync(device.DevAddr, device.DeviceID, device.AppEUI);
+            await ArduinoDevice.setKeyAsync(device.NwkSKey, device.AppSKey, device.AppKey);
+
+            await ArduinoDevice.SetupLora(TestFixtureCi.Configuration);
+
+            var joinSucceeded = await ArduinoDevice.setOTAAJoinAsyncWithRetry(LoRaArduinoSerial._otaa_join_cmd_t.JOIN, 20000, 5);
+            Assert.True(joinSucceeded, "Join failed");
+
+            var logMsg = $"Duplicate message received from station with EUI";
```
spygi, line 37: Shall we make this a constant and re-use from `ConcentratorDeduplication` and here? This way if someone adjusts the log message the test won't fail unnecessarily.

MaggieSalak: Unfortunately we can't assert on the exact same message here as we don't know the station EUI. Removed the '$'

Mandur:
I just tested and my rak833 is toasted. We would need to get another one. Normaly we have several of these devices. Otherwise I also have another raspberry pi hat but that would pose some logistical issues on the CI setup

Mandur:
Question: Do you plan to enable this test on the CI later once we have the infra?

MaggieSalak:
> Question: Do you plan to enable this test on the CI later once we have the infra?

Yes, that's the goal once the infra allows it

Tests/E2E/MultiConcentratorTests.cs
MaggieSalak: Changed to use a constant

Mandur (approved):

Mandur:
Is this work tracked or should we create additional user stories?

MaggieSalak:
> Is this work tracked or should we create additional user stories?

It is tracked in this story #836. But maybe it makes sense to split it then since we can't enable the test in the CI right away.

PR #856, opened 2021-11-20
Updates to working with docs

roel4ez:
This PR updates the documentation on how to work with documentation.

### DIFF
--- a/docs/user-guide/documentation.md
+++ b/docs/user-guide/documentation.md
@@ -1,32 +1,64 @@
 # Working with the documentation
 
 This documentation site is built using [MkDocs](https://www.mkdocs.org/) and
-[mkdocs-material](https://squidfunk.github.io/mkdocs-material/). The static documentation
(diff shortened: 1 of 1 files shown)


docs/user-guide/documentation.md
```diff
@@ -1,32 +1,64 @@
...
 ```bash title="serve documentation locally"
-> docker pull squidfunk/mkdocs-material
+docker pull squidfunk/mkdocs-material
 
 # in the folder where the `docs/main` branch lives locally:
-> docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material
+docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material
 ```
 
 Now you can see the site running locally on <http://localhost:8080>. You can change
```
MaggieSalak, line 38: Not your change but it should be localhost:8000

roel4ez: updated local address

MaggieSalak (approved):

Issue #827, opened 2021-11-17
`LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests` test fails sometimes


atifaziz:
## Expected Behavior

Expected integration tests to pass on each run.

## Current Behavior

Instead the `LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.When_Same_Message_Comes_Multiple_Times_Result_Depends_On_Which_Concentrator_It_Was_Sent_From` test fails with the following details:

```
  Failed LoRaWan.Tests.Integration.LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.When_Same_Message_Comes_Multiple_Times_Result_Depends_On_Which_Concentrator_It_Was_Sent_From(isSameStation: False) [894 ms]
  Error Message:
   Assert.Equal() Failure
Expected: 1
Actual:   2
  Stack Trace:
     at LoRaWan.Tests.Integration.LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.When_Same_Message_Comes_Multiple_Times_Result_Depends_On_Which_Concentrator_It_Was_Sent_From(Boolean isSameStation) in /home/runner/work/iotedge-lorawan-starterkit/iotedge-lorawan-starterkit/Tests/Integration/LnsProtocolMessageProcessorConcentratorDeduplicationIntegrationTests.cs:line 109
--- End of stack trace from previous location where exception was thrown ---
```

## Steps to Reproduce

The test failure is sporadic and [occurs during some CI builds](https://github.com/Azure/iotedge-lorawan-starterkit/runs/4229340357?check_suite_focus=true#step:11:684) and goes away with re-runs.

## Logs

- 📎 [`logs_4021.zip`](https://github.com/Azure/iotedge-lorawan-starterkit/files/7552323/logs_4021.zip)

spygi:
Thanks for reporting this, dup of #824

PR #859, opened 2021-11-22
Ensures sockets are disposed properly.

spygi:
# PR for issue #827

## What is being addressed

This PR ensures that the web sockets we are opening are properly disposed at the end of the test.

The original issue #827  is about stabilizing the test in the CI, however I was not able to reproduce the failure anymore after multiple tries in dev. I also was not able to reproduce this locally running in a loop up to 50 test runs. Potentially the underlying issue was fixed already in #847 

## How is this addressed

By calling `Dispose()` on the sockets.


MaggieSalak (approved):

danigian (changes requested):
Why instead of setting the private fields, can't we just use the "using" keyword on lines 78/79?

danigian (approved):

Issue #771, opened 2021-11-10
Review error handling in TryGetDownstreamChannelFrequency for all regions


MaggieSalak:
## Description

Currently in the implementation in different regions we always return a bool indicating if calculation of downstream frequency succeeded. In certain cases (e.g. when upstream data rate or frequency is invalid for a given region) it might be better to throw an exception instead of returning `false`.

Issue #658, opened 2021-10-28
Validate Region.cs logic as pointed out in PR comments


danigian:
## Expected Behavior

There are some comments post-PR merge that would need a little bit of attention.
Please refer to comments in #656 about:

- [x] GetMaxPayloadSize and direct dereference of values from dictionary
- [x] Call to IsValidUpstreamFrequencyAndDataRate could be avoided in GetDownstreamDR
- [x] Logger levels
- [x] Casing of fields

PR #861, opened 2021-11-22
Region error handling

MaggieSalak:
# PR for issues #658 and #771

## What is being addressed

- Error handling in case upstream data rate or frequency is invalid
- Issues in Region pointed out in a closed PR #656 

## How is this addressed

- In case upstream data rate or frequency is invalid, we consistently throw a `LoRaProcessingException`. Before the exception was only thrown if data rate was invalid but for incorrect frequency we were returning false.
- Unit tests for region limits were updated as before in some cases they were using valid frequencies instead of ones outside of the region's limits.
- Small issues pointed in a closed PR #656 were fixed (e.g. missing comment, rx1DrOffset validation).


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
```diff
@@ -119,10 +120,12 @@ public override bool TryGetJoinChannelIndex(double frequency, out int channelInd
         [Obsolete("#655 - This Rxpk based implementation will go away as soon as the complete LNS implementation is done.")]
         public override bool TryGetDownstreamChannelFrequency(Rxpk upstreamChannel, out double frequency, DeviceJoinInfo deviceJoinInfo)
         {
-            frequency = 0;
+            if (deviceJoinInfo is null) throw new ArgumentNullException(nameof(deviceJoinInfo));
```
Mandur, line 123: Wouldn't devicejoininfo be null in non china regions?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
```diff
@@ -171,15 +171,17 @@ public override bool TryGetDownstreamChannelFrequency(Rxpk upstreamChannel, out
         /// <summary>
         /// Logic to get the correct downstream transmission frequency for region CN470.
         /// <param name="upstreamFrequency">The frequency at which the message was transmitted.</param>
-        /// <param name="dataRate">The upstream data rate.</param>
+        /// <param name="upstreamDataRate">The upstream data rate.</param>
         /// <param name="deviceJoinInfo">Join info for the device, if applicable.</param>
         /// </summary>
-        public override bool TryGetDownstreamChannelFrequency(double upstreamFrequency, ushort dataRate, out double downstreamFrequency, DeviceJoinInfo deviceJoinInfo)
+        public override bool TryGetDownstreamChannelFrequency(double upstreamFrequency, out double downstreamFrequency, ushort? upstreamDataRate = null, DeviceJoinInfo deviceJoinInfo = default)
```
Mandur, line 177: How come the upstream datarate could be nullable, I don't see it being used in this method

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionEU868.cs
```diff
@@ -68,36 +69,28 @@ public override bool TryGetDownstreamChannelFrequency(Rxpk upstreamChannel, out
...
 
         /// <summary>
         /// Logic to get the correct downstream transmission frequency for region EU868.
         /// </summary>
         /// <param name="upstreamFrequency">Frequency on which the message was transmitted.</param>
-        /// <param name="dataRate">Data rate at which the message was transmitted.</param>
+        /// <param name="upstreamDataRate">Data rate at which the message was transmitted.</param>
         /// <param name="deviceJoinInfo">Join info for the device, if applicable.</param>
-        public override bool TryGetDownstreamChannelFrequency(double upstreamFrequency, ushort dataRate, out double downstreamFrequency, DeviceJoinInfo deviceJoinInfo = null)
+        public override bool TryGetDownstreamChannelFrequency(double upstreamFrequency, out double downstreamFrequency, ushort? upstreamDataRate = null, DeviceJoinInfo deviceJoinInfo = null)
```
Mandur, line 86: Same here I don't see why upstream datarate is required

Tests/Unit/LoRaTools/RegionAS923TestData.cs
```diff
@@ -68,9 +68,9 @@ public static IEnumerable<object[]> TestRegionFrequencyData()
                new object[] { region, 900, 8 },
                new object[] { region, 914.5, 9 },
                new object[] { region, 930, 10 },
-               new object[] { region, 923.4, 18 },
-               new object[] { region, 925.5, 90 },
-               new object[] { region, 923.2, 100 },
+               new object[] { region, 928.4, 18 },
```
Mandur, line 71: Can you elaborate on why we change the test cases?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
MaggieSalak: Correct but this is in CN Region and here `deviceJoinInfo` is required, that's why I added the validation.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
MaggieSalak: It is used only in the US region. The downstream frequency in US depends on the data rate. In other regions it's not used. Do you have another suggestion how to do this better?

Tests/Unit/LoRaTools/RegionAS923TestData.cs
MaggieSalak: Before some of the test cases were actually using valid frequency for the region (e.g. AS923 has range 915 - 928 MHz) so e.g. 923.4 as we had before is a valid frequency. They were passing because previously we were checking dr and frequency together is `IsDataRateAndFrequencyValid()` so this would return false if only data rate was out of allowed range but not the frequency.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionEU868.cs
MaggieSalak: Replied in the other comment.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
Mandur: forgot about this ;)

Mandur (approved):

Issue #747, opened 2021-11-09
[Story] Raise custom metrics and alerts


bastbu:
## Description

We have a list of custom metrics/alerts that we want to raise, defined as part of #468. We should consider the .NET 6 Metrics API as an abstraction for our custom metrics, that can be exported to the Prometheus endpoint and to Application Insights.

## Acceptance criteria

- [ ] Custom metrics are exported to Application Insights
- [ ] Custom metrics are made available on Prometheus endpoint
~~- [ ] Alerts based on custom metrics are deployed when user is opted in to Application Insights~~

bastbu:
Alerts will be extracted from this story and handled by @eedorenko

bastbu:
Extracting the alerts part to #891

PR #887, opened 2021-11-23
Feature/747 custom metrics exporters

bastbu:
# PR for issue #747

## What is being addressed

With this PR we add support to raise custom metrics via the `System.Diagnostics.Metrics` API and export them to either Application Insights or on the Prometheus endpoint. We only add support for non-observable metrics (`Histogram<T>` and `Counter<T>`).

## How is this addressed

Since for both Application Insights and the prometheus-net SDK custom dimensions/tags need to be defined per metric, we introduce a registry of Custom Metrics that we track. We can only determine at runtime whether the metric raised has the correct custom dimensions/tags.

An example metric for the number of `JoinRequests` is added to the solution to illustrate how the registry can be used to raise metrics.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -190,6 +195,7 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
                     break;
                 case LnsMessageType.JoinRequest:
                     LogReceivedMessage(this.logger, "jreq", json, null);
+                    JoinRequestCounter.Add(1, new[] { KeyValuePair.Create(MetricRegistry.GatewayIdTagName, (object)stationEui) });
```
danigian, line 198: We might want to track this metric later, after deduplication

bastbu: Addressed this in 98676615807f87182133e578e24f20c3ac20cd46

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -199,6 +204,8 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
                             break;
                         }
 
+                        JoinRequestCounter.Add(1, new[] { KeyValuePair.Create(MetricRegistry.GatewayIdTagName, (object)stationEui) });
```
Mandur, line 207: Should we track the join request after the mic was successfully checked? At this stage could still be from another network?

danigian (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
bastbu: Addressed this in 07808a34fe97b4981c5d12ca4eeeb025a077d4af

Issue #858, opened 2021-11-22
Basic station docker image should permit optional setting of the stationEUI


Mandur:
## Description

Basic station docker image should permit optional setting of the stationEUI

PR #890, opened 2021-11-24
Fix station EUI

Mandur:
# PR for issue #858

## What is being addressed

Enable basics station to have its station EUI statically set

### DIFF
--- a/LoRaEngine/modules/LoRaBasicsStationModule/station.conf
+++ b/LoRaEngine/modules/LoRaBasicsStationModule/station.conf
@@ -20,8 +20,7 @@
         /* chan_multiSF_X, chan_Lora_std, chan_FSK provided by LNS */
     },
     "station_conf": {
-        "placeholder": "placeholder",
-        "logfile_size": "10mb",
+        "routerIdPlaceholder": "routerIdPlaceholder",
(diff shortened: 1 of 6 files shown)


LoRaEngine/modules/LoRaBasicsStationModule/station.conf
```diff
@@ -20,6 +20,8 @@
         /* chan_multiSF_X, chan_Lora_std, chan_FSK provided by LNS */
     },
     "station_conf": {
+        "placeholder": "placeholder",
```
danigian, line 23: for future extensibility, should we rename this placeholder something like routerIdPlaceholder?

LoRaEngine/modules/LoRaBasicsStationModule/station.conf
```diff
@@ -20,6 +20,8 @@
...
+        "logfile_size": "10mb",
```
danigian, line 24: was this new addition on purpose?

Mandur: Nope will remove thanks for the the catch

danigian (approved):

Issue #893, opened 2021-11-24
Nested lists are not properly rendered with mkdocs


danigian:
Nested lists are not properly rendered with mkdocs. See more details in 'mdx_truly_sane_lists' official [github repo](https://github.com/radude/mdx_truly_sane_lists)

PR #894, opened 2021-11-24
Fixing rendering of lists with mdx_truly_sane_lists

danigian:
# PR for issue #893 

## What is being addressed

Fixing rendering of nested lists with mdx_truly_sane_lists


danigian:
Deleted in favor of #895

PR #904, opened 2021-11-25
Feature/568 cache redesign

Mandur:
# PR for issue #568

**All this work, including the below post was made by @p-schuler and @atifaziz** 

## What is being addressed

This PR is a redesign of the local device cache as well as some fixes that are affected by those changes. 

The main issues that are addressed are:
- leaking device connections
- non deterministic device disposal
- single instance LoRaDevice management
- lack of unit testing
- IsOurDevice update logic not centralized

## How is this addressed

The `LoRaDevice` instances were cached in the `LoRaDeviceRegistry`. The logic for the cache was spread in that class. Multiple contributors for the cache would call through callbacks to update the cache inside of the `LoRaDeviceRegistry`. One issue was that if a device got evited from the cache, it was not disposed. The connection would stay with the connection manager. On the next creation of the device, the connection would simply get overwritten with the new instance and the old connection would leak. The current designed tried to decouple the instances of the `LoRaDevice` from the connections. This made the removal of the connection very hard, as the "life" device instances were not tracked. 

This change introduces a class that manages cached devices - `LoRaDeviceCache`. The service can be injected into other services for easy re-use, the logic is encapsulated and unit tested. The `LoRaDevice` instances are no longer removed after the fixed 2 days period. Instead, there is logic to update the existing cache, as long as devices are active (send messages). If a device stops sending messages, the device is automatically removed and the connection is closed.

The configuration can be controlled through the `LoRaDeviceCacheOptions`:

- `ValidationInterval`: How often the validation of the device cache is run. The local device cache will be evaluated after the specified time and all devices that need to be removed or refreshed are processed. 
- `MaxUnobservedLifetime`: The time we allow a device to be not sending any messages, before we consider it stale and remove it from the cache. This should be a high value, as LoRa sensors are optimized for saving power and might only send a message, once a week for example.
- `RefreshInterval`: If a device is in cache and it exeeds this time since the last refresh, a twin refresh is triggered and the cached data is updated.

The `LoRaDevice` is a single instance / DevEUI in this new implementation. All creations go through the factory, which is also caching the instance. 

The relation between `LoRaDeviceClientConnectionManager` and `LoRaDevice` is 1:1. This would allow us to mange the connection directly from the `LoRaDevice`. #911 was added to track this as a separate item.

There is no guarantee that you can't create another `LoRaDevice` instance or connection. You can do this bypassing the factory. To avoid this we would need to move the factory logic to a factory in the `LoRaDevice` and make the `ctor` private.

- `IsOurDevice`: this was calculated at multiple stages. The logic was duplicated in all those places. We only need to evaluate this property once, when we load the twins (or refresh them). The logic is wrapped in the `LoRaDevice` and a caller can ask to refresh the information. It's set when creating a device (we have the information from the twins in the device info) and when loading the twins.
- There was a small glitch of repeated logic when creating the `TimeWatcher` in the `JoinRequestMessageHandler`. The logic is now re-used.

There are a few follow up that still need to be addressed. We explicitly moved them to different stories to keep this as contained as possible:

- [ ] #911
- [ ] #912
- [ ] #913
- [ ] #927
- [ ] #934
- [ ] #941
- [ ] #960


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
```diff
@@ -347,13 +347,13 @@ public record LoRaDeviceCacheOptions
         /// as LoRa sensors are optimized for saving power and
         /// might only send a message, once a week for example.
         /// </summary>
-        public TimeSpan MaxUnobservedLifetime { get; set; }
```
p-schuler: This was on purpose not immutable. The goal is the same as with the logger to be able to drive these values through twins and have them updated while running. To do this, they need to be mutable.

atifaziz: Fair enough but then what was the thinking behind making the options a record?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
```diff
@@ -0,0 +1,352 @@
...
+                    catch (LoRaProcessingException ex)
+                    {
+                        // retry on next iteration
+                        this.logger.LogError(ex, "Failed to refresh device.");
+                    }
+                }
+            }
+        }
+
+        protected virtual Task RefreshDeviceAsync(LoRaDevice device, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
Mandur: 100/sec as per https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling, could this be a problem? I think it just throttle right? It does not error out?

Mandur (commented):
Publishing some few things

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -40,61 +41,39 @@ internal enum LoaderState
...
-            this.configuration = configuration;
-            this.registerDeviceAction = registerDeviceAction;
             this.logger = logger;
             this.state = LoaderState.QueryingDevices;
-            this.loadingDevicesFailed = false;
             this.queueLock = new object();
             this.queuedRequests = new List<LoRaRequest>();
-            _ = TaskUtil.RunOnThreadPool(async () =>
-            {
-                using var scope = this.logger.BeginDeviceAddressScope(this.devAddr);
```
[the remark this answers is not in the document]
Mandur: As far as I could see it is actually coming from the main thread

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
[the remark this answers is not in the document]
Mandur: actually, those requests will hit the function cache, so that shouldn't be an issue

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -146,60 +118,70 @@ private async Task Load()
             }
         }
 
-        private async Task<List<LoRaDevice>> CreateDevicesAsync(IReadOnlyList<IoTHubDeviceInfo> devices)
+        protected async Task CreateDevicesAsync(IReadOnlyList<IoTHubDeviceInfo> devices)
```
Mandur, line 120: Question what would happen if

We have device A and B sharing the same devaddress, A is loaded on gateway 1. Now B starts sending messages and Gateway 2 wins the race, now the two gateways have both device in cache

Mandur: If I understood correctly the connection is opened only at message send so this should not be a problem, but is the twin get not potentially conflicting?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -146,60 +118,70 @@ private async Task Load()
...
                 {
-                    if (deviceTask.IsCompletedSuccessfully)
+                    foreach (var deviceTask in initTasks)
                     {
-                        var device = await deviceTask;
-
-                        if (device != null)
+                        if (deviceTask.IsCompletedSuccessfully)
                         {
+                            var device = await deviceTask;
```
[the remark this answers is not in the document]
p-schuler: not sure what you mean? Here we only handle the success case (it's explicitly validated through the IsCompletedSuccessfully) - we get the device through the await deviceTask; What do we need to raise?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -146,60 +118,70 @@ private async Task Load()
...
                     {
-                        using var scope = this.logger.BeginDeviceScope(foundDevice.DevEUI);
-                        var loRaDevice = this.deviceFactory.Create(foundDevice);
-                        initTasks.Add(InitializeDeviceAsync(loRaDevice));
+                        if (string.IsNullOrEmpty(cachedDevice.DevAddr))
+                        {
+                            // device in cache from a previous join that we didn't complete
+                            // (lost race with another gw) - refresh the twins now and keep it
+                            // in the cache
+                            refreshTasks ??= new List<Task<bool>>();
```
Mandur, line 146: Can refreshtask be anything other than null here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -223,199 +232,229 @@ internal LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClientConnectionMa
...
+            }
+
+            if (twin.Properties.Reported.Contains(TwinProperty.LastProcessingStationEui))
+            {
+                var stationEui = StationEui.Parse(twin.Properties.Reported[TwinProperty.LastProcessingStationEui].Value as string);
+                lastProcessingStationEui = new ChangeTrackingProperty<StationEui>(TwinProperty.LastProcessingStationEui, stationEui);
             }
 
-            return false;
+            LastUpdate = DateTimeOffset.UtcNow;
```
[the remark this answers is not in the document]
p-schuler: No. This represents the last twin update and not when we saw the device on the network the last time.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaRequest.cs
```diff
@@ -45,7 +45,7 @@ protected LoRaRequest(LoRaPayload payload)
 
         internal void NotifyFailed(LoRaDevice loRaDevice, LoRaDeviceRequestFailedReason reason, Exception exception = null) => NotifyFailed(loRaDevice?.DevEUI, reason, exception);
 
-        internal void NotifyFailed(LoRaDeviceRequestFailedReason reason, Exception exception = null) => NotifyFailed((string)null, reason, exception);
+        internal virtual void NotifyFailed(LoRaDeviceRequestFailedReason reason, Exception exception = null) => NotifyFailed((string)null, reason, exception);
```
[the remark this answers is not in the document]
p-schuler: yes

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
```diff
@@ -0,0 +1,352 @@
...
+        {
+            _ = device ?? throw new ArgumentNullException(nameof(device));
+            return device.InitializeAsync(this.configuration, cancellationToken);
+        }
+
+        public bool Remove(string devEui)
+        {
+            lock (this.syncLock)
```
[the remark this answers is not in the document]
p-schuler: Absolutely! Thanks for catching. Removed with 6f60c5b36a7310f6c559130c19e4608ddaa94a7f

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -146,60 +118,70 @@ private async Task Load()
...
+                    if (!this.loraDeviceCache.TryGetByDevEui(foundDevice.DevEUI, out var cachedDevice))
+                    {
+                        initTasks.Add(this.deviceFactory.CreateAndRegisterAsync(foundDevice, CancellationToken.None));
+                    }
+                    else
                     {
-                        using var scope = this.logger.BeginDeviceScope(foundDevice.DevEUI);
-                        var loRaDevice = this.deviceFactory.Create(foundDevice);
-                        initTasks.Add(InitializeDeviceAsync(loRaDevice));
+                        if (string.IsNullOrEmpty(cachedDevice.DevAddr))
```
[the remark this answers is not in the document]
p-schuler: not necessarily. I tried to explain it with the comment:

// device in cache from a previous join that we didn't complete
// (lost race with another gw) - refresh the twins now and keep it
// in the cache

So it is the case, where 2 gw get the join, both load them but only one generates the session keys. They are stored in the twins, and if we get a message, we want to refresh from the twins, to ensure our cached device does have the session keys.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -146,60 +117,70 @@ private async Task Load()
...
-                        var device = await deviceTask;
-
-                        if (device != null)
+                        if (deviceTask.IsCompletedSuccessfully)
                         {
+                            var device = await deviceTask;
+                            // run initializers
+                            InitializeDevice(device);
                             createdDevices.Add(device);
                         }
```
Mandur, line 178: It looks like this createddevice list is only used to count the number of device created or am I missing something? Couldn't we replace this by a counter if yes?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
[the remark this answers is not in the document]
Mandur: I was thinking of a log describing the siutation, but maybe you raise it somewhere else and I missed it

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinDeviceLoader.cs
```diff
@@ -3,56 +3,71 @@
...
-            if (await loRaDevice.InitializeAsync())
+            await this.joinLock.WaitAsync();
+            try
             {
-                return loRaDevice;
+                if (this.deviceCache.TryGetByDevEui(this.ioTHubDevice.DevEUI, out var cachedDevice))
+                {
+                    return cachedDevice;
+                }
+                return await this.deviceFactory.CreateAndRegisterAsync(this.ioTHubDevice, CancellationToken.None);
```
Mandur, line 44: I have a question regarding join requests let's discuss

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
```diff
@@ -0,0 +1,344 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public class LoRaDeviceCache : IDisposable
+    {
+        private readonly LoRaDeviceCacheOptions options;
+        private readonly ConcurrentDictionary<string, ConcurrentDictionary<string, LoRaDevice>> devAddrCache = new();
```
Mandur, line 22: Out of interrest, as we are always locking, what is the advantage of using a concurrent dict here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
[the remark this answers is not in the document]
Mandur: This comment was supposed to be on 166 initially.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
Mandur: It's a loop :D

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinDeviceLoader.cs
Mandur: We discussed, we will add a clear cache on deveui when a join request is received

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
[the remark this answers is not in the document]
p-schuler: There is logging happening when we set the flag (see the exception handler above).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
p-schuler: That is true. I consider this a small change, and include it in this scope.

p-schuler: 941095378e1afdccfa4f7d88e44d45a739ee7376

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinDeviceLoader.cs
p-schuler: tracked in #1024

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
p-schuler: We do allow concurrent access when refreshing the cache. Hence, we need it there.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -95,6 +95,8 @@ public void ConfigureServices(IServiceCollection services)
                         .AddSingleton<LoRaDeviceAPIServiceBase, LoRaDeviceAPIService>()
                         .AddSingleton<WebSocketWriterRegistry<StationEui, string>>()
                         .AddSingleton<IPacketForwarder, DownstreamSender>()
+                        .AddSingleton<LoRaDeviceCache>()
+                        .AddSingleton(x => new LoRaDeviceCacheOptions { MaxUnobservedLifetime = TimeSpan.FromDays(10), RefreshInterval = TimeSpan.FromDays(2), ValidationInterval = TimeSpan.FromMinutes(10) })
```
danigian, line 99: Can't we just:

```suggestion
                        .AddSingleton(new LoRaDeviceCacheOptions { MaxUnobservedLifetime = TimeSpan.FromDays(10), RefreshInterval = TimeSpan.FromDays(2), ValidationInterval = TimeSpan.FromMinutes(10) })
```
?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -223,199 +232,229 @@ internal LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClientConnectionMa
...
+            if (twin.Properties.Desired.Contains(TwinProperty.Deduplication))
+            {
+                var val = twin.Properties.Desired[TwinProperty.Deduplication].Value as string;
+                _ = Enum.TryParse<DeduplicationMode>(val, true, out var mode);
+                Deduplication = mode;
+            }
+
+            if (twin.Properties.Desired.Contains(TwinProperty.ClassType))
+            {
+                if (string.Equals("c", (string)twin.Properties.Desired[TwinProperty.ClassType], StringComparison.OrdinalIgnoreCase))
```
danigian, line 390: nit: Can we combine the two "if" in one?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -223,199 +232,229 @@ internal LoRaDevice(string devAddr, string devEUI, ILoRaDeviceClientConnectionMa
...
+            }
+
+            if (twin.Properties.Reported.Contains(TwinProperty.Region))
+            {
+                var regionValue = twin.Properties.Reported[TwinProperty.Region].Value as string;
+                if (Enum.TryParse<LoRaRegionType>(regionValue, true, out var loRaRegion))
                 {
-                    var value = GetTwinPropertyIntValue(twin.Properties.Desired[TwinProperty.KeepAliveTimeout].Value);
-                    if (value > 0)
+                    if (Enum.IsDefined(typeof(LoRaRegionType), loRaRegion))
```
danigian, line 406: Is this if requirement? TryParse cannot be true and, at the same time, the outcome loraRegion is not a LoRaRegionType, no?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
```diff
@@ -0,0 +1,344 @@
...
+        private readonly object syncLock = new object();
+        private readonly NetworkServerConfiguration configuration;
+        private readonly ILogger<LoRaDeviceCache> logger;
+        private CancellationTokenSource? ctsDispose;
+        private readonly StatisticsTracker statisticsTracker = new StatisticsTracker();
+
+        protected LoRaDeviceCache(LoRaDeviceCacheOptions options, NetworkServerConfiguration configuration, ILogger<LoRaDeviceCache> logger, CancellationToken externalRefreshCancellationToken)
+        {
+            this.options = options;
+            this.ctsDispose = externalRefreshCancellationToken.CanBeCanceled ? CancellationTokenSource.CreateLinkedTokenSource(externalRefreshCancellationToken) : new CancellationTokenSource();
```
danigian, line 33: What's the purpose of creating a linked token source when the only user of this protected constructor is actually always passing a CancellationToken.None?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
p-schuler: I know that those are all flagged by the tool now because the indention changed. I didn't plan to do a complete refactoring of all the code in the LoRaDevice as part of that change. There is going to be a push to get rid of all the dynamic type handling in there, we can do some more cleanup then.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
p-schuler: Totally.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
p-schuler: Impossible. But that's not my change. :) I should have pointed out that this file is tricky to review, as I removed one level of indention for readability, but that caused the diff tool to mark a lot of existing code as being changed. Maybe if you ignore white spaces, it would be a bit better. Anyway, there are quite a few improvements we can do in the parsing logic of the LoRaDevice. We have an item open to fix dynamic type casting in there and we can consider some fixes there.

danigian (approved):

p-schuler:
The multi concentrator E2E tests fail due to the fact that we only have 1 concentrator per LNS running atm. Closing this PR now due to others depending on it. We will validate the test once the concentrator setup is fixed.

PR #908, opened 2021-11-25
Feature/873 client certificate validation

danigian:
# PR for issue #873 

## What is being addressed

Implementing the logic to verify client certificate at their connection

## How is this addressed

- [x] BasicsStationConfigurationService refactored to cache entire Twin and expanded to get a list of allowed client certificate thumbprints from Twin
- [x] NetworkServerConfiguration expanded for allowing a 'CLIENT_CERTIFICATE_MODE' parameter
- [x] BasicsStationNetworkServer to conditionally allow/require client certificate (based on configuration) and use ClientCertificateValidatorService for validating validity
- [x] Additional Unit tests


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                this.logger.LogError("{Class}: Could not find a possible StationEui in '{CommonName}'.", nameof(ClientCertificateValidatorService), commonName);
+                return false;
+            }
+
+            // Logging any chain related issue, but not failing on it.
+            foreach (var status in chain.ChainStatus)
+            {
+                using var scope = this.logger.BeginEuiScope(stationEui);
+                this.logger.LogDebug("{Class}: {Status} {StatusInformation}", nameof(ClientCertificateValidatorService), status.Status, status.StatusInformation);
```
bastbu, line 54: Shouldn't we leave the `{Class}` prefix up to the logger to decide whether it should be included or not? Since we provided the generic type argument to the `ILogger<T>`, the loggers should pick up the correct category names.

Tests/Unit/NetworkServer/BasicsStation/BasicsStationNetworkServerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        private string? notProtectedPfx;
+
+        [Theory]
+        [InlineData(true, ClientCertificateMode.NoCertificate)]
+        [InlineData(true, ClientCertificateMode.AllowCertificate)]
+        [InlineData(true, ClientCertificateMode.RequireCertificate)]
+        [InlineData(false, ClientCertificateMode.NoCertificate)]
+        [InlineData(false, ClientCertificateMode.AllowCertificate)]
+        [InlineData(false, ClientCertificateMode.RequireCertificate)]
+        public void GivenCertificateConfiguration_ConfigureHttpsSettings_Succeeds(bool serverPfxPasswordProtected, ClientCertificateMode clientCertificateMode)
```
bastbu, line 33: Will there be integration tests/E2E tests for this scenario as well?

Tests/Unit/NetworkServer/BasicsStation/BasicsStationNetworkServerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                Assert.NotNull(httpsConnectionAdapterOptions.ClientCertificateValidation);
+            } else
+            {
+                Assert.Null(httpsConnectionAdapterOptions.ClientCertificateValidation);
+            }
+
+        }
+
+        internal static async Task<string> CreatePfxCertificate(bool passwordProtected, string stationEui = "001122FFFEAABBCC")
```
bastbu, line 67: [nit] should have suffix `Async`

Tests/Unit/NetworkServer/BasicsStation/ClientCertificateValidatorServiceTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            File.Delete(this.clientCertificateWithEuiPath);
+            File.Delete(this.clientCertificateWithInvalidEuiPath);
+            return Task.CompletedTask;
+        }
+
+        public async Task InitializeAsync()
+        {
+            this.clientCertificateWithEuiPath = await BasicsStationNetworkServerTests.CreatePfxCertificate(false);
+            this.clientCertificateWithInvalidEuiPath = await BasicsStationNetworkServerTests.CreatePfxCertificate(false, InvalidStationEui);
```
bastbu, line 118: Why create this file for every test when we only need it in one test?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            this.stationConfigurationService = stationConfigurationService;
+            this.logger = logger;
+        }
+
+        // Following synchronous implementation is needed for ASP.NET Core ClientCertificateValidation
+        public virtual bool Validate(X509Certificate2 certificate, X509Chain? chain, SslPolicyErrors sslPolicyErrors) =>
+            ValidateAsync(certificate, chain, sslPolicyErrors, default).GetAwaiter().GetResult();
+
+        public virtual async Task<bool> ValidateAsync(X509Certificate2 certificate, X509Chain? chain, SslPolicyErrors sslPolicyErrors, CancellationToken token)
```
bastbu, line 33: A very open-ended question that I think will be resolved at the end: we don't expect this to happen often, right? So, async validation that we will block on the call-site is not that big of an issue, right? Do we have any alternatives that we considered, though?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -86,6 +86,7 @@ public void ConfigureServices(IServiceCollection services)
                         .AddSingleton<LoRaDeviceAPIServiceBase, LoRaDeviceAPIService>()
                         .AddSingleton<WebSocketWriterRegistry<StationEui, string>>()
                         .AddSingleton<IPacketForwarder, DownstreamSender>()
+                        .AddSingleton<ClientCertificateValidatorService>()
```
bastbu, line 89: Do we want to register this unconditionally? If we don't have client certs validation enabled, we don't need it, no?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (certificate is null) throw new ArgumentNullException(nameof(certificate));
+            if (chain is null) throw new ArgumentNullException(nameof(chain));
+
+            var commonName = certificate.GetNameInfo(X509NameType.SimpleName, false);
+            StationEui stationEui;
+            if (Regex.Match(commonName, "([a-fA-F0-9]{2}[-:]?){8}") is { } regex && regex.Success)
+            {
+                stationEui = StationEui.Parse(regex.Value);
```
bastbu, line 42: It would be nice to handle this exception here (or use `TryParse` instead) and go through the `else` path as well with the same logging message. We can already anticipate that an exception might be thrown here if the EUI is not valid.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public ClientCertificateValidatorService(IBasicsStationConfigurationService stationConfigurationService,
+                                                 ILogger<ClientCertificateValidatorService> logger)
+        {
+            this.stationConfigurationService = stationConfigurationService;
+            this.logger = logger;
+        }
+
+        // Following synchronous implementation is needed for ASP.NET Core ClientCertificateValidation
+        public virtual bool Validate(X509Certificate2 certificate, X509Chain? chain, SslPolicyErrors sslPolicyErrors) =>
```
bastbu, line 30: [nit] I would argue that we can do this on the call site instead of introducing an overload for this. This is not something we should "encourage" to reuse in other places

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
```diff
@@ -61,28 +66,39 @@ public async Task<string> GetRouterConfigMessageAsync(StationEui stationEui, Can
...
+        }
 
+        public async Task<string[]> GetAllowedClientThumbprintsAsync(StationEui stationEui, CancellationToken cancellationToken)
+        {
+            var twin = await GetTwinAsync(stationEui, cancellationToken);
             try
             {
-                var configJson = ((object)twin.Properties.Desired[RouterConfigPropertyName]).ToString();
-                return LnsStationConfiguration.GetConfiguration(configJson);
+                var thumbprintsArrayJson = (string)twin.Properties.Desired[ClientThumbprintPropertyName].ToString();
```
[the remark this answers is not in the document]
danigian: I can't see any warning from Code QL, but I can definitely do what you suggest

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServer.cs
```diff
@@ -66,5 +60,23 @@ public static async Task RunServerAsync(NetworkServerConfiguration configuration
...
+        internal static void ConfigureHttpsSettings(NetworkServerConfiguration configuration,
+                                                    IServiceProvider serviceProvider,
+                                                    HttpsConnectionAdapterOptions https)
+        {
+            https.ServerCertificate = string.IsNullOrEmpty(configuration.LnsServerPfxPassword) ? new X509Certificate2(configuration.LnsServerPfxPath)
+                                                                                               : new X509Certificate2(configuration.LnsServerPfxPath,
+                                                                                                                      configuration.LnsServerPfxPassword,
+                                                                                                                      X509KeyStorageFlags.DefaultKeySet);
+
+            if (Enum.TryParse<ClientCertificateMode>(configuration.ClientCertificateMode, out var requiredClientCertificateMode)
```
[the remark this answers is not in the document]
danigian: Moved to config part, with a Parse and not a TryParse

bastbu (approved):
Tested on my machine, looks good to me. Let's make sure that we do not forget to add the documentation later on, it's quite important to know that the client cert needs to have a specific common name.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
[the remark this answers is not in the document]
danigian: You are right, it should not be null

Tests/Unit/NetworkServer/BasicsStation/BasicsStationNetworkServerTests.cs
danigian: There will be some E2E in the end

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
danigian: Unfortunately there are no other alternatives, since the Client Certificate validation is happening synchronously on ASP.NET side. We tried to find alternatives with Atif, but we concluded that it's just fine to do this check synchronously for now

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
```diff
@@ -32,27 +35,29 @@ internal sealed class BasicsStationConfigurationService : IBasicsStationConfigur
...
-
-        public async Task<string> GetRouterConfigMessageAsync(StationEui stationEui, CancellationToken cancellationToken)
+        private async Task<Twin> GetTwinAsync(StationEui stationEui, CancellationToken cancellationToken)
         {
-            var cacheKey = $"{CachePrefixName}{stationEui}";
+            var cacheKey = $"{ConcentratorTwinCachePrefixName}{stationEui}";
 
             if (this.cache.TryGetValue(cacheKey, out var result))
-                return (string)result;
+                return (Twin)result;
```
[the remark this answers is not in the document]
danigian: Now storing only desired properties

Issue #876, opened 2021-11-23
Modify device provisioning CLI to support certificate rotation


danigian:
The device provisioning CLI should support certificate rotation:
- we support uploading a new certificate bundle
- we support updating the credential thumbprints
- we support removing credential thumbprints (revoke)

PR #909, opened 2021-11-25
Feature/876 certificate rotation cli

bastbu:
# PR for issue #876

## What is being addressed

With this PR we introduce support for certificate rotation. This includes uploading new certificates, and revoking old client thumbprints

## How is this addressed

We add two new verbs to the CLI. One can be used to upload a new client certificate, while the other can be used to revoke an accepted thumbprint.

### DIFF
--- a/Tools/Cli-LoRa-Device-Provisioning/Program.cs
+++ b/Tools/Cli-LoRa-Device-Provisioning/Program.cs
@@ -174,11 +174,19 @@ namespace LoRaWan.Tools.CLI
                     return false;
                 }
 
-                return await UploadCertificateBundleAsync(opts.CertificateBundleLocation, opts.StationEui, async (crcHash, bundleStorageUri) =>
+                if (opts.NoCups)
                 {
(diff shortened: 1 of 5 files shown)


Tools/Cli-LoRa-Device-Provisioning/Program.cs
```diff
@@ -223,6 +202,80 @@ private static async Task<bool> RunAddAndReturnExitCode(AddOptions opts)
...
+            if (twin is null)
+            {
+                StatusConsole.WriteLogLine(MessageType.Error, "Device was not found in IoT Hub. Please create it first.");
+                return false;
+            }
+
+            var twinJObject = JsonConvert.DeserializeObject<JObject>(twin.Properties.Desired.ToJson());
+            var clientThumprints = twinJObject[TwinProperty.ClientThumbprint];
+            var t = clientThumprints.FirstOrDefault(t => t.ToString().Equals(opts.ClientCertificateThumbprint, StringComparison.OrdinalIgnoreCase));
+            t?.Remove();
```
danigian, line 274: I think we should log an error if thumbprint is not found

danigian (approved):

Issue #915, opened 2021-11-26
Adjust CUPS ADR with Facade Function related fixes


danigian:
- Call should be a GET and not a POST
- We should specify in ADR that for first version we are using the Storage only and its connection string (instead of MSI)
- Request should only specify cups or lns, response should not return a json but can return a plaintext with the base64 encoded bundle

PR #917, opened 2021-11-26
Updating CUPS ADR

danigian:
# PR for issue #915 

## What is being addressed

This PR is addressing changes described in #915 
These are following the work in #916

### DIFF
--- a/docs/adr/006_cups.md
+++ b/docs/adr/006_cups.md
@@ -6,7 +6,7 @@
 
 **Author**: Daniele Antonio Maggio
 
-**Status**: Proposed
+**Status**: Accepted
 
 __________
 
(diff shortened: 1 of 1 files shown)


docs/adr/006_cups.md
```diff
@@ -221,12 +221,14 @@ else *CredentialUrl is plain Url
...
 end
 Function->>LoRaWanNetworkSrvModule: Return credential blob
-Note right of LoRaWanNetworkSrvModule: Response should be a json including base64 encoded credential
+Note right of LoRaWanNetworkSrvModule: Response should be a base64 encoded credential sent as plaintext
 
 ```
 
 In addition to this, the Azure Function must be able to properly authenticate to target sink via Managed Identity (when using Azure Blob Storage or KeyVault).
 
+For the very first version of the implementation, only Azure Storage Account should be implemented and only connection string authentication is going to be supported.
```
Mandur, line 230: Connection string for auth Function-> storage right? 
Would it be really harder to use MI straight away?

Mandur (approved):

docs/adr/006_cups.md
danigian: Not harder, but requires more changes in the templates

Issue #902, opened 2021-11-25
Remove LoRaWanPktFwdModule from solution


danigian:
We should remove LoRaWanPktFwdModule from solution

PR #919, opened 2021-11-27
Remove Packet forwarder

Mandur:
# PR for issue #902

Remove pkt fwd support


atifaziz (commented):
There is still the `Docs/LoRaWanPktFwdRAK833USB` submodule that was added in #45. Shouldn't that be removed as part of this?

Mandur:
I wasn't sure about this one, Should we add then daniele repo as submoduble?

atifaziz:
> Should we add then daniele repo as submoduble?

Sorry, what's the Daniele repo?

danigian (approved):

Issue #914, opened 2021-11-26
Integration tests against redis should be part of the same collection


bastbu:
## Expected Behavior

Integration tests using the Redis instance should not run in parallel. 

## Current Behavior

Tests using the `RedisFixture` should not run at the same time. Some test classes will always fail because they cannot start a redis instance at the same time as another class. If we run `DevAddrcacheTest` and `PreferredGatewayTestWithRedis` at the same time we can reproduce this error.

PR #924, opened 2021-11-29
Reuse existing redis containers for integration tests

bastbu:
# PR for issue #914

## What is being addressed

With the existing test setup against Redis, there was interference between existing docker Redis containers and test redis containers. With this PR we reuse already existing Redis containers and make sure that all tests classes relying on Redis run in sequence.

### DIFF
--- a/Tests/Integration/RedisFixture.cs
+++ b/Tests/Integration/RedisFixture.cs
@@ -95,7 +95,7 @@ namespace LoRaWan.Tests.Integration
             catch (DockerApiException e) when (e.StatusCode == HttpStatusCode.Conflict)
             {
                 var container = containers.FirstOrDefault(c => c.Names.Contains("/" + GetContainerName(RedisPort)));
-                if (container.State.Equals("exited", StringComparison.OrdinalIgnoreCase))
+                if (container is { } c && c.State.Equals("exited", StringComparison.OrdinalIgnoreCase))
                 {
                     Console.WriteLine("Starting existing container.");
                     await client.Containers.StartContainerAsync(container.ID, new ContainerStartParameters());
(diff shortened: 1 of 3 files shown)


bastbu:
> Aren't you concerned that data in Redis from a previous test run might impact the current test run?

@Mandur The same goes for tests in the same test class that run against the same Redis instance. In that case we also need to solve the interference problem on test code level. If we fail to do that, it is a bug and should be reported as such. Currently we have only two test classes that run against the Redis container.

If you do not want to reuse Redis containers that were created specifically for this test code, I can also revert reusing the containers and just improve the error message. As a consequence we will get the failures again on our local development machines if the containers already exist and are in state exited, but it's also not worse than before.

spygi:
>  The same goes for tests in the same test class that run against the same Redis instance. In that case we also need to solve the interference problem on test code level.

This is true but technically speaking Mik has a point: previously we could have "interference" between tests in the same class. Now that all tests run against the same container we have potentially introduced "interference" between multiple test classes.

In practice I think it's not an issue since we run sequentially the tests and assert on a 1-1 case.

Tests/Integration/RedisFixture.cs
```diff
@@ -57,44 +59,52 @@ private async Task StartRedisContainer()
                 {
                     Hostname = "localhost"
                 };
-                this.redisPort = Interlocked.Increment(ref uniqueRedisPort);
```
spygi, line 60: [nit] we could let the team know that they don't need the second redis container anymore. Not sure if any docs need to be updated but I would assume no.

spygi (approved):

Tests/Integration/RedisFixture.cs
bastbu: I checked the docs too, and didn't find an occurrence for this. I will let the team know if we decide to go ahead with one Redis container, and else abandon the PR.

Mandur:
I checked the tests and indeed every guid is generated by the newguid. I think collision risk is acceptable as @spygi pointed out, let's merge this.

Issue #881, opened 2021-11-23
The join table for China should be added


Mandur:
There is a join table missing. This about selecting the downstream frequency to respond to the join message

PR #925, opened 2021-11-29
Add downstream join frequencies for region CN470

MaggieSalak:
# PR for issue #881 

## What is being addressed

Only upstream join frequencies were previously supported for region CN470.

## How is this addressed

- Upstream join frequencies were replaces with a list of tuples using format: (upstream join freq, downstream join freq).
- Unit tests for join channels in China were expanded.

The changes were previously added in a different branch and tested manually: https://github.com/Azure/iotedge-lorawan-starterkit/commit/01b8ddac19275879ff81908f9a53e43010eadffb

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
@@ -15,6 +15,8 @@ namespace LoRaTools.Regions
     {
         private const ushort MAX_RX_DELAY = 15;
 
+        protected const double EPSILON = 0.00001;
+
         public LoRaRegionType LoRaRegion { get; set; }
 
         /// <summary>
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
@@ -20,7 +20,7 @@ namespace LoRaTools.Regions
 
         // Dictionary mapping upstream join frequencies to a tuple containing
         // the corresponding downstream join frequency and the channel index
-        public Dictionary<double, (double, int)> UpstreamJoinFrequenciesToDownstreamAndChannelIndex { get; }
+        public Dictionary<double, (double downstreamFreq, int joinChannelIndex)> UpstreamJoinFrequenciesToDownstreamAndChannelIndex { get; }
 
         public RegionCN470()
             : base(LoRaRegionType.CN470)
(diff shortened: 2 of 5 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
```diff
@@ -69,6 +69,14 @@ public RegionCN470()
             MaxADRDataRate = 7;
             RegionLimits = new RegionLimits((min: 470.3, max: 509.7), validDatarates, validDatarates, 0, 0);
 
+            UpstreamAndDownstreamJoinFrequencies = new List<(double, double)>
```
Mandur, line 72: Is there a reason we using a list here, Wouldn't a dictionary perform better?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
```diff
@@ -15,6 +15,8 @@ public abstract class Region
     {
         private const ushort MAX_RX_DELAY = 15;
 
+        protected const double EPSILON = 0.00001;
```
Mandur, line 18: Why do we have this tolerance? I never saw the frequencies having some delta with plain values from the packet forwarder/basic station onward.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
MaggieSalak: No because the frequencies need to be in the given order as they map to the channel number (which is the index in the list).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/Region.cs
MaggieSalak: I thought this would be safer than just comparing the equality of two doubles, here is what I found in the official docs https://docs.microsoft.com/en-us/dotnet/api/system.double.equals?redirectedfrom=MSDN&view=net-6.0#System_Double_Equals_System_Double_ where using a threshold seems to be recommended when comparing doubles. In this case it might be indeed not required since we are not doing any arithmetic operations on the frequencies so I'm opening to reverting this - let me know.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470.cs
```diff
@@ -69,6 +71,14 @@ public RegionCN470()
             MaxADRDataRate = 7;
             RegionLimits = new RegionLimits((min: 470.3, max: 509.7), validDatarates, validDatarates, 0, 0);
 
+            UpstreamJoinFrequenciesToDownstreamAndChannelIndex = new Dictionary<double, (double, int)>
```
Mandur, line 74: Could we give names to the tuple elements, (just so we remember what they are in the future). Not sure the syntax is 100% thought
```suggestion
            UpstreamJoinFrequenciesToDownstreamAndChannelIndex = new Dictionary<double, (downstreamFreq: double, joinChannelIndex: int)>
```

Mandur (approved):

PR #935, opened 2021-11-30
Feature/747 raise custom metrics

bastbu:
# PR for issue #887 

## What is being addressed

With this PR we raise custom metrics as described in the ADR. The PR goes together with #936, as we make some changes to some of the metrics we raise. Specifically,
- metrics around error rates were removed, since they are hard to distinguish in code which source they have.
- metrics around C2D deliveries and abandons were removed, since they are tracked by IoT Hub automatically. This will imply that these metrics will not be available on the Prometheus endpoint.

Metrics around caching were not yet added, since they depend on #904 to be merged first.

To support more kinds of metrics, we add support for `ObservableGauge<T>` type metrics for both Application Insights and Prometheus exporters.

## How is this addressed

Metrics are always added (optionally) via the DI `Meter`. For test code, we never raise metrics since they might interfere with each other (as they rely on callbacks for each listener and will get notified from other tests, potentially lacking the correct `AsyncLocal` state in the `RegistryMetricTagBag`).

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
@@ -35,6 +35,7 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
         private readonly ILogger<LnsProtocolMessageProcessor> logger;
         private readonly RegistryMetricTagBag registryMetricTagBag;
         private readonly Counter<int> joinRequestCounter;
+        private readonly Counter<int> uplinkMessageCounter;
 
         public LnsProtocolMessageProcessor(IBasicsStationConfigurationService basicsStationConfigurationService,
                                            WebSocketWriterRegistry<StationEui, string> socketWriterRegistry,
@@ -55,6 +56,7 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
             this.logger = logger;
             this.registryMetricTagBag = registryMetricTagBag;
             this.joinRequestCounter = meter?.CreateCounter<int>(MetricRegistry.JoinRequests);
+            this.uplinkMessageCounter = meter?.CreateCounter<int>(MetricRegistry.D2CMessagesReceived);
         }
 
         internal async Task<HttpContext> ProcessIncomingRequestAsync(HttpContext httpContext,
@@ -240,6 +242,7 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
                         }
 
                         using var scope = this.logger.BeginDeviceAddressScope(updf.DevAddr);
+                        this.uplinkMessageCounter.Add(1);
 
                         var routerRegion = await this.basicsStationConfigurationService.GetRegionAsync(stationEui, cancellationToken);
                         var rxpk = new BasicStationToRxpk(updf.RadioMetadata, routerRegion);
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
@@ -38,7 +38,6 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
         private readonly RegistryMetricTagBag registryMetricTagBag;
         private readonly Counter<int> joinRequestCounter;
         private readonly Counter<int> uplinkMessageCounter;
-        private readonly Histogram<double> dataMessageDispatchLatency;
 
         public LnsProtocolMessageProcessor(IBasicsStationConfigurationService basicsStationConfigurationService,
                                            WebSocketWriterRegistry<StationEui, string> socketWriterRegistry,
@@ -60,7 +59,6 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
             this.registryMetricTagBag = registryMetricTagBag;
             this.joinRequestCounter = meter?.CreateCounter<int>(MetricRegistry.JoinRequests);
             this.uplinkMessageCounter = meter?.CreateCounter<int>(MetricRegistry.D2CMessagesReceived);
-            this.dataMessageDispatchLatency = meter?.CreateHistogram<double>(MetricRegistry.DataMessageDispatchLatency);
         }
 
         internal async Task<HttpContext> ProcessIncomingRequestAsync(HttpContext httpContext,
@@ -194,9 +192,7 @@ namespace LoRaWan.NetworkServer.BasicsStation.Processors
                                                   string json,
                                                   CancellationToken cancellationToken)
         {
-            var messageType = LnsData.MessageTypeReader.Read(json);
-            var stopwatch = Stopwatch.StartNew();
-            switch (messageType)
+            switch (LnsData.MessageTypeReader.Read(json))
             {
                 case LnsMessageType.Version:
                     var stationVersion = LnsData.VersionMessageReader.Read(json);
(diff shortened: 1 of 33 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -240,6 +248,7 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
                         }
 
                         using var scope = this.logger.BeginDeviceAddressScope(updf.DevAddr);
+                        this.uplinkMessageCounter?.Add(1);
```
Mandur, line 251: Why are all the metrics nullable? I thought we always raised them in prometheus no matter what?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -188,7 +194,9 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
                                                   string json,
                                                   CancellationToken cancellationToken)
         {
-            switch (LnsData.MessageTypeReader.Read(json))
+            var messageType = LnsData.MessageTypeReader.Read(json);
+            var stopwatch = Stopwatch.StartNew();
```
Mandur, line 198: We now have a stopwatch here and another one in the DefaultloRaProcessinghandler loop. Maybe we could remove it from the handler and remove it here. **Nothing to do now**, just floating the ideas to take opinions

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -395,17 +408,20 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
                 {
                     if (confirmDownlinkMessageBuilderResp.DownlinkPktFwdMessage == null)
                     {
+                        this.receiveWindowMissed?.Add(1);
                         this.logger.LogInformation($"out of time for downstream message, will abandon cloud to device message id: {cloudToDeviceMessage.MessageId ?? "undefined"}");
                         _ = cloudToDeviceMessage.AbandonAsync();
                     }
                     else if (confirmDownlinkMessageBuilderResp.IsMessageTooLong)
                     {
+                        this.receiveWindowMissed?.Add(1);
```
Mandur, line 424: This is not rx windows miss. We are not sending as the datarate the device is using doesn't allow us to carry a message of this size. Could be useful to have a metric for that as I am adding a way to change Datarate for devices, someone could programmatically request a DR change when this metric here is too frequent

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
```diff
@@ -113,6 +120,7 @@ public IWebSocketWriterHandle<TMessage> Register(TKey key, IWebSocketWriter<TMes
             {
                 if (this.sockets.TryGetValue(key, out var current))
                 {
+                    this.stationConnectivityLostCounter?.Add(1);
```
danigian, line 123: shouldn't we track the key/station eui here?

Tests/Unit/NetworkServer/ApplicationInsightsMetricExporterTests.cs
```diff
@@ -122,6 +127,28 @@ public void ApplicationInsights_Metrics_Collection_Raises_Histogram_Metrics()
...
+        [Fact]
+        public async void When_ObservableGauge_Is_Recorded_Should_Export_To_ApplicationInsights()
+        {
+            // arrange
+            var observeValue = new Mock<Func<Measurement<int>>>();
+            var stationEui = new StationEui(1);
+            var measurement = new Measurement<int>(1, KeyValuePair.Create(MetricRegistry.GatewayIdTagName, (object)stationEui));
+            _ = observeValue.Setup(ov => ov.Invoke()).Returns(measurement);
+            using var meter = new Meter("LoRaWan", "1.0");
+            var observableGauge = meter.CreateObservableGauge(ObservableGaugeMetric.Name, observeValue.Object);
```
danigian, line 139: Check the codeql warning related to useless assignment to local variable.

Tests/Unit/NetworkServer/PrometheusMetricExporterTests.cs
```diff
@@ -117,6 +125,25 @@ public void When_Histogram_Series_Is_Recorded_Should_Export_To_Prometheus()
...
+        [Fact]
+        public async void When_ObservableGauge_Is_Recorded_Should_Export_To_Prometheus()
+        {
+            // arrange
+            var observeValue = new Mock<Func<Measurement<int>>>();
+            var stationEui = new StationEui(1);
+            var measurement = new Measurement<int>(1, KeyValuePair.Create(MetricRegistry.GatewayIdTagName, (object?)stationEui));
+            observeValue.Setup(ov => ov.Invoke()).Returns(measurement);
+            using var meter = new Meter("LoRaWan", "1.0");
+            var observableGauge = meter.CreateObservableGauge(ObservableGauge.Name, observeValue.Object);
```
danigian, line 137: Check codeql warning

danigian (commented):
Left some comments.
In addition to that, apparently a submodule (not needed) was added:
image

Please remove

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
bastbu: This is a bit tricky. Since we rely on the presence of the Gateway ID in the tag bag (which uses `AsyncLocal` to make it available in "ambient environment", we could potentially have interference if we raise metrics when we test. Hence we allow optional raising of metrics. Alternatively, we could also inject a `Meter` with a different namespace from the test code, if you prefer that to allow the nullability of the metrics.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
bastbu: This changed with 06209ea163745c0d8ab6044e390e728c78689c78

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
bastbu: This is why we have the `RegistryMetricTagBag`. The station EUI is made available through ambient environment and gets picked up by the exporters when raising the metrics. With that, we don't have to explicitly attach the Gateway ID to every metric that we raise. It's essentially the same technique that is used by logging scopes.

Tests/Unit/NetworkServer/PrometheusMetricExporterTests.cs
bastbu: Thanks for the hint, since the CodeQL warnings break the build I usually fix those before publishing the PR.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: And added the metric that you suggested with 5ebf65313d67aaf1591c977d2faa0cc38488cfa9

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
Mandur: That make sense, let's continue like this

danigian (approved):

PR #936, opened 2021-11-30
Feature/747 custom metrics updates

bastbu:
# PR for issue #747

## What is being addressed

With this update to the observability ADR, which is based on #935, we remove some metrics that are hard to track (and not worth the effort) and some that are automatically tracked by IoT Hub and we don't need to manually track.

### DIFF
--- a/docs/adr/005_observability.md
+++ b/docs/adr/005_observability.md
@@ -83,10 +83,8 @@ In addition to this, we will support Application Insights metrics on an opt-in b
 | ActiveStationConnections   | Active connections to stations                               | LNS    | LoRaWan   | Gateway Id                 |
 | D2CMessagesReceived        | Number of messages received from device                      | LNS    | LoRaWan   | Gateway Id                 |
 | D2CMessageDeliveryLatency  | Time from when we received the message from the concentrator until we are done processing it | LNS    | LoRaWan   | Gateway Id                 |
-| D2CMessagesError | Number of errors processing downstream messages | LNS    | LoRaWan   | Gateway Id            |
 | D2CMessageSize             | Message size in bytes received from device                   | LNS    | LoRaWan   | Gateway Id                 |
 | C2DMessagesAbandoned       | Number of C2D messages abandoned                             | LNS    | LoRaWan   | Gateway Id                 |
-| C2DMessagesError           | Number of errors in sending messages to downstream           | LNS    | LoRaWan   | Gateway Id, To             |
 | C2DMessageSizeDownstream   | Message size in bytes sent downstream                        | LNS    | LoRaWan   | Gateway Id, To             |
 
 ### Alerts
--- a/docs/adr/005_observability.md
+++ b/docs/adr/005_observability.md
@@ -81,6 +81,7 @@ In addition to this, we will support Application Insights metrics on an opt-in b
 | JoinRequests               | Number of join requests                                      | LNS    | LoRaWan   | Gateway Id                 |
 | StationConnectivityLost    | Connection to LBS lost                                       | LNS    | LoRaWan   | Gateway Id                 |
 | ActiveStationConnections   | Active connections to stations                               | LNS    | LoRaWan   | Gateway Id                 |
+| ProcessingErrors           | Number of errors in LNS processing                           | LNS    | LoRaWan   | Gateway Id                 |
 | D2CMessagesReceived        | Number of messages received from device                      | LNS    | LoRaWan   | Gateway Id                 |
 | D2CMessageDeliveryLatency  | Time from when we dispatched the message sent from the concentrator until we are done processing it | LNS | LoRaWan | Gateway Id |
 | D2CMessageSize             | Message size in bytes received from device                   | LNS    | LoRaWan   | Gateway Id                 |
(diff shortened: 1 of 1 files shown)


docs/adr/005_observability.md
```diff
@@ -74,24 +74,19 @@ In addition to this, we will support Application Insights metrics on an opt-in b
...
-| D2CMessagesError           | Number of errors in sending messages to upstream             | LNS    | LoRaWan   | Gateway Id, To            |
-| D2CMessagesProcessingError | Number of errors processing (decoding, decrypting) messages  | LNS    | LoRaWan   | Gateway Id, To            |
+| DataMessageDeliveryLatency | Time from when we dispatched the message sent from the concentrator until we are done processing it | LNS    | LoRaWan   | Gateway Id                 |
+| DataMessageDispatchLatency | Time from when we received the message from the concentrator until we dispatch the message to be handled (internal metric) | LNS | LoRaWan | Gateway Id, Message Type |
 | D2CMessageSize             | Message size in bytes received from device                   | LNS    | LoRaWan   | Gateway Id                 |
-| D2CMessageSizeUpstream     | Message size in bytes sent upstream                          | LNS    | LoRaWan   | Gateway Id, To            |
 | C2DMessagesAbandoned       | Number of C2D messages abandoned                             | LNS    | LoRaWan   | Gateway Id                 |
```
Mandur, line 88: [Brainstorming] Would it be something to track the signal strength (RSSI) and signal to noise ratio (RSSI)

Mandur (commented):
LGTM to me. I would wait that @p-schuler to review it as he is much more in depth in such things ;)

danigian (approved):

docs/adr/005_observability.md
```diff
@@ -74,24 +74,19 @@ In addition to this, we will support Application Insights metrics on an opt-in b
...
+| D2CMessageDeliveryLatency  | Time from when we dispatched the message sent from the concentrator until we are done processing it | LNS | LoRaWan | Gateway Id |
 | D2CMessageSize             | Message size in bytes received from device                   | LNS    | LoRaWan   | Gateway Id                 |
-| D2CMessageSizeUpstream     | Message size in bytes sent upstream                          | LNS    | LoRaWan   | Gateway Id, To            |
 | C2DMessagesAbandoned       | Number of C2D messages abandoned                             | LNS    | LoRaWan   | Gateway Id                 |
-| C2DMessagesDelivered       | Number of messages sent downstream                           | LNS    | LoRaWan   | Gateway Id, To             |
-| C2DMessagesError           | Number of errors in sending messages to downstream           | LNS    | LoRaWan   | Gateway Id, To             |
-| C2DMessageSizeDownstream   | Message size in bytes sent downstream                        | LNS    | LoRaWan   | Gateway Id, To             |
+| C2DMessageSizeDownstream   | Message size in bytes sent downstream                        | LNS    | LoRaWan   | Gateway Id                 |
```
p-schuler, line 88: What's the reason for this? Isn't it enough, if we know they are too long? Not sure what we will derive from this number and how we act on it?

docs/adr/005_observability.md
p-schuler: What would you use it for? Optimize the deployment of the concentrators? You would need to know that devices have a low RSSI, no? Tracking the device id as dimension, is probably not possible due to the high cardinality.

p-schuler (approved):
I'm not sure about the C2DMessageSizeDownstream requirement. If we have a good case, we can add it.

docs/adr/005_observability.md
bastbu: Removed with 213b3ce25d47e9330be9c718e390697c0c77dc3c

docs/adr/005_observability.md
```diff
@@ -74,24 +74,17 @@ In addition to this, we will support Application Insights metrics on an opt-in b
...
 | DeviceLoadRequests         | Number of device load requests                               | LNS    | LoRaWan   | Gateway Id                 |
 | JoinRequests               | Number of join requests                                      | LNS    | LoRaWan   | Gateway Id                 |
 | StationConnectivityLost    | Connection to LBS lost                                       | LNS    | LoRaWan   | Gateway Id                 |
 | ActiveStationConnections   | Active connections to stations                               | LNS    | LoRaWan   | Gateway Id                 |
 | D2CMessagesReceived        | Number of messages received from device                      | LNS    | LoRaWan   | Gateway Id                 |
-| D2CMessageDeliveryLatency  | Time from when we received the message from the concentrator until we are done processing it | LNS    | LoRaWan   | Gateway Id                 |
-| D2CMessagesDelivered       | Number of messages sent to upstream                          | LNS    | LoRaWan   | Gateway Id, To             |
-| D2CMessagesError           | Number of errors in sending messages to upstream             | LNS    | LoRaWan   | Gateway Id, To            |
```
eedorenko, line 87: HighUpstreamMessageErrorRatio and HighDownstreamMessageErrorRatio won't be possible without "error" metrics.

bastbu: I removed the error metrics as discussed with @p-schuler. Alerting on unhandled exceptions should be good enough, as in the LoRa case it's quite hard to distinguish whether errors occur due to downstream or upstream messages (the logic is interconnected due to the nature of LoRa with the receive windows for class A devices)

bastbu: Now I brought it back to also give users that do not use Azure Monitor insights around the number of Processing Errors.

Issue #900, opened 2021-11-25
Move uplink data frame deduplication later in the pipeline when we have information about the deduplication strategy chosen. 


spygi:
If it's mark or none let the message pass.

PR #937, opened 2021-11-30
Bug/900 respect deduplication strategy

spygi:
# PR for issue #900 

## What is being addressed

By moving the deduplication of concentrators earlier in the processing, we are not able to handle the scenario of triangulation: there we use the Mark deduplication strategy and we should not drop messages upstream even if they are duplicates. The deduplication strategy however is not available that early in the pipeline. 

With this PR we move the processing further later when we do have access to the device twin.

## How is this addressed

- The deduplication service is removed from the `LnsProtocolMessageProcessor` and injected in `JoinRequestMessageHandler` and `DefaultLoRaDataRequestHandler`
- The cache keys are kept the same as before as this PR focuses on the flow and edge cases.
  - There we miss some methods on the LoRaWan primitives that can be added in a different PR (out of scope).
- The `IConcentratorDeduplication` and concrete implementation do not need to use generics anymore: separate methods were introduced for simplifying the 2 flows.
- Unit tests
- Integration tests for the different configurations:
  - Confirmed and confirmed data messages (first and subsequent), join requests, ADR enabled messages, MAC commands, single gateway.

## Notable changes from existing behavior

- Previously when we encountered a duplicate message, we would not send it upstream, irrespective from the deduplication strategy. Now if strategy is not Drop message is sent upstream.
- We now include also unconfirmed messages in the deduplication (previously only for confirmed messages).

## Out of scope for this PR
- Class C devices, sending downstream messages using direct method #1056 
- Simplify Cache keys with DevEui information and unit test key generation #1003


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -202,11 +196,6 @@ internal async Task InternalHandleDataAsync(RouteValueDictionary routeValues, We
                     {
                         var jreq = LnsData.JoinRequestFrameReader.Read(json);
 
-                        if (this.joinRequestDeduplication.ShouldDrop(jreq, stationEui))
-                        {
-                            break;
-                        }
-
                         this.joinRequestCounter?.Add(1);
```
bastbu, line 199: [nit] please pay attention that the semantics of that metric does not change (we only count the number of join requests after dedup)

spygi: Since the deduplication is moved, we would need to move this later then.

spygi: Fixed with fc67d8b8bae90890ac4bba8569a57a53ec0ea7be

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -757,8 +776,11 @@ private bool ValidateRequest(LoRaRequest request, bool isFrameCounterFromNewlySt
             return valid;
         }
 
-        private async Task<bool> DetermineIfFramecounterIsFromNewlyStartedDeviceAsync(LoRaDevice loRaDevice, uint payloadFcnt, ILoRaDeviceFrameCounterUpdateStrategy frameCounterStrategy)
+        internal virtual async Task<bool> DetermineIfFramecounterIsFromNewlyStartedDeviceAsync(LoRaDevice loRaDevice, uint payloadFcnt, ILoRaDeviceFrameCounterUpdateStrategy frameCounterStrategy)
```
spygi, line 779: Change required in order to be able to mock it for integration test.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
```diff
@@ -72,7 +72,7 @@ public bool TryGetHandle(TKey key, [NotNullWhen(true)] out IWebSocketWriterHandl
         /// <remarks>
         /// Implementation does not throw when key is not present in the underlying dictionary.
         /// </remarks>
-        public bool IsSocketWriterOpen(TKey key)
+        public virtual bool IsSocketWriterOpen(TKey key)
```
spygi, line 75: Needed for integration test mock.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,19 +29,32 @@ internal sealed class ConcentratorDeduplication<T> :
...
             this.socketRegistry = socketRegistry;
             this.logger = logger;
         }
 
-        public bool ShouldDrop(T frame, StationEui stationEui)
+        /// <summary>
+        /// Checks if this class should process the specified request and decide later to drop or not.
+        /// </summary>
+        internal bool ShouldProcess(LoRaRequest loRaRequest, LoRaDevice loRaDevice)
+            => (loRaRequest.Payload is LoRaPayloadData && this.deduplicationStrategy.Create(loRaDevice) is DeduplicationStrategyDrop)
```
spygi, line 46: ## Problem
- Currently by default, deduplication strategy is None.
- Concentrator Deduplication in that case let's multiple requests pass through.
- The function does not set FrameCounterDown = 0
- We get a collision downstream

## Possible solutions
- Concentrator Deduplication drops if strategy is Drop **or** None, which is a bit weird.
- We introduce a different configuration for concentrator (vs "abusing" the one for the gateway/network server)

p-schuler: The problem here is only happening in single gateway setup. If we do not drop the messages and the configuration for the device is having a gateway assigned, we do simply increment the framecounter on request, without validating if that request for that particular device was already handled. Hence, there we would need a strategy that takes multi concentrator messages into account.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -676,11 +692,12 @@ private void LogFrameCounterDownState(LoRaDevice loRaDevice, uint newFcntDown)
             }
         }
 
-        private async Task<FunctionBundlerResult> TryUseBundler(LoRaRequest request, LoRaDevice loRaDevice, LoRaPayloadData loraPayload, bool useMultipleGateways)
+        private async Task<FunctionBundlerResult> TryUseBundler(LoRaRequest request, LoRaDevice loRaDevice, LoRaPayloadData loraPayload, bool useMultipleGateways, ConcentratorDeduplication.Result deduplicationResult)
         {
             FunctionBundlerResult bundlerResult = null;
-            if (useMultipleGateways)
+            if (useMultipleGateways && (deduplicationResult is ConcentratorDeduplication.Result.NotDuplicate || deduplicationResult is ConcentratorDeduplication.Result.DuplicateDueToResubmission))
```
spygi, line 698: Please review the `DuplicateDueToResubmission` case.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +36,103 @@ internal sealed class ConcentratorDeduplication<T> :
...
         {
-            if (frame == null) throw new ArgumentNullException(nameof(frame));
+            _ = loRaRequest ?? throw new ArgumentNullException(nameof(loRaRequest));
 
-            var key = CreateCacheKey(frame);
+            var key = CreateCacheKey(loRaRequest);
+            var stationEui = loRaRequest.StationEui;
 
             StationEui previousStation;
             lock (this.cache)
```
p-schuler, line 55: locking on the cache means we don't control the lock. Anyone, injecting the memory cache instance, could be locking on it. I would suggest adding a dedicated lock on a private field we control. Having a public accessible object we lock on can have very hard to track bugs like deadlocks (when nested locking occurs for example).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +36,103 @@ internal sealed class ConcentratorDeduplication<T> :
...
-        private void AddToCache(string key, StationEui stationEui)
-            => this.cache.Set(key, stationEui, new MemoryCacheEntryOptions()
-            {
-                SlidingExpiration = DefaultExpiration
-            });
-
-        public void Dispose() => this.cache.Dispose();
+        private bool ShouldDrop(LoRaRequest loRaRequest, LoRaDevice? loRaDevice)
+            => loRaRequest.Payload is LoRaPayloadJoinRequest
+            || (loRaRequest.Payload is LoRaPayloadData && this.deduplicationStrategy.Create(loRaDevice) is DeduplicationStrategyDrop);
```
p-schuler, line 136: I think we can simply look up the strategy on the LoRaDevice.Deduplication and avoid the creation of a object on the heap, that is garbage immediately.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +36,103 @@ internal sealed class ConcentratorDeduplication<T> :
...
                 }
             }
 
-            if (previousStation == stationEui)
+            if (LoRaRequest.RequiresConfirmation(loRaRequest) && previousStation == stationEui)
             {
-                // considered as a resubmit
-                this.logger.LogDebug("Message received from the same EUI {StationEui} as before, will not drop.", stationEui);
-                return false;
+                this.logger.LogDebug($"Message received from the same EUI {stationEui} as before, will not drop.");
```
p-schuler, line 69: message here is not clearly indicating that this is a confirmed resubmit and the same station is handling it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +36,103 @@ internal sealed class ConcentratorDeduplication<T> :
...
-                UpstreamDataFrame asDataFrame => CreateCacheKeyCore(asDataFrame),
-                JoinRequestFrame asJoinRequestFrame => CreateCacheKeyCore(asJoinRequestFrame),
-                _ => throw new ArgumentException($"{frame} with invalid type.")
+                LoRaPayloadData asDataPayload => CreateCacheKeyCore(asDataPayload),
+                LoRaPayloadJoinRequest asJoinPayload => CreateCacheKeyCore(asJoinPayload),
+                _ => throw new ArgumentException($"{loRaRequest} with invalid type.")
             };
 
-        private static string CreateCacheKeyCore(UpstreamDataFrame updf)
+        private static string CreateCacheKeyCore(LoRaPayloadData payload)
```
p-schuler, line 92: can we make those unit testable? I feel like the logic in here, needs to be tested thoroughly.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +36,103 @@ internal sealed class ConcentratorDeduplication<T> :
...
-                return false;
+                this.logger.LogDebug($"Message received from the same EUI {stationEui} as before, will not drop.");
+                return Result.DuplicateDueToResubmission;
             }
 
             // received from a different station
-            if (this.socketRegistry.IsSocketWriterOpen(previousStation))
+            if (ShouldDrop(loRaRequest, loRaDevice))
             {
                 this.logger.LogInformation($"{Constants.DuplicateMessageFromAnotherStationMsg} with EUI {stationEui}, will drop.");
```
p-schuler, line 76: information logs should be made sparse. This is a normal case that is potentially being spit out several times for every message (if we have drop and multiple concentrators connected). Can we change this to Debug?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaRequest.cs
```diff
@@ -61,6 +61,9 @@ public virtual void NotifySucceeded(LoRaDevice loRaDevice, DownlinkPktFwdMessage
 
         internal void SetStationEui(StationEui stationEui) => StationEui = stationEui;
 
+        internal static bool RequiresConfirmation(LoRaRequest loraRequest)
+            => loraRequest.Payload is LoRaPayloadData payload && (payload.IsConfirmed || payload.IsMacAnswerRequired);
```
p-schuler, line 65: why not make this part of the LoRaRequest, overwrite in LoRaPayloadData only and return false for the others? That way you can avoid the casting and we have it controlled by the LoRaPayloadData directly?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: True, changed in bb8d581975b886a8e2ed87aca879d1e826b43d05

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Not sure what you would like to see here. I added the "confirmedResubmit" c29c90d54bddea7d2ff1caf0f69f027f0a14df33 as part of the log because I understand this is the keyword you were missing although imo "confirmed resubmit" implies that message was confirmed already (which is not the case).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Right, as already discussed 🤦‍♂️ Changed in 0a8c12d482b230cb20195f4152e3eefc59b8c318 to use a `SemaphoreSlim`

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaRequest.cs
spygi: Good idea, done in 7d7d04a1b48d186903bb1e9878715632d2f2dd0c

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Right now we unit test whether we get the expected output. I think we discussed with Atif whether we should test further but dropped the idea as this would be an implementation detail bound to change (might remember wrong though)? This will be the case actually if we simplify keys in #1003 because now we have the DevEUI - if ok for you I will address this in that task.

Since we are using static methods within not sure what we can unit test further. Also the output is hashed so we can't observe whether a field is in there either - we can only check if the output changed.

p-schuler: What I was thinking is that we should test that the key actually changes when we change the relevant input parameter from the parameters. We also have a few variants with optional arguments (RawMessage) that may or may not be included and impacts the key generation. Even if the implementation may change, it would add value having tests for the current implementation. The linked item is not assigned yet and could be dropped as it is a nice to have.

There are multiple options on how you can make this testable. If you want to keep it static, you can mark it protected, so you can call it from a test class that inherits from the ConcentratorDeduplication. If you want to validate not only the hash but also the content, you could factor out the hash computation into a virtual protected method you can override that you could mock and not create a hash.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +39,110 @@ internal sealed class ConcentratorDeduplication<T> :
...
-            ILogger<IConcentratorDeduplication<T>> logger)
+            ILogger<IConcentratorDeduplication> logger)
         {
             this.cache = cache;
-            this.socketRegistry = socketRegistry;
             this.logger = logger;
         }
 
-        public bool ShouldDrop(T frame, StationEui stationEui)
+        public async Task<Result> CheckDuplicateAsync(LoRaRequest loRaRequest, LoRaDevice? loRaDevice)
```
p-schuler, line 48: Why did you make the method async? Was this to add the semaphore? You don't need that here. You can simply add an object as a field for example

`private static readonly object cacheLock = new object();`

since the concentrator deduplicator is also a singleton, in theory you could also use a non-static field, however, that would ensure, if the registration changes, you still lock the cache (which is being added as a singleton as well).

p-schuler, line 48: I think the logic could be simplified here and made more explicit depending on the request type (join vs. data). If we had 2 explicit entry points, we could avoid special casing the 2 cases in the code. We can still achieve re-usability by having common methods that are called from both. I can help you look into this if you like to see what I have in mind.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -98,10 +101,21 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
             // ABP device does not reset the Fcnt so in relax mode we should reset for 0 (LMIC based) or 1
             var isFrameCounterFromNewlyStartedDevice = await DetermineIfFramecounterIsFromNewlyStartedDeviceAsync(loRaDevice, payloadFcntAdjusted, frameCounterStrategy);
```
p-schuler, line 102: Did you consider the case where we have an ABP reset? As far as I read this then if we have a reset, all of the messages (from each concentrator), would reset the fcnt. Meaning that if we processed a message before from another concentrator, the fcnt we stored, would get reset and we would go to the function multiple times to perform the reset? What was the reason to add the deduplication after this?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -8,20 +8,29 @@ namespace LoRaWan.NetworkServer
...
-        IConcentratorDeduplication<T>, IDisposable
-        where T : class
+    public sealed class ConcentratorDeduplication :
+        IConcentratorDeduplication, IDisposable
     {
+        public enum Result
+        {
+            NotDuplicate,
+            DuplicateDueToResubmission,
+            SoftDuplicate, // detected as a duplicate but due to the DeduplicationStrategy marked as a "soft" duplicate
```
p-schuler, line 24: nit: not sure if we can come up with a better name. I'd like to have information about the fact that this is a concentrator related duplicate, but we let it go, because the dedup strategy is mark or none. Maybe ConcentratorDuplicatePass and then for the `Duplicate` ConcentratorDuplicateDrop? Any other thoughts?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +39,110 @@ internal sealed class ConcentratorDeduplication<T> :
...
+                    return Result.NotDuplicate;
                 }
             }
+            finally
+            {
+                _ = this.cacheSemaphore.Release();
+            }
 
-            if (previousStation == stationEui)
+            if (loRaRequest.Payload.RequiresConfirmation() && previousStation == stationEui)
```
p-schuler, line 75: nit: I would turn around the 2 conditions. 
1. it makes it clearer that this condition only applies to the same station that is also stored in the cache
2. the validation of the stationEUI is quicker and all subsequent messages, don't need to validate the fact if it RequiresConfirmation.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -111,7 +125,7 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
 
             try
             {
-                var bundlerResult = await TryUseBundler(request, loRaDevice, loraPayload, useMultipleGateways);
+                var bundlerResult = await TryUseBundler(request, loRaDevice, loraPayload, useMultipleGateways, concentratorDeduplicationResult);
```
p-schuler, line 128: can't add a comment on that line, as it wasn't changed (thanks gitHub), but on line 147 we do Perform the ADR not matter if it was a duplicate or not. Meaning that we have a roundtrip to the server no matter if it is a duplicate or not. 

We discussed this, and this might be desired. The ADR is a bit tricky to get right with multi concentrators, as the values will differ from each concentrator. If we collect all values from all the concentrators, we may adapt invalid parameters that satisfy all concentrators on avg, failing to optimize the DR to what it would be possible. 

A simple option would be to skip the ADR collection for duplicate or soft duplicate messages and only collect for the first one. We would make the assumption that the first one getting the message would be the one closest to the device (likely, but not necessary). In theory, the ADR would have to be adapted to take multiple concentrators into consideration and optimize for the one closest to the device.

p-schuler: One more thing: if we do the ADR for all of them, the requiresConfirmation may be reset and we end up sending multiple confirmations for that message (if the ADR table is complete). 

I suggest to do the following:

1. disable ADR processing for duplicate concentrator messages
2. add a backlog item to look at the ADR implementation and take multi concentrators into consideration

/cc @Mandur  to chime in here

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -98,10 +101,21 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
 
+            var concentratorDeduplicationResult = await this.concentratorDeduplication.CheckDuplicateAsync(request, loRaDevice);
+            if (concentratorDeduplicationResult is ConcentratorDeduplication.Result.Duplicate)
+            {
+                return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.DeduplicationDrop);
+            }
+            else if (concentratorDeduplicationResult is ConcentratorDeduplication.Result.SoftDuplicate)
+            {
+                // Request is allowed upstream but confirmation is skipped to avoid collisions on the air.
+                requiresConfirmation = false;
```
p-schuler, line 112: check the places where we set this flag. You absolutely don't want to send a message down in this case, hence, there should be no code path resetting this. However, for that, I believe you need a separate flag to indicate, that you want this to be enforced - for example, if it is a MacCommand that you already processed I think there are 2 issues:

1. we will potentially contact the server for a framecounter down
2. we will reset the flag and send the answer to the device again

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -20,6 +20,7 @@ public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
     {
         private readonly NetworkServerConfiguration configuration;
         private readonly ILoRaDeviceFrameCounterUpdateStrategyProvider frameCounterUpdateStrategyProvider;
+        private readonly IConcentratorDeduplication concentratorDeduplication;
```
p-schuler, line 23: Again: can't add comments on lines that were not changed, hence adding it here:

https://github.com/Azure/iotedge-lorawan-starterkit/pull/937/files?diff=unified&w=0#diff-68a0d3f9eca3be1aa5e03c4aac5d01e4cab390524ba9cc53e5b59bbd3746b7ccR195

Line 195: 

`var validFcntUp = isFrameCounterFromNewlyStartedDevice || (payloadFcntAdjusted > loRaDevice.FCntUp);`

How are we getting a validFcntUp ever with the same message? I'm confused how that will work, as the FCntUp on the device should reflect the last message sent. If we already processed the message, we would have that state in there?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: yes it was for that. Changed it with your suggestion in ceeab95c5ff2b16fbb7bc0705afe69c9ef4544a8

spygi: I can bring back the generics - I decided to remove them in this PR because I got a bit of a pushback on the previous review and especially now that we have a common type (`LoRaRequest`) as an entry point, I decided it was good enough without them. 

Shall we do the change here? Also, if you had another way in mind would be curious to learn :)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: > How are we getting a validFcntUp ever with the same message?

Not sure how/when this can happen. Do you see something breaking that we introduced here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Thanks for raising this, we said on our discussions last week to put ADR as out of scope but it was a bit mentioned on the side and we didn't define what "out of scope" means exactly. So it's good to discuss this, do you think we need to involve Ronnie for this decision? 

> disable ADR processing for duplicate concentrator messages

Here you mean the cases Duplicate, SoftDuplicate and Resubmission?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: What about `SoftDuplicateDueToDeduplicationStrategy`? I would avoid including "allow" or "drop" on the name because imo this service should only be concerned about the detection part and not with what will happen with the message. The decision to drop or allow depends on the caller and the semantics there could change in the future.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Changed in 745b009fa1dbc2b800b495f4700a523d6addeedb to reset only when not a duplicate.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -8,20 +8,29 @@ namespace LoRaWan.NetworkServer
...
     using Microsoft.Extensions.Caching.Memory;
     using Microsoft.Extensions.Logging;
 
-    internal sealed class ConcentratorDeduplication<T> :
-        IConcentratorDeduplication<T>, IDisposable
-        where T : class
+    public sealed class ConcentratorDeduplication :
+        IConcentratorDeduplication, IDisposable
     {
+        public enum Result
```
p-schuler, line 19: we should not nest this. It is part of the interface (the result is returned), hence it should not be nested in a specific implementation.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Changed to above suggestion in 94dbca2198ae7bf53718dba65f248cd5af317d54

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Done in 0671de319568b8c7273d9f0039c82f9df9c1660e

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: - Added a check to skip for soft duplicates in 7cb1b563072321e31e9cd8581cb87feeb46f78f6.
- Added a backlog item to re-enable ADR for multi concentrator support #1017 that could be discussed with Ronnie or @Mandur

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Please review 91aeb6300edbe6b0331ec51963c206bc47b0f38c

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Changed in 708d47d4e08ada6d83826be47122a5852771349d

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Added also a specific integration test that will break if we enable ADR for Soft duplicates: `When_ADR_Enabled_PerformADR_Should_Be_Disabled_For_SoftDuplicate`

Tests/Integration/DecoderTests.cs
```diff
@@ -802,9 +802,8 @@ public async Task When_Resent_Message_Using_Custom_Decoder_Returns_Complex_Objec
             Assert.NotNull(request2.ResponseDownlink);
             Assert.NotNull(loRaDeviceTelemetry);
             var rawPayload2 = Convert.ToBase64String(Encoding.UTF8.GetBytes("1"));
-            Assert.Equal(rawPayload, loRaDeviceTelemetry.Rawdata);
-            // Only the first message should be sent
-            LoRaDeviceClient.Verify(x => x.SendEventAsync(It.IsNotNull<LoRaDeviceTelemetry>(), null), Times.Once());
+            Assert.Equal(rawPayload2, loRaDeviceTelemetry.Rawdata);
```
spygi, line 805: unrelated fix.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: We removed the requires confirmation check since we want to deduplicate also unconfirmed messages.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -30,96 +27,114 @@ internal sealed class ConcentratorDeduplication<T> :
...
-                this.logger.LogDebug("Message received from the same EUI {StationEui} as before, will not drop.", stationEui);
-                return false;
+                this.logger.LogDebug($"{Constants.DuplicateMessageFromAnotherStationMsg} with EUI {loRaRequest.StationEui}.");
+                return ConcentratorDeduplicationResult.Duplicate;
             }
 
-            // received from a different station
-            if (this.socketRegistry.IsSocketWriterOpen(previousStation))
+            this.logger.LogDebug($"Message from station with EUI {loRaRequest.StationEui} marked as soft duplicate due to DeduplicationStrategy.");
+            return ConcentratorDeduplicationResult.SoftDuplicateDueToDeduplicationStrategy;
```
p-schuler, line 68: Let's simplify this:

```
            _ = loRaRequest ?? throw new ArgumentNullException(nameof(loRaRequest));
            _ = loRaDevice ?? throw new ArgumentNullException(nameof(loRaDevice));
            var key = CreateCacheKey((LoRaPayloadData)loRaRequest.Payload);
            if (EnsureFirstMessageInCache(key, loRaRequest, out var previousStation))
                return ConcentratorDeduplicationResult.NotDuplicate;

            var result = ConcentratorDeduplicationResult.SoftDuplicateDueToDeduplicationStrategy;

            if (previousStation == loRaRequest.StationEui)
            {
                result = ConcentratorDeduplicationResult.DuplicateDueToResubmission;
            }
            else if (loRaDevice.Deduplication == DeduplicationMode.Drop)
            {
                result = ConcentratorDeduplicationResult.Duplicate;
            }

            this.logger.LogDebug($"{Constants.DuplicateMessageFromAnotherStationMsg} {result}. Message was sent from station {loRaRequest.StationEui}.");
            return result;
```

This way we have only one log statement we need to write, we have a consistent log message for all cases and only the result is different and we can use the constant message to also check on the specific result in the E2E test.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
```diff
@@ -192,5 +192,8 @@ public static uint InferUpper32BitsForClientFcnt(ushort payloadFcnt, uint fcnt)
             var fcntServerUpper = fcnt & MaskHigher16;
             return fcntServerUpper | payloadFcnt;
         }
+
+        public virtual bool RequiresConfirmation()
```
p-schuler, line 196: can we make this a property?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -94,14 +97,26 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
 
+            var skipDownstreamToAvoidCollisions = false;
+            var concentratorDeduplicationResult = this.concentratorDeduplication.CheckDuplicateData(request, loRaDevice);
+            if (concentratorDeduplicationResult is ConcentratorDeduplicationResult.Duplicate)
+            {
+                return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.DeduplicationDrop);
+            }
+            else if (concentratorDeduplicationResult is ConcentratorDeduplicationResult.SoftDuplicateDueToDeduplicationStrategy)
+            {
+                // Request is allowed upstream but confirmation is skipped to avoid collisions on the air.
```
p-schuler, line 108: nit: 
```suggestion
                // Request is allowed upstream but confirmation is skipped to avoid sending the answer to the device multiple times and potentially cause collisions on the air.
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -156,29 +177,28 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
                 }
 
+                #region FrameCounterDown
                 // if deduplication already processed the next framecounter down, use that
                 var fcntDown = loRaADRResult?.FCntDown != null ? loRaADRResult.FCntDown : bundlerResult?.NextFCntDown;
 
-                if (fcntDown.HasValue)
-                {
-                    LogFrameCounterDownState(loRaDevice, fcntDown.Value);
-                }
```
p-schuler, line 165: why was this removed?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -156,29 +177,28 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
+                    // In multi gateway scenarios it means another gateway has won the race to handle this message, can stop now
                     if (fcntDown <= 0)
                     {
                         return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.HandledByAnotherGateway);
                     }
                 }
+                #endregion
 
-                var validFcntUp = isFrameCounterFromNewlyStartedDevice || (payloadFcntAdjusted > loRaDevice.FCntUp);
+                var validFcntUp = isFrameCounterFromNewlyStartedDevice
```
p-schuler, line 199: how about renaming this to canSendUpstream? It's not necessarily a valid fcnt in a sense that it incremented anymore, but we may still allow it to be sent upstream.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -260,20 +286,30 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
                         }
                     }
 
-                    if (!isConfirmedResubmit)
+                    var sendUpstream = concentratorDeduplicationResult is ConcentratorDeduplicationResult.NotDuplicate || loRaDevice.Deduplication is not DeduplicationMode.Drop;
```
p-schuler, line 289: can't we use the fcntup validation before to determine, if we want to send upstream?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -260,20 +286,30 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
-                        loRaDevice.SetFcntUp(payloadFcntAdjusted);
                     }
+
+                    loRaDevice.SetFcntUp(payloadFcntAdjusted);
+                }
+
+                #region Downstream
+                if (skipDownstreamToAvoidCollisions)
+                {
+                    this.logger.LogInformation($"skipping downstream messages due to deduplication ({timeWatcher.GetElapsedTime()})");
```
p-schuler, line 311: this should be a debug message. If we log this out, this can cause a lot of messages on devices with a lot of concentrators/devices.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: You mean add this test up when we check also the fcntup? We could do that but that would change the existing logic. Currently we are setting the `requiresConfirmation`, getting cloud to a device message (if not mistaken) and send class C messages (?), even if we skip sending to IoTHub.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Changed it in 76b18e9e0eca82c4d18079629de93bcd795090f7

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: 6b50b16630e5fd0d7bf083b28bf9c6af12311829

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: It was moved in `EnsureHasFcntDownAsync`: there we already check if we have a value for the fcntdown (0f11c3bce9a35e97b1cf3468a75b94202c6d06ef) 

Since `EnsureHasFcntDownAsync` is called from 3 other place, this change means we debug-log the frame counter down from those other places but I think a) it makes sense to have the same behavior in all these call sites b) that the logging is not too verbose. Let me know if you disagree.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IConcentratorDeduplication.cs
```diff
@@ -1,17 +1,51 @@
...
+        /// <summary>
+        /// Duplicate message due to resubmit of a confirmed
+        /// message of the same station (concentrator).
+        /// </summary>
+        DuplicateDueToResubmission,
+        /// <summary>
+        /// Detected as a duplicate but due to the DeduplicationStrategy,
+        /// marked only as a "soft" duplicate - allow upstream (Mark and None)
+        /// </summary>
+        SoftDuplicateDueToDeduplicationStrategy,
```
p-schuler, line 44: still getting confused by this name. How about simply naming this ConcentratorDuplicateAllowUpstream?

p-schuler: it's easier to understand when reading the places, where we use it, I think. It indicates, that when this result is set, we had a duplicate but it should still be allowed upstream.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
```diff
@@ -67,21 +67,14 @@ public WebSocketWriterRegistry(ILogger<WebSocketWriterRegistry<TKey, TMessage>>?
         /// </summary>
         public bool TryGetHandle(TKey key, [NotNullWhen(true)] out IWebSocketWriterHandle<TMessage>? handle)
         {
-            var keyFound = TryGetValue(key, out var value);
-            handle = value?.Handle;
-            return keyFound;
+            lock (this.sockets)
```
p-schuler, line 70: was this a bug? If so, should this be tracked separately?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
p-schuler: But wouldn't that mean, that we potentially log the value multiple times? Each time we call the Ensure method?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
p-schuler: what I meant is we have this to enter the upstream handling:

`var canSendUpstream = isFrameCounterFromNewlyStartedDevice
                                  || payloadFcntAdjusted > loRaDevice.FCntUp
                                  || (payloadFcntAdjusted == loRaDevice.FCntUp && concentratorDeduplicationResult is ConcentratorDeduplicationResult.SoftDuplicateDueToDeduplicationStrategy);`

wouldn't this be the indication already that we want to send upstream?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Yes we do

> this change means we debug-log the frame counter down from those other places too but I think a) it makes sense to have the same behavior in all these call sites b) that the logging is not too verbose.

Do you find this a problem (log level is debug anyway)?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
spygi: No, it was not a bug, only a refactoring happened here. See 7f8f9bfb54deb2a49028d1c5194ea0ada4db9cc1 for the full context.

Previously we had a helper private method called `TryGetValue` called from 2 points `IsSocketWriterOpen` and `IsSocketWriterOpen`. Since we decided to drop the connectivity check in the ConcentratorDeduplication, we can delete `IsSocketWriterOpen` which means the helper method `TryGetValue` is used from only one place so can be inlined.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IConcentratorDeduplication.cs
spygi: I don't think the deduplication should have knowledge about what will happen later in the RequestHandler. But since this keeps coming up, changed it to `DuplicateAllowUpstream` (dropped `Concentrator` from the name since the type already has it)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/WebSocketWriterRegistry.cs
p-schuler: Makes sense.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IConcentratorDeduplication.cs
p-schuler: Let's keep what it is as you mentioned, this would introduce knowledge of the handling in the request handler in here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
p-schuler: We can resolve this. This is needed to distinguish the case where we have a device reset or a resubmit and we still want to check the decoder result for downstream but not send upstream. The logic of the decoder validation should be extracted.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
p-schuler: as discussed. Let's make sure we don't emit the logs each time, but only when we really create it and have a valid counter.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
spygi: done d6f480766b705fe7d57b9d11f4d53d013d5b4ae9

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: In 406c7d48724d6b1fb7136b324450f5c98c5ee289 we log if we already have a valid counter down, otherwise we log it the first time we get one from the Function. (we rely on the fact that comparing with null will always be false - [there is a compiler warning for this](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0464))

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Has changed quite a bit since the comment and now is part of the test coverage - resolving this.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Resolving.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Created #1003 to keep track of this.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Done and covered with integration tests, resolving this.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Not sending downstream was addressed by adding flag `skipDownstreamToAvoidCollisions`

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IConcentratorDeduplication.cs
spygi: Reverted suggestion in 9449b06f5dd69f35c992338e889de84e3ed50e27

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: 0d923f9655c55d7aa762c6a6ba85fc717c2813a5

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -789,7 +883,9 @@ private async Task<bool> DetermineIfFramecounterIsFromNewlyStartedDeviceAsync(Lo
                         // known problem when device restarts, starts fcnt from zero
                         // We need to await this reset to avoid races on the server with deduplication and
                         // fcnt down calculations
-                        _ = await frameCounterStrategy.ResetAsync(loRaDevice, payloadFcnt, this.configuration.GatewayID);
+                        if (concentratorDeduplicationResult is ConcentratorDeduplicationResult.NotDuplicate)
+                            _ = await frameCounterStrategy.ResetAsync(loRaDevice, payloadFcnt, this.configuration.GatewayID);
+
                         isFrameCounterFromNewlyStartedDevice = true;
```
Mandur, line 889: Shouldn't this also be be in the previous if clause (concentrator?)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -120,18 +139,20 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
                     this.logger.LogDebug("ADR ack request received");
                 }
 
-                // ADR should be performed before the deduplication
-                // as we still want to collect the signal info, even if we drop
-                // it in the next step
-                if (loRaADRResult == null && loraPayload.IsDataRateNetworkControlled)
+                // ADR should be performed before the gateway deduplication as we still want to collect the signal info,
+                // even if we drop it in the next step.
+                // ADR is skipped for soft duplicates and will be enabled again in https://github.com/Azure/iotedge-lorawan-starterkit/issues/1017
```
Mandur, line 150: I am not even sure we should activate this for duplciates, maybe let's discuss the idea!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -156,30 +177,30 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
                     if (fcntDown <= 0)
                     {
                         return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.HandledByAnotherGateway);
                     }
                 }
+                #endregion
 
-                var validFcntUp = isFrameCounterFromNewlyStartedDevice || (payloadFcntAdjusted > loRaDevice.FCntUp);
-                if (validFcntUp || isConfirmedResubmit)
+                var canSendUpstream = isFrameCounterFromNewlyStartedDevice
```
Mandur, line 200: I think here we are not sending here upstream but only calling the decoders? There is another variable called sendUpstream below, what is the difference between the two?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -436,17 +465,17 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
                 return new LoRaDeviceRequestProcessResult(loRaDevice, request, confirmDownlinkMessageBuilderResp.DownlinkPktFwdMessage);
+                #endregion
             }
             finally
             {
-
                 try
                 {
-                    _ = await loRaDevice.SaveChangesAsync(force: stationEuiChanged);
+                    await SaveChangesToDevice(loRaDevice, stationEuiChanged);
```
Mandur, line 478: I think this method is still async therefore should have the async suffix

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: We can discuss in the linked item if ok for you. We also have an integration test covering exactly this case.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: > I think here we are not sending here upstream but only calling the decoders?

Correct. Patrick had a similar comment, the decoder logic could be potentially moved in the downstream part. I created #1074 for this since it's not a change introduced in this PR.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: It was not a programming error and there are 3 tests for unconfirmed first messages that fail if we add this in the previous if. However there might be a bug there: can we ever get a resubmission for an unconfirmed message? Only if it's a restarted device. But even then it would not send the exact same message, right?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Done bdb9238b59d1c2a28d85dd1c6f06e3ddb27e440b

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
p-schuler: we always need the decoder for the upstream but also for the downstream. I think the decoder call, will have to happen, if either, down or upstream is required. Because of the feature that the decoder could compose a C2D message.

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
```diff
@@ -65,14 +65,14 @@ public async Task AssertIoTHubDeviceMessageExistsAsync(string deviceID, string t
 
             if (assertionLevel == LogValidationAssertLevel.Error)
             {
-                var logs = string.Join("\n\t", searchResult.Logs.TakeLast(5));
+                var logs = string.Join("\n\t", searchResult.Logs.Select(x => x.ToString()).TakeLast(5));
```
p-schuler, line 68: is this related?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: sorry yes, I meant the particular logic for the C2D feature could be moved in the downstream part (in the story it's correct).

spygi: Will resolve this now but please re-open if you have further questions @Mandur

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
spygi: Discussed this with Patrick and he said that resubmission of unconfirmed message with frame counter 1 (first message) is ok to pass through a second time since we are using `IsABPRelaxedFrameCounter`

spygi: Added an explanation and curly braces to the if to make it more clear (fb4534a743c117c2f4f1216e939f6672c2577651)

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
spygi: No, reverted now 02caca3b8ec2d9e9effca4959a2208f7b83fe480 - was trying to get a more detailed log message but it's not needed anymore.

p-schuler (approved):
Once we log the issues we found (with the reply attach possibility) and the additional tests for OTAA, I'm good with this going in.

spygi:
> Once we log the issues we found (with the reply attach possibility) and the additional tests for OTAA, I'm good with this going in.

Added #1090 (tracked in the overarching feature) and #1091 (tracked in this bug) for these issues.

PR #944, opened 2021-12-01
Count processing errors and raise as metric

bastbu:
# PR for issue #747 

## What is being addressed

With this PR we start counting processing errors and raise them as metrics. This helps users that do not use Azure Monitor to still get insights into the number of errors occurred.

## How is this addressed

When logging (fallback) error messages, we increase the count for processing errors. In addition, we make sure to use the correct overload of the logger to ensure that the exceptions get raised properly in Application Insights

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
@@ -9,6 +9,7 @@ namespace LoRaWan.NetworkServer.BasicsStation.ModuleConnection
     using Microsoft.Extensions.Logging;
     using System;
     using System.Configuration;
+    using System.Diagnostics.Metrics;
     using System.Net;
     using System.Text.Json;
     using System.Threading;
@@ -21,6 +22,7 @@ namespace LoRaWan.NetworkServer.BasicsStation.ModuleConnection
         private readonly ILoRaDeviceRegistry loRaDeviceRegistry;
         private readonly LoRaDeviceAPIServiceBase loRaDeviceAPIService;
         private readonly ILogger<ModuleConnectionHost> logger;
+        private readonly Counter<int> processingErrorCounter;
         private ILoraModuleClient loRaModuleClient;
         private readonly ILoRaModuleClientFactory loRaModuleClientFactory;
 
@@ -30,7 +32,8 @@ namespace LoRaWan.NetworkServer.BasicsStation.ModuleConnection
             ILoRaModuleClientFactory loRaModuleClientFactory,
             ILoRaDeviceRegistry loRaDeviceRegistry,
             LoRaDeviceAPIServiceBase loRaDeviceAPIService,
-            ILogger<ModuleConnectionHost> logger)
+            ILogger<ModuleConnectionHost> logger,
+            Meter meter)
         {
             this.networkServerConfiguration = networkServerConfiguration ?? throw new ArgumentNullException(nameof(networkServerConfiguration));
             this.classCMessageSender = defaultClassCDevicesMessageSender ?? throw new ArgumentNullException(nameof(defaultClassCDevicesMessageSender));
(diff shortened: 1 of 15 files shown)


spygi:
There is the `LoRaProcessingException`, do we want to raise a metric when we throw one of these exceptions?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/ExceptionFilterUtility.cs
```diff
@@ -7,15 +7,23 @@ namespace LoRaWan
...
         }
 
-        public static bool False(Action action)
+        public static bool False(params Action[] actions)
         {
-            (action ?? throw new ArgumentNullException(nameof(action)))();
+            if (actions is null) throw new ArgumentNullException(nameof(actions));
+
+            foreach (var a in actions)
+                a();
```
spygi, line 25: If one of the 2 actions throws due to a programming mistake, should we make sure to execute the other one?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -156,7 +161,8 @@ internal Task OnDesiredPropertiesUpdate(TwinCollection desiredProperties, object
             {
                 this.logger.LogWarning($"A desired properties update was detected but the parameters are out of range with exception :  {ex}");
             }
-            catch (Exception ex) when (ExceptionFilterUtility.False(() => this.logger.LogError($"An exception occurred on desired property update: {ex}")))
+            catch (Exception ex) when (ExceptionFilterUtility.False(() => this.logger.LogError(ex, $"An exception occurred on desired property update: {ex}"),
+                                                                    () => this.processingErrorCounter?.Add(1)))
```
spygi, line 165: Is the safe navigation still required? Because I see in the tests you pass a `TestMeter` instance

spygi (approved):
LGTM, I think there are additional places we might want to increment the counter but I see this as the starting point which we can expand on.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
```diff
@@ -88,7 +91,11 @@ internal enum LoaderState
                     continuationAction(t, this);
                 }
             },
-            ex => this.logger.LogError($"Error while loading: {ex}."));
+            ex =>
+            {
+                this.logger.LogError($"Error while loading: {ex}.");
+                processingErrorCount?.Add(1);
```
spygi, line 97: [nit] What if we add this in the TaskUtil.RunOnThreadPool? This way we can avoid adding this in all the callers of it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/ExceptionFilterUtility.cs
bastbu: While I do agree that it would be more safe, I think in this case (the `ExceptionFilterUtility` has a very distinct use - side effects in `when` clauses) the additional complexity from aggregating the exceptions etc is not worth it. We should make very sure that we don't mask the original exception by generating other exceptions in the `True` and `False` arguments.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
bastbu: Removed with 567bfc030f0a954c1170638a56e74b26e6c2f964

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs
bastbu: Valid point, I changed the signature of the method to require passing a `Counter<int>`. This will make it explicit to any caller that exceptions are counted - in case of observed executions it's still up to the caller to decide whether to not pass the counter if the task is awaited.

spygi: Makes sense to have the counter nullable 👍

PR #948, opened 2021-12-01
Enable Gateway On the Ci

Mandur:
Enable edgeHub usage on the CI with the most recent fixes


danigian:
@Mandur, now that #976 is merged, what about testing again?

p-schuler (approved):
I'm surprised that NestedEdgeEnabled is default true. If we do not have a nested hierarchy I would assume that this would be disabled.

Mandur:
> I'm surprised that NestedEdgeEnabled is default true. If we do not have a nested hierarchy I would assume that this would be disabled.

This is the new way of the IoT Edge authentication... Therefore it is the new default... @p-schuler

Issue #923, opened 2021-11-29
ADR upstream logic is verified in integration tests ( function tests should verify that after 65 messages a Link ADR Request is sent back)


Mandur:

MaggieSalak:
After adapting the E2E test, the logic to adapt the data rate automatically in the LNS would be untested. This should be covered with integration tests.

Issue #922, opened 2021-11-29
Add a new test to send a LinkADRRequest and verify the rate adaptation is done correctly executed on fullci


Mandur:

MaggieSalak:
Instead of sending ~60 messages from the leaf device, send LinkADRRequest as C2D message and verify that the DR is changed.

Issue #921, opened 2021-11-29
Add option to send LinkADRRequest as C2D messages


Mandur:

Mandur:
started in feature/add-adr-link

MaggieSalak:
Error checks, unit tests missing.

PR #949, opened 2021-12-01
Feature/921 Add option to send LinkADRReqest as C2D message

MaggieSalak:
# PR for issue #921 

## What is being addressed

- `LinkADRReq` MAC command was not being handled correctly in the LNS and would never reach the leaf device

## How is this addressed

- `LinkADRReq` is correctly parsed from a MAC command
- Validation of required properties of the command added
- Unit tests added to cover `LinkADRReq` command parsing and error handling

## Out of scope

- E2E test verifying data rate change following a LinkADRReq MAC command - this is tracked in #922 
- Integration tests verifying the automatic DR change after certain number of messages - tracked in #923

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MACCommandJsonConverter.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MACCommandJsonConverter.cs
@@ -4,8 +4,6 @@
 namespace LoRaTools
 {
     using System;
-    using System.Collections.Generic;
-    using System.Globalization;
     using Newtonsoft.Json;
     using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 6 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MACCommandJsonConverter.cs
```diff
@@ -73,16 +75,39 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                     case Cid.One:
                     case Cid.LinkCheckCmd:
                     case Cid.LinkADRCmd:
+                    {
+                        if (!IsValidLindADRCmd(item, out var errors))
```
Mandur, line 79: Instead of this validation method why don't we use tryget and fail if it doesn't work?

MaggieSalak: Done in e9bd89bcf384c5592154f7eb6672c7af465c8264.

Tests/Unit/LoRaWan/MacCommandTest.cs
```diff
@@ -33,5 +34,29 @@ public void When_Serializing_Single_Object_Should_Create_Correct_Items()
...
+            var linkADRCmd = (LinkADRRequest)macCommand;
+            Assert.Equal(2, linkADRCmd.DataRate);
+            Assert.Equal(4, linkADRCmd.TxPower);
+            Assert.Equal(25, linkADRCmd.ChMask);
+            Assert.Equal(0, linkADRCmd.ChMaskCntl);
+            Assert.Equal(1, linkADRCmd.NbRep);
+        }
+
+        [Fact]
+        public void When_Serializing_Invalid_LinkAdrCmd_Should_Throw()
```
Mandur, line 54: Would be nice to change this to a theory to test multiple bad cases

Mandur (approved):

PR #954, opened 2021-12-02
Feature/740 iot hub logger

bastbu:
# PR for issue #740

## What is being addressed

With this PR we bring back the logger that was removed in #788 as DI logger.


LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("LoRaWan.Tests.Unit")]
```
spygi, line 8: [nit] can be moved in the `AssemblyInfo.cs`

Tests/Unit/Logger/IoTHubLoggerTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        private static DisposableValue<Mock<IotHubLogger>> SetupProviderAndLogger(IotHubLoggerConfiguration configuration)
+        {
+            var provider = new IotHubLoggerProvider(configuration, new Lazy<Task<ModuleClient>>((Task<ModuleClient>)null!));
+            return new DisposableValue<Mock<IotHubLogger>>(new Mock<IotHubLogger>(provider, null), provider);
+        }
+
+        private static async Task VerifyMessageAsync(Mock<IotHubLogger> logger, string message)
+        {
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed (false positive, verification only)
+            await logger.RetryVerifyAsync(x => x.SendAsync(message), Times.Once);
```
spygi, line 109: Here we are testing that our method `SendAsync` was called but not whether the SDK was actually used to send a message. If someone would remove the call to IoTHub these test would still pass. 

Not sure why we need a mock of the whole SUT. Can't we pass a mock only for the "moduleClientFactory" (instead of `null`) and verify calls directly on it?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    internal sealed class IotHubLoggerProvider : ILoggerProvider
+    {
+        private readonly ConcurrentDictionary<string, IotHubLogger> loggers = new();
+        private readonly Lazy<Task<ModuleClient>> moduleClientFactory;
+
+        public IotHubLoggerConfiguration Configuration { get; }
+        public LoggerExternalScopeProvider? ScopeProvider { get; }
+
+        public IotHubLoggerProvider(IotHubLoggerConfiguration configuration)
```
p-schuler, line 33: this setup won't support updates to the log level without re-starting the box. The ConsoleLogger has support for dynamic updates, I think we should do this here as well.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    try
+                    {
+                        var formattedMessage = LoggerHelper.AddScopeInformation(this.iotHubLoggerProvider.ScopeProvider, formatter(state, exception));
+                        await SendAsync(formattedMessage);
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.WriteLine($"Error when sending log to IoT Hub: {ex}");
+                        throw;
```
p-schuler, line 91: this is not doing much, right? The throw will go un-noticed. Not sure we need to re-throw here? Depending on the error, would we ever want to stop trying to send the messages? Maybe we can continue to send when we encounter an IoT hub exception that we consider transient and stop logging for other exceptions? Seems like a pretty expensive path, if we continuously fail?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                });
+            }
+        }
+
+        internal virtual async Task SendAsync(string message)
```
p-schuler, line 97: we are not using a CancellationToken here, as it is fire and forget. I think we could abort any pending requests, when we dispose the Logger (shutting down). Might be too much complexity?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
```diff
@@ -0,0 +1,114 @@
...
+    internal sealed class IotHubLoggerProvider : ILoggerProvider
+    {
+        private readonly ConcurrentDictionary<string, IotHubLogger> loggers = new();
+        private readonly Lazy<Task<ModuleClient>> moduleClientFactory;
+
+        public IotHubLoggerConfiguration Configuration { get; }
+        public LoggerExternalScopeProvider? ScopeProvider { get; }
+
+        public IotHubLoggerProvider(IotHubLoggerConfiguration configuration)
+            : this(configuration, new Lazy<Task<ModuleClient>>(ModuleClient.CreateFromEnvironmentAsync(new[] { new AmqpTransportSettings(TransportType.Amqp_Tcp_Only) })))
```
spygi, line 34: Not sure why we need the lazy initialization here? Is it if the constructor gets called multiple times? That would be only in the case of a programming mistake because the DI should ensure the constructor is called only once, right?

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
spygi, line 33: [nit] Could be set to `internal`.

spygi (changes requested):
Looks good, one thing that should be changed imo is verifying that we actually send messages to IoTHub (see comment on `IoTHubLoggerTests`) unless it's too complex

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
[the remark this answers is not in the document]
p-schuler: See my comment above. However, the exception will not bring down the application - a good article: https://devblogs.microsoft.com/pfxteam/tasks-and-unhandled-exceptions/

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
p-schuler: I think this is done to only create it, if we need to deliver a message. If the log level is not met, we will never create the instance - for example, if we only want to log errors, it won't get created unless we log an exception.

Tests/Unit/Logger/IoTHubLoggerTests.cs
bastbu: We can't mock the `ModuleClient`, since even though it is non-sealed, it does not have any `virtual` members that we could override/mock. We could discuss if we want to hide external components in a thin wrapper, but that would not solve the problem that the closest layer to the SDK is not unit-testable.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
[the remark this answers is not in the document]
bastbu: Let's track this thread where Patrick had the comment, will resolve this thread.

bastbu: Good point about the "permanently broken" logger, I will see what exceptions I can provoke and when I can "disable" the logger.

I don't think we should hide the exception here, though:
- It will generate a CA warning that we need to suppress
- It should be up to the caller to decide how to handle unhandled exceptions. If we ever decide to make use of something like `TaskScheduler.UnobservedTaskException`, we will not get these exceptions without knowing that we hide the exceptions here

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
bastbu: Since we can't use that feature, yet, I created a story for it: https://github.com/Azure/iotedge-lorawan-starterkit/issues/989. Will prepare for the story and implement what you suggested

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
bastbu: I had two reasons in mind here: Since the module instantiation is asynchronous, I would need to take a `Task<T>` or `Func<Task<T>>` as a factory (at least), but I can't wait for the initialization to finish in the constructor. Since this is already additional complexity, I decided to use `Lazy<T>` instead, as it gives us the benefit of only executing when invoked.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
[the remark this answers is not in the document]
spygi: > However, the exception will not bring down the application

No, not as the code is right now. I was saying that would be the case if we `await` the `Task.Run`

spygi (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
bastbu: I think in any case it might be worth to try to finish sending the log message. I agree that it may not be worth the complexity.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
bastbu: Addressed this in 4250a35c1e15a4fdf7edbae7525cc842b1ba6c73

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
bastbu: Did the first part in 4865a49dedb2289bbfb785cd1fdf9f847286807f

LoRaEngine/modules/LoRaWanNetworkSrvModule/Logger/IotHubLogger.cs
p-schuler: Agreed.

Issue #878, opened 2021-11-23
Add E2E test for verifying new CUPS scenarios


danigian:
- [x] #956
- [x] #994
- [x] Remove CI BasicStation startup script
- [x] #1030
- [x] #1063

~~- [ ] Add E2E Test for verifying credential rotation scenario~~ not useful anymore, since the CUPS test is actually rotating the certificates already

Issue #877, opened 2021-11-23
Provide clear documentation on supported authentication modes


danigian:
- [ ] Update https://azure.github.io/iotedge-lorawan-starterkit/dev/user-guide/station-credential-management/ with client certificate instructions
- [ ] Add documentation on how to generate certificates using CLI
- [ ] Add documentation on how to provision stations with the Device Provisioning CLI

PR #958, opened 2021-12-02
Feature/956 aio should use server auth

### LINKED ISSUE #956: Prepare E2E CI for using server certificates
Following setup should be reached with CI:

- One RPi is connecting to LNS on EFLOW without any authentication mode (already there)
- One RPi (named "AIO") is running LNS with server certificate and client certificate mode set to "Allow". Same RPi should run the concentrator module with the trust file for server auth only
- Concentrator on the CI should be enabled only for needed tests (namely multi concentrator tests and CUPS tests). It should make use of both server and client authentication to cover the last possible authentication scenario

Acceptance criteria for this task:
- [x] A new stage is added in the e2e-ci.yml for:
   - [x] Invoking certificate generation script and creating a root and a server certificate
   - [x] Copying with scp the required .pfx and .trust files to target RPi
- [x] The aio template is modified in such a way that importing .pfx and .trust is done

danigian:
# PR for issue #878

This PR is introducing a "certificates_job" in the E2E CI which is generating and copying over SSH to the target raspberry root and server certificates for enabling "server auth only" scenario testing.

The "aio" template is adjusted to let the Basic Station point to wss endpoint and let the NetworkServer start while importing a server certificate

Below a screenshot with logs from Basics Station module connecting with "server auth only" option:
image


.github/workflows/e2e-ci.yaml
```diff
@@ -182,6 +183,37 @@ jobs:
...
+      name: Generate server and client certificates
+      env:
+        SERVER_PFX_PASSWORD: ${{ secrets.SERVER_PFX_PASSWORD }}
+        SSH_PRIVATE_KEY_PATH: ${{ secrets.SSH_PRIVATE_KEY_PATH }}
+        RPI_USERNAME: ${{ secrets.RPI_USERNAME }}
+        RPI_HOST: ${{ secrets.RPI_HOST }}
+        CERT_REMOTE_PATH: ${{ needs.env_var.outputs.CertRemotePath }}
+      run: |
+        chmod +x ./Tools/BasicStation-Certificates-Generation/certificate-generate.sh
+        ./Tools/BasicStation-Certificates-Generation/certificate-generate.sh server $RPI_HOST $SERVER_PFX_PASSWORD
```
danigian, line 212: We are every time generating certificates here, this means that the module will be required to restart when the deployment is happening.
I have a doubt here, is it already happening that every time we deploy to RPi the IoT Edge deployment, the modules are restarting? @Mandur

.github/workflows/e2e-ci.yaml
```diff
@@ -85,6 +85,7 @@ jobs:
       E2ETestsToRun: ${{ steps.e2e_list_step.outputs.E2ETestsToRun }}
       StopFullCi: ${{ steps.check-if-run.outputs.StopFullCi }}
       TxPower: ${{ steps.set_txpower.outputs.TxPower }}
+      CertRemotePath: "/var/lorastarterkit/certs"
```
bastbu, line 88: Why define this as an output and instead of of defining it as a variable on `env` (less indirection)?

.github/workflows/e2e-ci.yaml
```diff
@@ -182,6 +183,37 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+      name: Checkout current branch
+
+    - id: generate_step
+      shell: bash
+      name: Generate server and client certificates
+      env:
+        SERVER_PFX_PASSWORD: ${{ secrets.SERVER_PFX_PASSWORD }}
+        SSH_PRIVATE_KEY_PATH: ${{ secrets.SSH_PRIVATE_KEY_PATH }}
```
bastbu, line 206: Should we add a brief description what this secret is? (I assume we need the cert to be present on the Github runner machine to connect to the Pi that we use for E2E tests? (same goes for the other secrets that were introduced, such as `RPI_USERNAME`)

LoRaEngine/deployment.ci.aio.template.json
```diff
@@ -129,13 +147,16 @@
...
+                "value": "wss://localhost:5001"
               },
               "RESET_PIN": {
                 "value": "$RESET_PIN"
               },
               "FIXED_STATION_EUI": {
                 "value": "$LBS_FIXED_STATION_EUI"
+              },
+              "TLS_SNI": {
+                "value": "false"
```
bastbu, line 159: If we use `localhost` as the CN for the server certificate and remove this env variable, wouldn't this work as well?

LoRaEngine/deployment.ci.aio.template.json
```diff
@@ -112,7 +127,10 @@
               "createOptions": {
                 "HostConfig": {
                   "NetworkMode": "host",
-                  "Privileged": true
+                  "Privileged": true,
+                  "Binds":  [
+                    "$CERT_REMOTE_PATH:/var/lorastarterkit/certs/"
```
bastbu, line 132: [nit] not sure if it's worth it, but to me it was not immediately clear how the LBS picks this up. Maybe it would be worth a comment that the location needs to match what we have in `LoRaEngine\modules\LoRaBasicsStationModule\start_basestation_amd64.sh`?

bastbu (approved):

.github/workflows/e2e-ci.yaml
Mandur: We should probably (as you suggested @bastbu) start documenting the CI and the different variable. This is really starting to become quite complex

Mandur: Created #1015 to follow up on that

.github/workflows/e2e-ci.yaml
Mandur: Yes it does

.github/workflows/e2e-ci.yaml
```diff
@@ -182,6 +183,37 @@ jobs:
...
+        RPI_USERNAME: ${{ secrets.RPI_USERNAME }}
+        RPI_HOST: ${{ secrets.RPI_HOST }}
+        CERT_REMOTE_PATH: ${{ needs.env_var.outputs.CertRemotePath }}
+      run: |
+        chmod +x ./Tools/BasicStation-Certificates-Generation/certificate-generate.sh
```
Mandur, line 211: QQ can't we commit it directly with the chmod +x instead of running it here?

LoRaEngine/deployment.ci.aio.template.json
```diff
@@ -68,13 +68,22 @@
...
+                  ],
                   "PortBindings": {
-                    "5000/tcp": [
+                    "5001/tcp": [
                       {
-                        "HostPort": "5000"
+                        "HostPort": "5001"
+                      }
+                    ],
+                    "5002/tcp": [
```
Mandur, line 84: What is 5002 for?

LoRaEngine/deployment.ci.aio.template.json
```diff
@@ -68,13 +68,22 @@
...
+                  "5001/tcp": {},
+                  "5002/tcp": {}
                 },
                 "HostConfig": {
+                  "Binds":  [
+                    "$CERT_REMOTE_PATH:/var/lorastarterkit/certs/"
+                  ],
                   "PortBindings": {
-                    "5000/tcp": [
+                    "5001/tcp": [
```
Mandur, line 79: @bastbu Do we need to change something on the metrics collector? I remember this listen on 5000? Would we also need to generate a cert for it?

Mandur (approved):

LoRaEngine/deployment.ci.aio.template.json
danigian: It's already documented here https://azure.github.io/iotedge-lorawan-starterkit/dev/user-guide/station-credential-management/#server-authentication

I will add client authentication documentation with #877

LoRaEngine/deployment.ci.aio.template.json
danigian: CUPS

.github/workflows/e2e-ci.yaml
danigian: Adding a variable on env would require more setup, which is actually not strictly required here. I will resolve this and we can consider a refactoring in the future

.github/workflows/e2e-ci.yaml
danigian: Yes we could, but apparently modifying it on windows (as all our dev machines) could make the executable bit disappear. Therefore, I will just leave the chmod +x here

LoRaEngine/deployment.ci.aio.template.json
danigian: #1025 has been created for checking the behavior

PR #996, opened 2021-12-06
Revert "Bump Microsoft.Azure.Devices from 1.33.0 to 1.36.0 (#986)"

danigian:
This reverts commit 3e2671520d14fe44503e102f17adda06c1803899 as it's causing some timeouts

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,7 @@
     </PropertyGroup>
     <PropertyGroup>
       <DockerDotNetX509Version>3.125.2</DockerDotNetX509Version>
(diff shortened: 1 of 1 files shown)


Mandur:
I ran three different run on dev and all the three did not present the issue we saw with this sdk version.... Both seems to have failed because of transient errors... This is very strange... I would suggest to abort this revert. WDYT?

https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1549651506
https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1549160002
https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1548869895

Issue #994, opened 2021-12-06
Extend LoRaBasicsStationModule capabilities to support client auth and CUPS


danigian:
The LoRaBasicsStationModule should be expanded to allow import of Client certificates for TC/CUPS endpoints and also a CUPS_URI field

PR #997, opened 2021-12-06
Extend LoRaBasicsStationModule capabilities to support client auth and CUPS

danigian:
# PR for issue #994 

## What is being addressed

Extending LoRaBasicsStationModule capabilities to support client auth certificate imports and CUPS

### DIFF
--- a/LoRaEngine/modules/LoRaBasicsStationModule/helper-functions.sh
+++ b/LoRaEngine/modules/LoRaBasicsStationModule/helper-functions.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+
+DEFAULT_CERTS_PATH="/var/lorastarterkit/certs"
+DEFAULT_TC_TRUST_PATH="$DEFAULT_CERTS_PATH/tc.trust"
+DEFAULT_TC_CRT_PATH="$DEFAULT_CERTS_PATH/tc.crt"
+DEFAULT_TC_KEY_PATH="$DEFAULT_CERTS_PATH/tc.key"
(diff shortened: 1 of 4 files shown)


LoRaEngine/modules/LoRaBasicsStationModule/helper-functions.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
Mandur, line 1: I guess you also have to copy this file in the basicstation dockerfile?

LoRaEngine/modules/LoRaBasicsStationModule/helper-functions.sh
```diff
@@ -0,0 +1,82 @@
...
+        fi
+    else
+        cp -v $1 tc.trust
+    fi
+}
+
+tcCertCopy() {
+    if [[ "$TC_URI" == *"wss"* ]]; then
+        echo "A secure protocol was specified for LNS endpoint. Copying over certificate files".
+        conditionalCopy "$TC_TRUST_PATH" "$DEFAULT_TC_TRUST_PATH"
```
Mandur, line 28: Should we check as this point if TC_TRUST_PATH etc... is set and error out?

Mandur: Oh just saw you do it in the conditional copy, even smarter

Mandur (approved):

LoRaEngine/modules/LoRaBasicsStationModule/helper-functions.sh
danigian: Done, thanks!

Issue #995, opened 2021-12-06
Gateway ID should be raised with Metrics Custom Dimensions


bastbu:
## Expected Behavior
Gateway ID should be raised with Metrics Custom Dimensions

## Current behavior

Only concentrator IDs are raised with Custom Dimensions

PR #999, opened 2021-12-06
Bug/995 Gateway ID Tags

bastbu:
# PR for issue #995

## What is being addressed

With this PR we add the Gateway ID to all custom metrics that we raise. To support easier extensibility for (potential) future tags from the tag bag, we first refactor the logic to fetch tags in order to be more efficient by creating less allocations.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApplicationInsightsMetricExporter.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApplicationInsightsMetricExporter.cs
@@ -60,7 +60,7 @@ namespace LoRaWan.NetworkServer
                 case 2: _ = metric.TrackValue(measurement, dimensions[0], dimensions[1]); break;
                 case 3: _ = metric.TrackValue(measurement, dimensions[0], dimensions[1], dimensions[2]); break;
                 case 4: _ = metric.TrackValue(measurement, dimensions[0], dimensions[1], dimensions[2], dimensions[3]); break;
-                default: throw new NotImplementedException("Metrics tracking in Application Insights for more than three dimensions it not supported");
+                default: throw new NotImplementedException("We do not support tracking more than four custom dimensions.");
             }
         }
     }
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApplicationInsightsMetricExporter.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApplicationInsightsMetricExporter.cs
@@ -19,10 +19,12 @@ namespace LoRaWan.NetworkServer
     /// </summary>
     internal class ApplicationInsightsMetricExporter : RegistryMetricExporter
     {
-        private readonly IDictionary<string, (Metric, MetricIdentifier)> metricRegistry;
+        private readonly IDictionary<string, (Metric, string[])> metricRegistry;
         private readonly RegistryMetricTagBag metricTagBag;
 
-        public ApplicationInsightsMetricExporter(TelemetryClient telemetryClient, RegistryMetricTagBag metricTagBag, ILogger<ApplicationInsightsMetricExporter> logger)
+        public ApplicationInsightsMetricExporter(TelemetryClient telemetryClient,
+                                                 RegistryMetricTagBag metricTagBag,
+                                                 ILogger<ApplicationInsightsMetricExporter> logger)
             : this(telemetryClient, MetricRegistry.RegistryLookup, metricTagBag, logger)
         { }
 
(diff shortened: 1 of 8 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApplicationInsightsMetricExporter.cs
```diff
@@ -35,7 +37,7 @@ public ApplicationInsightsMetricExporter(TelemetryClient telemetryClient, Regist
             this.metricRegistry = registryLookup.ToDictionary(m => m.Key, m =>
             {
                 var id = new MetricIdentifier(MetricRegistry.Namespace, m.Value.Name, m.Value.Tags);
-                return (telemetryClient.GetMetric(id), id);
+                return (telemetryClient.GetMetric(id), id.GetDimensionNames().ToArray());
```
bastbu, line 40: unrelated change, but fixed as part of this PR since I touched the code in any case

Tests/Unit/NetworkServer/ApplicationInsightsMetricExporterTests.cs
```diff
@@ -179,22 +179,6 @@ public void When_Raising_Metric_And_Missing_Dimensions_Should_Fail()
             Assert.Throws<LoRaProcessingException>(() => counter.Add(1));
         }
 
-        [Fact]
-        public void When_Raising_Metric_Dimensions_Should_Be_Case_Insensitive()
```
bastbu, line 183: Case-sensitivity is not needed and was only inconsistently applied

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApplicationInsightsMetricExporter.cs
```diff
@@ -58,6 +59,7 @@ internal virtual void TrackValue(Metric metric, double measurement, params strin
                 case 1: _ = metric.TrackValue(measurement, dimensions[0]); break;
                 case 2: _ = metric.TrackValue(measurement, dimensions[0], dimensions[1]); break;
                 case 3: _ = metric.TrackValue(measurement, dimensions[0], dimensions[1], dimensions[2]); break;
+                case 4: _ = metric.TrackValue(measurement, dimensions[0], dimensions[1], dimensions[2], dimensions[3]); break;
```
p-schuler, line 62: What's that limitation we are talking about here? I see that the TelemetryClient supports up to 7 dimensions?

https://docs.microsoft.com/en-us/dotnet/api/microsoft.applicationinsights.metric.trackvalue?view=azure-dotnet#Microsoft_ApplicationInsights_Metric_TrackValue_System_Double_System_String_System_String_System_String_System_String_System_String_System_String_System_String_

bastbu: Yes, that's correct. However, we don't need seven dimensions, we use max three dimensions as of now.

p-schuler: That's ok. However, the exception suggests that it is not supported in App Insights. Can we make that clear?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MetricRegistry.cs
```diff
@@ -109,24 +116,42 @@ internal static class MetricExporterHelper
         /// Gets the tags ordered by the original order of the tags.
         /// Falls back to common tags that are present in the RegistryMetricTagBag.
         /// </summary>
-        public static string[] GetTagsInOrder(IEnumerable<string> tagNames, ReadOnlySpan<KeyValuePair<string, object?>> tags, RegistryMetricTagBag metricTagBag)
+        public static string[] GetTagsInOrder(IReadOnlyList<string> tagNames, ReadOnlySpan<KeyValuePair<string, object?>> tags, RegistryMetricTagBag metricTagBag)
```
p-schuler, line 119: if we provide more values than initially dimensions specified, wouldn't we simply drop the value?

p-schuler (approved):
The validation of the values to definitions might be a good addition. Otherwise, LGTM

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ApplicationInsightsMetricExporter.cs
bastbu: Done in a8a99e505562fb6b46612b1a9ea4f3a0cdb34194

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MetricRegistry.cs
bastbu: Done in 8a74c1eb66044ec1ea885e1cb2568aad5e5e32bf

PR #1001, opened 2021-12-06
Reuse `FrameControl` & flags

atifaziz:
# PR for issue #965

## What is being addressed

Opportunities to reuse `FrameControl` and `FCtrlFlags` to avoid duplication.

## How is this addressed

- `FCtrl` has been removed as this was redundant with `FCtrlFlags`.
- Entire code base has been adjusted to use `FrameControl` and `FCtrlFlags`.
- `FrameControl` has been:
  - enhanced with Boolean properties (after evaluating uses) to make bit manipulation (setting & resetting bits) easier.
  - modified to store flags separate from options length and converts from/to a LoRaWAN-based byte when needed.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadDataLns.cs
```diff
@@ -51,20 +52,18 @@ MacMessageType.JoinAccept or
...
+            FrameControlFlags = fctrlFlags;
 
             // Setting Fcnt
             Fcnt = new byte[sizeof(ushort)];
             BinaryPrimitives.WriteUInt16LittleEndian(Fcnt.Span, counter);
 
             // Setting FOpts
-            var foptsSize = control.OptionsLength;
-            Fopts = new byte[foptsSize];
+            Fopts = new byte[options.Length / 2];
```
bastbu, line 62: When I checked I did not see this logic asserted in unit tests. I can't see how the division by 2 comes in, could you explain?

bastbu (approved):
LGTM. We could think about running the full ci for this PR, up to you whether we want to do this or not.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadDataLns.cs
atifaziz: `options` is a hexdecimal-encoded string therefore the byte length will be half of the string length.

atifaziz: Created issue #1014 to track the lack of tests for `LoRaPayloadDataLns`.

PR #1002, opened 2021-12-07
[Bug] Raise receive window hit metric in DefaultLoRaDataRequestHandler

bastbu:
## What is being addressed

With this PR we start raising a metric when a receive window is hit that went previously unnoticed.


Mandur:
good spot

Mandur (approved):

Issue #898, opened 2021-11-25
Code Analysis Warnings not treated as errors in provisioning tool.


bastbu:
## Expected Behavior

Code analysis warnings should be treated as errors in the provisioning CLI.

## Current Behavior

Code analysis warnings are not treated as errors in the provisioning CLI.

PR #1004, opened 2021-12-07
Treat Code Analysis warnings as errors in CLI project

bastbu:
# PR for issue #898

## What is being addressed

With this PR we fix existing CA/IDE warnings for the CLI project and treat CA/IDE warnings as errors.

## How is this addressed

Warnings that have many occurrences are addressed on a warning-per-commit basis, while warnings that only have one or two occurrences per CA/IDE warning are addressed in bulk.

### DIFF
--- a/Tools/Cli-LoRa-Device-Provisioning/Cli-LoRa-Device-Provisioning.csproj
+++ b/Tools/Cli-LoRa-Device-Provisioning/Cli-LoRa-Device-Provisioning.csproj
@@ -5,9 +5,6 @@
     <TargetFramework>$(TargetFramework)</TargetFramework>
     <RootNamespace>LoRaWan.Tools.CLI</RootNamespace>
     <AssemblyName>loradeviceprovisioning</AssemblyName>
-    <CodeAnalysisTreatWarningsAsErrors>false</CodeAnalysisTreatWarningsAsErrors>
-    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
-    <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
     <Authors>Sascha Corti (saschac@microsoft.com)</Authors>
     <Version>2.0.0-alpha</Version>
     <Company>Microsoft</Company>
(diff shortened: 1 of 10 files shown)


Tools/Cli-LoRa-Device-Provisioning/Helpers/IoTDeviceHelper.cs
```diff
@@ -1353,5 +1332,29 @@ public async Task<bool> RemoveDevice(string devEui, ConfigurationHelper configur
             Console.WriteLine("done.");
             return true;
         }
+
+        private static async Task<bool> ExecuteWithIotHubErrorHandlingAsync(Func<Task> executeAsync, string errorMessageContext = null)
```
spygi, line 1333: [nit] do you think this makes sense to keep as a separate method, given that this is referenced only once?

spygi (approved):
LGTM! Thanks for the atomic commits, made it easier to review this.

Tools/Cli-LoRa-Device-Provisioning/Cli-LoRa-Device-Provisioning.csproj
```diff
@@ -5,9 +5,6 @@
     <TargetFramework>$(TargetFramework)</TargetFramework>
     <RootNamespace>LoRaWan.Tools.CLI</RootNamespace>
     <AssemblyName>loradeviceprovisioning</AssemblyName>
-    <CodeAnalysisTreatWarningsAsErrors>false</CodeAnalysisTreatWarningsAsErrors>
-    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
-    <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
```
spygi, line 10: If not mistaken this [is disabled](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enforcecodestyleinbuild) by default, don't we want to enable it?

bastbu: This is automatically picked up from the `Directory.Build.props` that we have in the solution.

Tools/Cli-LoRa-Device-Provisioning/Helpers/IoTDeviceHelper.cs
bastbu: Done in 77aeb36e4dc075aec50b2f18c5c7b254f086a76f

PR #1011, opened 2021-12-07
Remove alert from ADR

bastbu:
## What is being addressed

We will not raise the alert `HighUpstreamMessagesLostRatio` as we can either define a static threshold for the metric, or create a separate metric (which is hard to do in our code base) to track when a message is lost. Since all lost messages generate exceptions, we can also observe the exception metric to detect when we lose messages.


p-schuler (approved):
LGTM

PR #1020, opened 2021-12-08
Use `Hertz` wherever a frequency is implied

atifaziz:
# PR for issue #967

## What is being addressed

Mixed use of integral types like `int` and `double` to represent frequencies at different metric scales.

## How is this addressed

Replace with `Hertz`.

## More Information

What's changed in the primitives:

- New `Mega` type that allows strong expression of _millions of some unit_ (that is, stronger than `double`), which is primarily used to represent MHz. It has the following members:
  - `ToString` overloads for seamless string formatting support
  - `Value` that returns the original `double`
  - `Units` returns the value in units, e.g. `new Mega(1.234e6)` &rarr; `1234000`
- `Hertz` members like `Kilo`, `Mega` and `Giga` have been renamed to `InKilo`, `InMega` and `InGiga`.
- `Hertz` instances can now be compared together.
- `Hertz` instances can be subtracted from each to discover the offset from one to the other.
- `Hertz` instances support addition of `long` and `Mega` offsets.
- `Hertz` instances support _explicit_ conversion to `ulong` (should be used very rarely).
- `Hertz` instances can be _implicitly_ created from `Mega`.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
```diff
@@ -86,7 +86,7 @@ private string Message(DownlinkPktFwdMessage message, Region region)
                 writer.WriteNumber("RX1DR", dataRate);
                 writer.WriteNumber("RX1Freq", (ulong)(message.Txpk.Freq * 1e6));
                 writer.WriteNumber("RX2DR", region.GetDefaultRX2ReceiveWindow().DataRate);
-                writer.WriteNumber("RX2Freq", (ulong)(region.GetDefaultRX2ReceiveWindow().Frequency * 1e6));
+                writer.WriteNumber("RX2Freq", (ulong)region.GetDefaultRX2ReceiveWindow().Frequency);
```
MaggieSalak, line 89: Instead of casting to `ulong`, could we use `AsUInt64` here?

atifaziz: What do you see as the difference/benefit of one over the other. To me, the cast is as explicit as having to use a property.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470RP1.cs
```diff
@@ -66,20 +68,17 @@ public RegionCN470RP1()
...
         /// </summary>
-        public override bool TryGetDownstreamChannelFrequency(double upstreamFrequency, out double downstreamFrequency, ushort? upstreamDataRate = null, DeviceJoinInfo deviceJoinInfo = null)
+        public override bool TryGetDownstreamChannelFrequency(Hertz upstreamFrequency, out Hertz downstreamFrequency, ushort? upstreamDataRate = null, DeviceJoinInfo deviceJoinInfo = null)
         {
             if (!IsValidUpstreamFrequency(upstreamFrequency))
                 throw new LoRaProcessingException($"Invalid upstream frequency {upstreamFrequency}", LoRaProcessingErrorCode.InvalidFrequency);
 
             var upstreamChannelNumber = (int)Math.Round(
-                (upstreamFrequency - StartingUpstreamFrequency) / FrequencyIncrement,
+                (upstreamFrequency - StartingUpstreamFrequency) / FrequencyIncrement.Units,
```
MaggieSalak, line 77: I'm not sure I understand this expression. The dividend here is `Hertz` and divisor is a `double`. We don't explicitly define the / operator in `Hertz` so I'm not quite sure how this works; could you help me understand?

atifaziz: Subtracting two frequencies returns the difference as a `long`:

```c#
public static long operator -(Hertz a, Hertz b) => checked((long)a.value - (long)b.value);
```

My thinking was that the difference between two frequencies isn't a frequency but just a number that represents the units of difference. So that expression is really a `long` being divided by a `double`. The whole idea of the primitives is that you can define the rules governing the correct use of its values without much ceremony.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
MaggieSalak: Makes sense, thanks for clarifying

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470RP2.cs
```diff
@@ -280,46 +302,32 @@ public override RX2ReceiveWindow GetDefaultRX2ReceiveWindow(DeviceJoinInfo devic
...
-        }
-
         [Obsolete("#655 - This Rxpk based implementation will go away as soon as the complete LNS implementation is done.")]
         private static int GetChannelNumber(Rxpk upstreamChannel, double startUpstreamFreq, int startChannelNumber = 0) =>
-            startChannelNumber + (int)Math.Round((upstreamChannel.Freq - startUpstreamFreq) / FrequencyIncrement, 0, MidpointRounding.AwayFromZero);
-        private static int GetChannelNumber(double upstreamChannelFrequency, double startUpstreamFreq, int startChannelNumber = 0) =>
-            startChannelNumber + (int)Math.Round((upstreamChannelFrequency - startUpstreamFreq) / FrequencyIncrement, 0, MidpointRounding.AwayFromZero);
+            startChannelNumber + (int)Math.Round((upstreamChannel.Freq - startUpstreamFreq) / FrequencyIncrement.Value, 0, MidpointRounding.AwayFromZero);
+        private static int GetChannelNumber(Hertz upstreamChannelFrequency, Hertz startUpstreamFreq, int startChannelNumber = 0) =>
+            startChannelNumber + (int)Math.Round((upstreamChannelFrequency.InMega - startUpstreamFreq.InMega) / FrequencyIncrement.Value, 0, MidpointRounding.AwayFromZero);
```
MaggieSalak, line 331: This is correct but I think we could do the same as in RegionCN470RP1
```
private static int GetChannelNumber(Hertz upstreamChannelFrequency, Hertz startUpstreamFreq, int startChannelNumber = 0) =>
startChannelNumber + (int)Math.Round((upstreamChannelFrequency - startUpstreamFreq) / FrequencyIncrement.Units, 0, MidpointRounding.AwayFromZero);
```

Tests/Unit/LoRaTools/RegionCN470RP2TestData.cs
```diff
@@ -63,13 +62,14 @@ public static IEnumerable<object[]> TestRegionFrequencyData()
...
-           {
-               new object[] { region, 470, 8, 0 },
-               new object[] { region, 510, 10, 0 },
-               new object[] { region, 509.8, 100, 0 },
-               new object[] { region, 469.9, 110, 0 },
-           };
+            from x in new (Hertz Frequency, ushort DataRate)[]
+            {
+                (Mega(470.0),   8),
+                (Mega(510.9),  10),
```
MaggieSalak, line 68: Here the test data is changed from what it was before (510).

MaggieSalak (approved):

Tests/Unit/LoRaTools/RegionCN470RP2TestData.cs
atifaziz: Wow, you have eagle eyes! Thanks for spotting!

[Eagle eyes!](https://giphy.com/gifs/bbcearth-animal-dYdGA39uJWMs0EJdy8)

atifaziz: Fixed with 5bcfaaba0f0cd1b5f6b4e5fad22796b0ae01fcd5.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Regions/RegionCN470RP2.cs
atifaziz: Sorry, but I am confused. There is no getting of channel number in `RegionCN470RP1`.

atifaziz: Okay, I think I got it now. You mean the channel number math in `TryGetDownstreamChannelFrequency` of `RegionCN470RP1`.  Done in ced2959e2ac67572cc3dbceae23d1252c870a59b.

PR #1026, opened 2021-12-09
Building both NetworkServer and BasicStation modules

danigian:
We should build both docker images in the pipeline


Mandur (approved):
After discussing I agree with the idea

PR #1031, opened 2021-12-09
Modify MultiConcentrator Test to run the start_basicsstation.sh programmatically

### LINKED ISSUE #1030: Modify MultiConcentrator Test to run the start_basicsstation.sh programmatically

danigian:
# PR for issue #1030 

## How is this addressed

- [x] Refactoring LoRaEngine/modules/LoRaBasicsStationModule/start_basicsstation.sh to allow STATION_PATH and RADIODEV parametrization
- [x] Adding new configuration in Tests/Common/TestConfiguration.cs
- [x] Adding two methods for Starting and Stopping BasicsStation in Tests/Common/TestUtils.cs
- [x] Implementing IAsyncLifetime for Tests/E2E/MultiConcentratorTests.cs and letting it start stop the BasicsStation

## Validation
- Full CI run here https://github.com/Azure/iotedge-lorawan-starterkit/runs/4507078494?check_suite_focus=true (check multiconcentrator tests and the fact that basic station is programmatically started)


Tests/Common/TestUtils.cs
```diff
@@ -182,5 +185,107 @@ public static TimeSpan GetStartTimeOffsetForSecondWindow()
...
+                        UseShellExecute = false,
+                        CreateNoWindow = true,
+                    }
+                };
+                _ = folderCreateProcess.Start();
+                if (!folderCreateProcess.WaitForExit(30_000))
+                {
+                    throw new TimeoutException("Creating and copying over SSH the needed files took more than expected.");
+                }
+                tempDirectory = $"/tmp/{randomDirectoryName}";
```
Mandur, line 268: Nit: could we output this tempdirectory variable so if needed we could fetch the basic station log from the ci

Mandur (approved):

Tests/Common/TestUtils.cs
danigian: Temporary directory gets cleaned up while disposing the test class, therefore we can't fetch the logs afterwards.
We could decide to log always everything, but it could be very verbose.

Let's keep this discussion separate from this PR and keep this as an improvement point for the future

Issue #950, opened 2021-12-01
[Story] Metrics are raised on CUPS endpoint


bastbu:
- [ ] Metrics for the CUPS endpoint are raised / CUPS endpoint complies with existing metrics definitions (e.g. unhandled exceptions etc)

PR #1034, opened 2021-12-09
Adding unhandled exception metric to CupsProtocolMessageProcessor

danigian:
# PR for issue #950

## What is being addressed

Raising metrics for unhandled exception in CUPS Protocol Processor


Tests/Unit/NetworkServer/CupsProtocolMessageProcessorTests.cs
```diff
@@ -41,7 +41,9 @@ public CupsProtocolMessageProcessorTests()
             this.logger = new Mock<ILogger<CupsProtocolMessageProcessor>>();
             this.processor = new CupsProtocolMessageProcessor(this.basicsStationConfigurationService.Object,
                                                               this.deviceAPIServiceBase.Object,
-                                                              this.logger.Object);
+                                                              this.logger.Object,
+                                                              // Do not pass meter since metric testing will be unreliable due to interference from test classes running in parallel.
+                                                              null);
```
bastbu, line 46: [nit] we could also pass a `TestMeter.Instance` if you don't like the nullability here

bastbu (approved):

Tests/Unit/NetworkServer/CupsProtocolMessageProcessorTests.cs
danigian: I like nullability ❤️

Issue #970, opened 2021-12-03
Replace `LoRaMessageType` with `MacMessageType`


atifaziz:

atifaziz:
Closed via merge of PR #1046 in 1951dfe916d6920747b3a06356974f6ca331fbb4.

PR #1046, opened 2021-12-10
🚧 Consolidate `LoRaMessageType` and `MacMessageType`

atifaziz:
# PR for issue #970

## What is being addressed

(coming soon)

## How is this addressed

(coming soon)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
```diff
@@ -140,25 +141,21 @@ public static bool TryCreateLoRaPayload(Rxpk rxpk, out LoRaPayload loRaPayloadMe
             if (rxpk is null) throw new ArgumentNullException(nameof(rxpk));
 
             var convertedInputMessage = Convert.FromBase64String(rxpk.Data);
-            var messageType = convertedInputMessage[0];
+            var messageType = new MacHeader(convertedInputMessage[0]).MessageType;
```
bastbu, line 144: Aren't we more lenient here with respect to the conversions that we accept? Previously, if we had a `convertedInputMessage[0]` equal to let's say `51` it would not match in the `switch` statement. Now, if we have `51` it just gets shifted as `51 >> 5` and we get back a valid `MessageType`. Is that change safe?

bastbu: Edit: I now see that we have the necessary checks in the constructor. Will resolve the comment.

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -207,8 +207,8 @@ public LoRaPayloadData(LoRaMessageType mhdr, byte[] devAddr, FrameControlFlags f
             var macPyldSize = devAddr.Length + FrameControl.Size + fcnt.Length + fOptsLen + frmPayloadLen + fPortLen;
             RawMessage = new byte[1 + macPyldSize + 4];
             Mhdr = new Memory<byte>(RawMessage, 0, 1);
-            RawMessage[0] = (byte)mhdr;
-            LoRaMessageType = mhdr;
+            RawMessage[0] = (byte)(new MacHeader(messageType));
```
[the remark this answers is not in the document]
atifaziz: Test added with 9395c6f87abaf83b39f520c7875e509b26d4e11c.

PR #1049, opened 2021-12-12
Dynamically get Ip Address of the CI device

### LINKED ISSUE #1007: Azure P2S devices can switch IP
We use two P2S connections on Azure to connect our runners and the EFLOW VM (always on IP .0):
- One P2S connection from the test runner (itestup) for the EFLOW VM to send back its log on premise (edge -> cloud)
- One P2S on ITestArm2 for it to send messages to the EFLOW VM (edge -> cloud)

It seems that when the CI is turned off for a long time (>3 days) the IP leases are reset, and it could be that the two swap IP. As there is no cross-premise DNS resolution we must use IP address in the CI, leading to problems to EFLOW sending back the logs to the edge as itestup could be either on Ip 1 or 2.

Potential ideas:
- Make the concentrator on itestup send messages to the EFLOW VM to only have one P2S connection
- Programmatically get the network status at iotedge deployment time to deploy the correct values
- Have a Site-to-site connection

Mandur:
# PR for issue #1007

## What is being addressed

Itestup could change IP address

## How is this addressed

The IP address is now checked at runtime. Secret LOG_TO_TCP_ADDRESS will be cleaned after PR merge. 
Successfull deplyoment on the branch: https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1570237441


.github/workflows/e2e-ci.yaml
```diff
@@ -214,6 +213,14 @@ jobs:
         scp -i $SSH_PRIVATE_KEY_PATH ./Tools/BasicStation-Certificates-Generation/ca/root-ca.crt $RPI_USERNAME@$RPI_HOST:$CERT_REMOTE_PATH/tc.trust
         scp -i $SSH_PRIVATE_KEY_PATH ./Tools/BasicStation-Certificates-Generation/server/$RPI_HOST.pfx $RPI_USERNAME@$RPI_HOST:$CERT_REMOTE_PATH/server.pfx
 
+    - id: get_itestup_ip_address
```
danigian, line 216: nit: what about calling it "selfHostedAgent" instead of "itestup"?

danigian (approved):

Issue #942, opened 2021-11-30
[Story] Raise leftover metrics


bastbu:
## Description

In order to not generate more conflicts for #904 we moved out raising metrics that have to do with caching from #747. We should raise these metrics as part of this story.

## Acceptance criteria

- [ ] DeviceLoadRequests metric is raised
- [ ] DeviceCacheHit metric is raised

PR #1052, opened 2021-12-13
Raise DeviceLoadRequests and DeviceCacheHit metrics

MaggieSalak:
# PR for issue #942 

## What is being addressed

Metrics related to caching and device loading are raised.

## How is this addressed

- New metrics defined in `MetricRegistry`
- `DeviceCacheHits` metric is raised whenever a device is found in device cache
- `DeviceLoadRequests` metric is raised whenever a device load is attempted, including devices that are later successfully retrieved from cache


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
```diff
@@ -128,6 +132,8 @@ public ILoRaDeviceRequestQueue GetLoRaRequestQueue(LoRaRequest request)
         {
             if (request is null) throw new ArgumentNullException(nameof(request));
 
+            this.deviceLoadRequests?.Add(1);
```
bastbu, line 135: It seems a bit unintuitive to have a `DeviceLoadRequest` inside the `GetLoRaRequestQueue`. If I see it correctly the actual loading is done in `DeviceLoaderSynchronizer.LoadAsync`, maybe we should track it farther down in the stack, at a different location in the `LoRaDeviceRegistry`?

If the reason for the metrics being raised in different call sites is that we didn't encapsulate the caching methods (and we have to manually query the cache before issuing a search, therefore have to raise the metric on the call site), then we could also reconsider the semantics of `DeviceLoadRequest` and track the number of requests/queries that did hit the cache. The total number of load requests will always be the sum of the number of cache hits and cache misses.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -62,6 +64,7 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
                     this.logger.LogInformation("join request received");
 
                     loRaDevice = await this.deviceRegistry.GetDeviceForJoinRequestAsync(devEUI, devNonce);
+                    this.deviceLoadRequests?.Add(1);
```
bastbu, line 67: Why not move this one inside the `LoRaDeviceRegistry` as well, then we have all `DeviceLoadRequests` metrics in one place?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
```diff
@@ -62,6 +63,8 @@ public sealed class LoRaDeviceRegistry : ILoRaDeviceRegistry
             this.getOrCreateLoadingDevicesRequestQueueLock = new object();
             this.getOrCreateJoinDeviceLoaderLock = new object();
             this.meter = meter;
+            this.deviceCacheHits = meter?.CreateCounter<int>(MetricRegistry.DeviceCacheHits);
```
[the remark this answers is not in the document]
MaggieSalak: Good point, done in 0a78c9a76364fb30d9571d809010dc62f056e187

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
MaggieSalak: Done in 1ccb5eff69ce23adc0649c27b60ade46eface220

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
MaggieSalak: As per discussion on Teams, I pushed 34d1ea22b33a39afd68e632b41b44b00d1c1a3ca where the `DeviceLoadRequests` metric is used to count requests made against the Lora device API service and not the cache hits. The `deviceLoadRequests are now kept track of in the `LoRaDeviceAPIService.cs`. Let me know if you have further comments about this.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
```diff
@@ -240,7 +244,10 @@ public bool TryGetByDevEui(string devEUI, [MaybeNullWhen(returnValue: false)] ou
         {
             lock (this.syncLock)
             {
-                _ = this.euiCache.TryGetValue(devEUI, out device);
+                if (this.euiCache.TryGetValue(devEUI, out device))
+                {
+                    this.deviceCacheHits?.Add(1);
```
bastbu, line 249: [nit] we could also track it together in `TrackCacheStats`, where we essentially also keep track of the statistics for the internal implementation

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
```diff
@@ -61,7 +57,6 @@ public sealed class LoRaDeviceRegistry : ILoRaDeviceRegistry
             DevAddrReloadInterval = TimeSpan.FromSeconds(30);
             this.getOrCreateLoadingDevicesRequestQueueLock = new object();
             this.getOrCreateJoinDeviceLoaderLock = new object();
-            this.meter = meter;
```
bastbu, line 64: Nice, thanks for the cleanup!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
```diff
@@ -20,14 +21,17 @@ public sealed class LoRaDeviceAPIService : LoRaDeviceAPIServiceBase
...
         public LoRaDeviceAPIService(NetworkServerConfiguration configuration,
                                     IServiceFacadeHttpClientProvider serviceFacadeHttpClientProvider,
-                                    ILogger<LoRaDeviceAPIService> logger)
+                                    ILogger<LoRaDeviceAPIService> logger,
+                                    Meter meter)
             : base(configuration)
         {
             this.serviceFacadeHttpClientProvider = serviceFacadeHttpClientProvider;
             this.logger = logger;
+            this.deviceLoadRequests = meter?.CreateCounter<int>(MetricRegistry.DeviceLoadRequests);
```
bastbu, line 34: [nit] same question here, maybe we can not use safe navigation, this would give us slightly more guarantees when executing the tests

bastbu (approved):

Issue #837, opened 2021-11-17
Manual testing of China frequency


roel4ez:
we can prepare the concentrator and prepare a checklist for executing the test in a confined environment

- [x] #839

Mandur:
@danigian I guess the low TX power is blocked, right?

PR #1054, opened 2021-12-13
Feature/fix cn

Mandur:
# PR for issue #837

## What is being addressed

Fix some things in the China frequencies after the manual tests. 
Start removing the Dynamic region assignation (TryResolveRegion) and refactor tests.


Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
```diff
@@ -0,0 +1,113 @@
+
+#include <LoRaWan.h>
+// THOSE ARDUINO SAMPLE REQUIRES CHANGES IN THE ARDUINO CODE TO RUN DESCRIBED IN THE MAIN REPO
+// THE CONCENTRATOR IS EXPECTED TO HAVE THE FOLLOWING FREQUENCIES ACTIVATED
+// 498.3, 498.7, 498.9, 499.1, 499.3, 499.5, 499.7, 499.9
+// please refer to the repo documentation for further information
+
+//set to true to send confirmed data up messages
+bool confirmed = true;
+//application information, should be similar to what was provisiionned in the device twins
```
p-schuler, line 10:
```suggestion
//application information, should be similar to what was provisioned in the device twins
```

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
```diff
@@ -0,0 +1,113 @@
...
+char data[10];
+char buffer[256];
+int i = 0;
+int lastCall = 0;
+
+
+void setup(void)
+{
+  SerialUSB.begin(115200);
+  while (!SerialUSB);
```
p-schuler, line 42: That's a very tight loop, I assume we do this everywhere? Any way we could sleep for a bit before checking again?

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
```diff
@@ -0,0 +1,113 @@
...
+    else
+      result = lora.transferPacket(data, 10);
+    i++;
+
+    if (result)
+    {
+      short length;
+      short rssi;
+
+      memset(buffer, 0, 256);
```
p-schuler, line 94: should we make the 256 a variable we can use? It's re-used in a few places. Same for the length of the data array?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
```diff
@@ -85,8 +85,8 @@ private string Message(DownlinkPktFwdMessage message, Region region)
                 writer.WriteNumber("RxDelay", message.LnsRxDelay);
                 writer.WriteNumber("RX1DR", dataRate);
                 writer.WriteNumber("RX1Freq", (ulong)(message.Txpk.Freq * 1e6));
-                writer.WriteNumber("RX2DR", region.GetDefaultRX2ReceiveWindow().DataRate);
-                writer.WriteNumber("RX2Freq", (ulong)region.GetDefaultRX2ReceiveWindow().Frequency);
+                writer.WriteNumber("RX2DR", region.GetDefaultRX2ReceiveWindow(message.DeviceJoinInfo).DataRate);
```
p-schuler, line 88: We are getting the default RX2 receive window twice. I see that some of the regions have quite some logic to get to the window. Would it make sense to cache that information and only get it once?

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
Mandur: What is your worry there? this code would trigger onlywhen we actively open the serial

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
```diff
@@ -81,22 +80,14 @@ public void DispatchRequest(LoRaRequest request)
...
-                {
-                    // log is generated in Region factory
-                    // move here once V2 goes GA
-                    request.NotifyFailed(LoRaDeviceRequestFailedReason.InvalidRegion);
-                    return;
-                }
-
-                this.loraRegion = currentRegion;
+                // log is generated in Region factory
+                // move here once V2 goes GA
```
p-schuler, line 86: how do we track this?

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
atifaziz: Why not just use `sizeof(buffer)` here?

Tests/Common/MessageProcessorTestBase.cs
```diff
@@ -71,10 +74,12 @@ public MessageProcessorTestBase()
             LoRaDeviceClient = new Mock<ILoRaDeviceClient>();
             this.cache = new MemoryCache(new MemoryCacheOptions() { ExpirationScanFrequency = TimeSpan.FromSeconds(5) });
             ConnectionManager = new LoRaDeviceClientConnectionManager(this.cache, NullLogger<LoRaDeviceClientConnectionManager>.Instance);
-
             var requestHandler = new DefaultLoRaDataRequestHandler(ServerConfiguration, FrameCounterUpdateStrategyProvider, new LoRaPayloadDecoder(NullLogger<LoRaPayloadDecoder>.Instance), deduplicationFactory, adrStrategyProvider, adrManagerFactory, functionBundlerProvider, NullLogger<DefaultLoRaDataRequestHandler>.Instance, meter: null);
-
             LoRaDeviceFactory = new TestLoRaDeviceFactory(ServerConfiguration, LoRaDeviceClient.Object, ConnectionManager, DeviceCache, requestHandler);
+
+            // By default we pick EU868 region.
+            Region = Enum.TryParse<LoRaRegionType>(Environment.GetEnvironmentVariable("REGION"), out var loraRegionType) ?
```
atifaziz, line 81: Consider making the parsing case insensitive:

```suggestion
            Region = Enum.TryParse<LoRaRegionType>(Environment.GetEnvironmentVariable("REGION"), ignoreCase: true, out var loraRegionType) ?
```

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
p-schuler: much better! didn't know what was available in the arduino env.

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
p-schuler: "we actively open the serial"? we do this here right? We wait for it to become available, no? If it is not, this is a very tight loop. I think we have this everywhere and it seems to work, so I assume it's ok?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
Mandur: No idea, I did not write this, being honest I don't even know what this log mean. Do you?

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
Mandur: Yeah, this is a copy from their sample, to be honest I don't know how much is the backoff/retry on this.

Mandur: If you see a problem with it I can make some research but this thing has always worked so far and it's in the sample repo

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
Mandur: Changed in every arduino sample. Resolve if that is ok for you

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
```diff
@@ -0,0 +1,113 @@
+
+#include <LoRaWan.h>
+// THOSE ARDUINO SAMPLE REQUIRES CHANGES IN THE ARDUINO CODE TO RUN DESCRIBED IN THE MAIN REPO
+// THE CONCENTRATOR IS EXPECTED TO HAVE THE FOLLOWING FREQUENCIES ACTIVATED
+// 498.3, 498.7, 498.9, 499.1, 499.3, 499.5, 499.7, 499.9
```
Mandur, line 5: This is RP1

Mandur: wrong

Arduino/CN470RP2/TransmissionTestABPLoRa/TransmissionTestABPLoRa.ino
p-schuler: I'm still seeing this:

memset(buffer, 0, 256);

did you push?

p-schuler: looks like the browser was on a stale commit. Looks good.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
p-schuler: This is a very old comment. Let's remove it I don't think it's accurate anymore.

p-schuler (approved):
LGTM

PR #1058, opened 2021-12-13
Fixing log variables

danigian:


Mandur:
Sorry for introducing this bug & thanks for fixing it!

.github/workflows/e2e-ci.yaml
```diff
@@ -256,6 +256,7 @@ jobs:
           APPINSIGHTS_INSTRUMENTATIONKEY: ${{ secrets.APPINSIGHTS_INSTRUMENTATIONKEY }}
           CERT_REMOTE_PATH: ${{ needs.env_var.outputs.CertRemotePath }}
           SERVER_PFX_PASSWORD: ${{ secrets.SERVER_PFX_PASSWORD }}
+          NET_SRV_LOG_TO_TCP_ADDRESS: "itestup"
```
Mandur, line 259: Is this really needed?

danigian: If we don't set this here, the template won't be able to replace the value, therefore its execution on AIO will fail

PR #1059, opened 2021-12-13
Use `DevNonce` primitive

bastbu:
# PR for issue #963

## What is being addressed

With this PR we reuse `DevNonce` when possible.

## How is this addressed

As part of this PR, broken test setup, originating from how dev nonces were generated in the `SimulatedDevice` were fixed as well.


LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -91,7 +94,7 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
                 var cacheKeyDevNonce = string.Concat(devEUI, ":", devNonce);
                 var lockKeyDevNonce = string.Concat(cacheKeyDevNonce, ":joinlockdevnonce");
 
-                if (this.cacheStore.StringSet(cacheKeyDevNonce, devNonce, TimeSpan.FromMinutes(5), onlyIfNotExists: true))
+                if (this.cacheStore.StringSet(cacheKeyDevNonce, devNonce?.ToString(), TimeSpan.FromMinutes(5), onlyIfNotExists: true))
```
bastbu, line 97: I decided to not add value-type support on the cache store, since the `devNonce` is present as `string` key already due to `string.Concat(devEUI, ":", devNonce)`

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -91,7 +90,7 @@ async Task ProcessJoinRequestAsync(LoRaRequest request)
                     }
 
                     // Make sure that is a new request and not a replay
-                    if (!string.IsNullOrEmpty(loRaDevice.DevNonce) && loRaDevice.DevNonce == devNonce)
+                    if (loRaDevice.DevNonce is not null && loRaDevice.DevNonce == joinReq.DevNonce)
```
atifaziz, line 93: nit:

```suggestion
                    if (loRaDevice.DevNonce is { } devNonce && devNonce == joinReq.DevNonce)
```

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -59,6 +60,8 @@ public DeviceGetter(RegistryManager registryManager, ILoRaDeviceCacheStore cache
 
             try
             {
+                var devNonceBuffer = new byte[DevNonce.Size];
+                DevNonce? devNonce = Hexadecimal.TryParse(rawDevNonce, devNonceBuffer) ? DevNonce.Read(devNonceBuffer) : null;
```
atifaziz, line 64: `Hexadecimal.TryParse` has an overload for `ulong`. Why not add one for `ushort` to make this easier and not need an array of two bytes for `devNonceBuffer`? Another option would be to use a local function to get around the limitation of not being able to use `Span<>` in `async` methods.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -322,7 +322,12 @@ public virtual async Task<bool> InitializeAsync(NetworkServerConfiguration confi
                     NetID = twin.Properties.Reported[TwinProperty.NetID].Value as string;
 
                 if (twin.Properties.Reported.Contains(TwinProperty.DevNonce))
-                    DevNonce = twin.Properties.Reported[TwinProperty.DevNonce].Value as string;
+                {
+                    var devNonceString = twin.Properties.Reported[TwinProperty.DevNonce].Value as string;
+                    var buffer = new byte[LoRaWan.DevNonce.Size];
+                    _ = Hexadecimal.TryParse(devNonceString, buffer);
```
atifaziz, line 328: Same comment as earlier. I think an overload of `Hexadecimal.TryParse` for `ushort` will help here too.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
```diff
@@ -106,7 +107,7 @@ private async Task<SearchDevicesResult> SearchDevicesAsync(string gatewayID = nu
                 ["DevAddr"] = devAddr,
                 ["DevEUI"] = devEUI,
                 ["AppEUI"] = appEUI,
-                ["DevNonce"] = devNonce
+                ["DevNonce"] = devNonce?.ToString("N", CultureInfo.InvariantCulture)
```
atifaziz, line 110: Does this need to be in hex? The nonce is just a number so it seems more natural just to express it in base 10 decimal.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
```diff
@@ -124,7 +125,10 @@ public static byte[] CalculateKey(LoRaPayloadKeyType keyType, byte[] appnonce, b
             aes.Padding = PaddingMode.None;
             aes.IV = new byte[16];
 
-            var pt = type.Concat(appnonce).Concat(netid).Concat(devnonce).Concat(new byte[7]).ToArray();
+            var devNonceBytes = new byte[DevNonce.Size];
+            _ = devnonce.Write(devNonceBytes);
+
+            var pt = type.Concat(appnonce).Concat(netid).Concat(devNonceBytes).Concat(new byte[7]).ToArray();
```
atifaziz, line 131: What about simplifying this to the following that doesn't require as many allocations and conversion and would allow direct use of `DevNonce.Write` without going through a two-byte array?

```c#
var pt = new byte[appnonce.Length + netid.Length + DevNonce.Size + 7];
Array.Copy(appnonce, pt, appnonce.Length);
Array.Copy(netid, 0, pt, appnonce.Length, netid.Length);
_ = devnonce.Write(pt.AsSpan(appnonce.Length + netid.Length));
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
```diff
@@ -109,7 +110,7 @@ public byte[] CalculateMic(string appKey, byte[] algoinput)
         /// <summary>
         /// Calculate the Netwok and Application Server Key used to encrypt data and compute MIC.
         /// </summary>
-        public static byte[] CalculateKey(LoRaPayloadKeyType keyType, byte[] appnonce, byte[] netid, byte[] devnonce, byte[] appKey)
+        public static byte[] CalculateKey(LoRaPayloadKeyType keyType, byte[] appnonce, byte[] netid, DevNonce devnonce, byte[] appKey)
```
atifaziz, line 113: Consider renaming `devnonce` to `devNonce`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinRequest.cs
```diff
@@ -57,10 +53,10 @@ public LoRaPayloadJoinRequest(byte[] inputMessage)
             // get the DevEUI
             DevEUI = new Memory<byte>(inputMessage, 9, 8);
             // get the DevNonce
-            DevNonce = new Memory<byte>(inputMessage, 17, 2);
+            DevNonce = DevNonce.Read(inputMessage[17..19].AsSpan());
```
atifaziz, line 56: Slicing the array directly with `[17..19]` creates a _copy_ of the array slice. What you want here is a span over the source array:

```suggestion
            DevNonce = DevNonce.Read(inputMessage.AsSpan(17));
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinRequest.cs
```diff
@@ -121,7 +117,9 @@ public override byte[] GetByteMessage()
             messageArray.AddRange(Mhdr.ToArray());
             messageArray.AddRange(AppEUI.ToArray());
             messageArray.AddRange(DevEUI.ToArray());
-            messageArray.AddRange(DevNonce.ToArray());
+            var buffer = new byte[DevNonce.Size];
```
atifaziz, line 120: If we can determine the output buffer length in advance, then why not copy to the array instead of a dynamic list, array conversions and then here an intermediate two-byte array allocation for the nonce?

Tests/Common/DevNonceExtensions.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Common
+{
+    public static class DevNonceExtensions
+    {
+        public static byte[] GetBytes(this DevNonce devNonce)
```
atifaziz, line 8: Great compromise! 👍🏼

Tests/Integration/DevAddrCacheTest.cs
```diff
@@ -237,12 +237,13 @@ public async Task Multi_Gateway_When_DevAddr_Is_Not_In_Cache_Query_Iot_Hub_Only_
...
             var deviceGetter = new DeviceGetter(registryManagerMock.Object, this.cache);
             // Simulate three queries
+            var devNonce = new DevNonce(0xABCD);
             var tasks = new Task[4]
               {
-                deviceGetter.GetDeviceList(null, gateway1, "ABCD", devAddrJoining),
-                deviceGetter.GetDeviceList(null, gateway1, "ABCD", devAddrJoining),
-                deviceGetter.GetDeviceList(null, gateway2, "ABCD", devAddrJoining),
-                deviceGetter.GetDeviceList(null, gateway2, "ABCD", devAddrJoining)
+                deviceGetter.GetDeviceList(null, gateway1, devNonce, devAddrJoining),
```
atifaziz, line 243: nit: Perhaps we can use this opportunity to simplify the whole query instead of extracting the repeated device nonce into a variable:

```c#
var tasks =
    from gw in new[] { gateway1, gateway2 }
    select Enumerable.Repeat(gw, 2) into gws // repeat each gateway twice
    from gw in gws
    select deviceGetter.GetDeviceList(null, gw, new DevNonce(0xABCD), devAddrJoining);
```

What do you think?

Tests/Common/TestUtils.cs
```diff
@@ -133,7 +133,7 @@ public static Twin CreateTwin(Dictionary<string, object> desired = null, Diction
                 { TwinProperty.DevAddr, simulatedDevice.DevAddr },
                 { TwinProperty.AppSKey, simulatedDevice.AppSKey },
                 { TwinProperty.NwkSKey, simulatedDevice.NwkSKey },
-                { TwinProperty.DevNonce, simulatedDevice.DevNonce },
+                { TwinProperty.DevNonce, simulatedDevice.DevNonce.ToString("N", null) },
```
atifaziz, line 136: I would question why this needs to be in hexadecimal.

Tests/Integration/DevAddrCacheTest.cs
```diff
@@ -335,11 +336,12 @@ public async Task Multi_Gateway_When_DevAddr_Is_In_Cache_Without_Key_Should_Not_
             await LockDevAddrHelper.PrepareLocksForTests(this.cache, lockToTake);
 
             var deviceGetter = new DeviceGetter(registryManagerMock.Object, this.cache);
+            var devNonce = new DevNonce(0xABCD);
             var tasks = new Task[3]
             {
-                deviceGetter.GetDeviceList(null, gatewayId, "ABCD", devAddrJoining),
-                deviceGetter.GetDeviceList(null, gatewayId, "ABCD", devAddrJoining),
-                deviceGetter.GetDeviceList(null, gatewayId, "ABCD", devAddrJoining),
+                deviceGetter.GetDeviceList(null, gatewayId, devNonce, devAddrJoining),
```
atifaziz, line 342: Like the previous case, we can simplify this via a query so the dev nonce is repeated once:

```c#
var tasks =
    from gw in Enumerable.Repeat(gatewayId, 3)
    select deviceGetter.GetDeviceList(null, gw, new DevNonce(0xABCD), devAddrJoining);
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
```diff
@@ -36,7 +37,10 @@ public static string CalculateKey(byte[] type, byte[] appnonce, byte[] netid, by
 #pragma warning restore CA5358 // Review cipher mode usage with cryptography experts
             aes.Padding = PaddingMode.None;
 
-            var pt = type.Concat(appnonce).Concat(netid).Concat(devnonce).Concat(new byte[7]).ToArray();
+            var devNonceBytes = new byte[DevNonce.Size];
+            _ = devnonce.Write(devNonceBytes);
+
+            var pt = type.Concat(appnonce).Concat(netid).Concat(devNonceBytes).Concat(new byte[7]).ToArray();
```
atifaziz, line 43: Again, we can do like suggested for `LoRaPayload.CalculateKey`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevNonce.cs
```diff
@@ -28,7 +27,29 @@ namespace LoRaWan
...
-        public override string ToString() => this.value.ToString(CultureInfo.InvariantCulture);
+        public override string ToString() => this.value.ToString(null, null);
+
+        public string ToString(string? format, IFormatProvider? formatProvider)
+        {
+#pragma warning disable format
+            return format switch
+            {
+                "N"                                 => ToHex(LetterCase.Upper),
+                "n"                                 => ToHex(LetterCase.Lower),
```
atifaziz, line 38: `X` and `x` would be more standard here, but perhaps we don't need it at all? I don't think we should be treating dev nonce values in hexadecimal notation assuming they are just monotonically increasing numbers?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
```diff
@@ -201,11 +201,11 @@ private JoinDeviceLoader GetOrCreateJoinDeviceLoader(IoTHubDeviceInfo ioTHubDevi
         /// <summary>
         /// Searchs for devices that match the join request.
         /// </summary>
-        public async Task<LoRaDevice> GetDeviceForJoinRequestAsync(string devEUI, string devNonce)
+        public async Task<LoRaDevice> GetDeviceForJoinRequestAsync(string devEUI, DevNonce devNonce)
```
atifaziz, line 204: Since there was a check for null or empty device nonce before, wouldn't the right translation here be a nullable `DevNonce`?

atifaziz (commented):
See my comments/suggestions; a few tweaks and then this should be good to go. Thanks!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
bastbu: Addressed with 683e769799ae8a952bd44c48407d86ee7ec751a9

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevNonce.cs
bastbu: I will remove the hex representation entirely

Tests/Integration/DevAddrCacheTest.cs
bastbu: Done in e5b0eb53bd64a93a01c97f8a125a504d89ea93c8

Tests/Integration/DevAddrCacheTest.cs
bastbu: 794f3b35b48935865ef9badb14c0adff9d05e6b2

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
bastbu: Reverted in 9f3eb36e738fbaa1a3f376ab154540aa659eb33c

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevNonce.cs
bastbu: Done in 9f3eb36e738fbaa1a3f376ab154540aa659eb33c

Tests/Common/TestUtils.cs
bastbu: Changed in 9f3eb36e738fbaa1a3f376ab154540aa659eb33c

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinRequest.cs
bastbu: Done in fd28236af32ae579bb5c855c957271aba387e04b

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
bastbu: Done in c6e00385d1a8b7384e1e813a8c8d64b343160f05

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
bastbu: Done in c6e00385d1a8b7384e1e813a8c8d64b343160f05

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
bastbu: Valid point, but based on the usage it seems that we didn't need the nullability in the first place. Do you think there's a bug somewhere?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -322,7 +322,10 @@ public virtual async Task<bool> InitializeAsync(NetworkServerConfiguration confi
                     NetID = twin.Properties.Reported[TwinProperty.NetID].Value as string;
 
                 if (twin.Properties.Reported.Contains(TwinProperty.DevNonce))
-                    DevNonce = twin.Properties.Reported[TwinProperty.DevNonce].Value as string;
+                {
+                    var devNonceString = twin.Properties.Reported[TwinProperty.DevNonce].Value as string;
+                    DevNonce = ushort.TryParse(devNonceString, out var d) ? new DevNonce(d) : null;
```
atifaziz, line 327: I am surprised that analyzers did not pick up (here and elsewhere in this PR) on using the overload that requires `NumberStyles` and `IFormatProvider`, like so:

```c#
DevNonce = ushort.TryParse(devNonceString, NumberStyles.None, CultureInfo.InvariantCulture, out var d) ? new DevNonce(d) : null;
```

Tests/Unit/LoRaWan/DevNonceTests.cs
```diff
@@ -4,6 +4,7 @@
 namespace LoRaWan.Tests.Unit
 {
     using System;
+    using System.Globalization;
```
atifaziz, line 7: We can remove this import:

```suggestion
```

atifaziz (approved):
Apart from [one comment regarding parsing](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1059#discussion_r768850030), LGTM.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -91,7 +93,7 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
                 var cacheKeyDevNonce = string.Concat(devEUI, ":", devNonce);
                 var lockKeyDevNonce = string.Concat(cacheKeyDevNonce, ":joinlockdevnonce");
```
atifaziz, line 94: FYI, noticed that this is unused and opened PR #1070 to remove it as it's unrelated to this.

bastbu: Thanks, will take a look!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
bastbu: Addressed this in 1cf5d208c6122f300c2223dc70ecb893aba9928d

PR #1073, opened 2021-12-15
Use `DevAddr` in `LoraKeysManagerFacade`

bastbu:
# PR for issue #962

## What is being addressed

With this PR we use `DevAddr` in `LoRaKeysManagerFacade`.

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
@@ -48,7 +48,7 @@ namespace LoraKeysManagerFacade
             }
 
             // ABP parameters
-            var devAddr = req.Query["DevAddr"];
+            string devAddrString = req.Query["DevAddr"];
             // OTAA parameters
             var devEUI = req.Query["DevEUI"];
             string rawDevNonce = req.Query["DevNonce"];
@@ -62,6 +62,7 @@ namespace LoraKeysManagerFacade
             try
             {
                 DevNonce? devNonce = ushort.TryParse(rawDevNonce, NumberStyles.None, CultureInfo.InvariantCulture, out var d) ? new DevNonce(d) : null;
+                DevAddr? devAddr = Hexadecimal.TryParse(devAddrString, out uint rawDevAddr) ? new DevAddr(rawDevAddr) : null;
                 var results = await GetDeviceList(devEUI, gatewayId, devNonce, devAddr, log);
                 var json = JsonConvert.SerializeObject(results);
                 return new OkObjectResult(json);
@@ -81,7 +82,7 @@ namespace LoraKeysManagerFacade
             }
         }
 
-        public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string gatewayId, DevNonce? devNonce, string devAddr, ILogger log = null)
+        public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string gatewayId, DevNonce? devNonce, DevAddr? devAddr, ILogger log = null)
         {
             var results = new List<IoTHubDeviceInfo>();
 
@@ -119,18 +120,17 @@ namespace LoraKeysManagerFacade
                     throw new DeviceNonceUsedException();
                 }
             }
-            else if (devAddr != null)
+            else if (devAddr is { } da)
             {
                 // ABP or normal message
 
                 // TODO check for sql injection
-                devAddr = devAddr.Replace('\'', ' ');
                 var devAddrCache = new LoRaDevAddrCache(this.cacheStore, this.registryManager, log, gatewayId);
-                if (await devAddrCache.TryTakeDevAddrUpdateLock(devAddr))
+                if (await devAddrCache.TryTakeDevAddrUpdateLock(da))
                 {
                     try
                     {
-                        if (devAddrCache.TryGetInfo(devAddr, out var devAddressesInfo))
+                        if (devAddrCache.TryGetInfo(da, out var devAddressesInfo))
                         {
                             for (var i = 0; i < devAddressesInfo.Count; i++)
                             {
@@ -159,7 +159,7 @@ namespace LoraKeysManagerFacade
                         // if the device is not found is the cache we query, if there was something, it is probably not our device.
                         if (results.Count == 0 && devAddressesInfo == null)
                         {
-                            var query = this.registryManager.CreateQuery($"SELECT * FROM devices WHERE properties.desired.DevAddr = '{devAddr}' OR properties.reported.DevAddr ='{devAddr}'", 100);
+                            var query = this.registryManager.CreateQuery($"SELECT * FROM devices WHERE properties.desired.DevAddr = '{da}' OR properties.reported.DevAddr ='{da}'", 100);
                             var resultCount = 0;
                             while (query.HasMoreResults)
                             {
@@ -172,7 +172,7 @@ namespace LoraKeysManagerFacade
                                         var device = await this.registryManager.GetDeviceAsync(twin.DeviceId);
                                         var iotHubDeviceInfo = new DevAddrCacheInfo
                                         {
-                                            DevAddr = devAddr,
+                                            DevAddr = da,
                                             DevEUI = twin.DeviceId,
                                             PrimaryKey = device.Authentication.SymmetricKey.PrimaryKey,
                                             GatewayId = twin.GetGatewayID(),
@@ -192,7 +192,7 @@ namespace LoraKeysManagerFacade
                             {
                                 _ = devAddrCache.StoreInfo(new DevAddrCacheInfo()
                                 {
-                                    DevAddr = devAddr,
+                                    DevAddr = da,
                                     DevEUI = string.Empty
                                 });
                             }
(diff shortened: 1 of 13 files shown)


LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -62,6 +62,7 @@ public DeviceGetter(RegistryManager registryManager, ILoRaDeviceCacheStore cache
             try
             {
                 DevNonce? devNonce = ushort.TryParse(rawDevNonce, NumberStyles.None, CultureInfo.InvariantCulture, out var d) ? new DevNonce(d) : null;
+                DevAddr? devAddr = Hexadecimal.TryParse(devAddrString, out uint rawDevAddr) ? new DevAddr(rawDevAddr) : null;
```
p-schuler, line 65: shouldn't we encapsulate DevAddr parsing logic into the DevAddr?

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -119,18 +120,17 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
                     throw new DeviceNonceUsedException();
                 }
             }
-            else if (devAddr != null)
+            else if (devAddr is { } da)
```
p-schuler, line 123: This makes it worse for me to read. Now I need to understand that da is the devAddr, that was passed in. If we keep the simple != null and use devAddr, things are easier to read.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
bastbu: Ah yes, apologies, had the change in a different commit and did not test/push it yet. Will do that

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
bastbu: How do I tell the compiler in that case that `devAddr` is indeed not null? It has type `DevAddr?`

atifaziz: I agree with Patrick. `da` is not readable and somewhat arbitrary. I would recommend the convention of adding the `some` prefix (borrowing from optional types, which is what nullable value types are) so for `devAddr`, that would become `someDevAddr`. Another (less preferrable) option would be to add the `Value` suffix, as in `devAddrValue`. The `some` prefix is preferred because for `value` (like in the case of property setters where you don't have a choice to name the parameter), `valueValue` looks odd compared to `someValue`.

bastbu: I addressed the bad variable name with b14b97814bc6c62552dea274ac3414defca92fcc, but this doesn't solve the original comment. Should I still change something?

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
bastbu: Addressed and tested with 0ccc5f6d6bc95b7f66651e17c233478a726324ef

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
```diff
@@ -27,6 +27,15 @@ namespace LoRaWan
 
         public DevAddr(uint value) => this.value = value;
 
+        public DevAddr(int networkId, int networkAddress)
+            : this((networkId, networkAddress) switch
+            {
+                ( > 0x7F, _) => throw new ArgumentException(null, nameof(networkId)),
+                (_, > (int)NetworkAddressMask) => throw new ArgumentException(null, nameof(networkAddress)),
```
atifaziz, line 34: Something looks wrong here. The value is being directly compared for equality with the mask. Did you not mean a range check?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
```diff
@@ -27,6 +27,15 @@ namespace LoRaWan
...
+                _ => ((uint)networkId << 25) | (uint)networkAddress
```
atifaziz, line 35: I think we can safely use unchecked conversions at this point:

```suggestion
                var (id, addr) => unchecked(((uint)id << 25) | (uint)addr)
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
```diff
@@ -27,6 +27,15 @@ namespace LoRaWan
 
         public DevAddr(uint value) => this.value = value;
 
+        public DevAddr(int networkId, int networkAddress)
+            : this((networkId, networkAddress) switch
+            {
+                ( > 0x7F, _) => throw new ArgumentException(null, nameof(networkId)),
```
atifaziz, line 33: What about negative values? Think the following would be better:

```suggestion
                ( < 0 or >= 0x80, _) => throw new ArgumentException(null, nameof(networkId)),
```

Tests/Unit/LoRaWan/DevAddrTests.cs
```diff
@@ -3,13 +3,40 @@
...
+            // assert
+            Assert.Equal(maxValue, result);
+        }
+
+        [Fact]
+        public void Init_Validates_Limits()
+        {
+            var maxValue = new DevAddr(uint.MaxValue);
+
+            Assert.Throws<ArgumentException>(() => new DevAddr(maxValue.NetworkId + 1, maxValue.NetworkAddress));
```
atifaziz, line 36: I don't think the bounds being checked here are correct for each part of the `DevAddr`. Also, we don't really assert which of the two arguments failed.

Tests/Unit/LoRaWan/DevAddrTests.cs
```diff
@@ -3,13 +3,40 @@
...
+        public void Init(uint value)
+        {
+            // arrange
+            var maxValue = new DevAddr(value);
+
+            // act
+            var result = new DevAddr(maxValue.NetworkId, maxValue.NetworkAddress);
+
+            // assert
+            Assert.Equal(maxValue, result);
```
atifaziz, line 28: This seems to be a test for equality rather than initialization.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
atifaziz: My 2 cents…it solves the original problem. It's harder to see the correlation between `devAddr` and `da` than between `devAddr` and `someDevAddr`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
bastbu: You're right, I should also take into account negative values (was still thinking in terms of `uint` here, which is wrong). The larger than check on the `NewtorkAddressMask` should be correct, though. Do you think it's wrong?

Tests/Unit/LoRaWan/DevAddrTests.cs
bastbu: Addressed this with 01e2056508a2ea255e01e447bf07fce5aa20ecc4

Tests/Unit/LoRaWan/DevAddrTests.cs
```diff
@@ -3,13 +3,40 @@
...
         private readonly DevAddr subject = new(0xeb6f7bde);
 
+        [Theory]
+        [InlineData(uint.MaxValue)]
+        [InlineData(uint.MinValue)]
+        [InlineData(0xeb6f7bde)]
+        public void Init(uint value)
+        {
+            // arrange
+            var maxValue = new DevAddr(value);
```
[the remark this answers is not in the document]
bastbu: Done in 01e2056508a2ea255e01e447bf07fce5aa20ecc4

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
bastbu: Done in 01e2056508a2ea255e01e447bf07fce5aa20ecc4

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
atifaziz: The upper limit check is fine now but also needs to account for negative values.

atifaziz: Never mind. For some reason, I was seeing an outdated version of the file.

atifaziz (approved):
Awesome.

LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
p-schuler: Works for me. Atif and me had a discussion about the general use of this pattern and one thing that convinces me is that it makes the pattern the same for both, nullable value and ref types. This way it is consistent and you can avoid the use of .Value.

PR #1078, opened 2021-12-15
Fix broken log statements

bastbu:
## What is being addressed

With this PR we fix a couple of broken log statements


p-schuler (approved):
LGTM

Issue #1079, opened 2021-12-15
Tags are not always set when raising metrics


bastbu:
## Expected Behavior

Raising metrics should not throw.

## Current Behavior

When raising some metrics, we do not always have the necessary tags set. This results in failing E2E tests due to missing tags.

https://github.com/Azure/iotedge-lorawan-starterkit/runs/4535964485?check_suite_focus=true

PR #1080, opened 2021-12-15
Remove `ConcentratorId` from `DeviceLoadRequests` custom metric

bastbu:
# PR for issue #1079

## What is being addressed

With this PR we remove tags from the `DeviceLoadRequests` metric, that cannot be known at the time we raise the metric. The only tag raised with this metric is the `GatewayId`, which is always known for code executing in the LNS.


danigian (approved):

p-schuler:
I assume there is no similar concept for accessing logging scopes when raising metrics?

bastbu:
> I assume there is no similar concept for accessing logging scopes when raising metrics? I think it could be useful also to determine the number of times a particular concentrator wins a race. But I think we would have to role our own?

@p-schuler Essentially this is what we're doing with the tag bag - we rely on the same mechanism that is used for logging scopes. We can't guarantee that we know the `ConcentratorId` already in all places where we call the cache - since the custom dimensions need to be statically defined this means that we can only not use the concentrator id tag.

bastbu:
Successful Class C E2E test run here: https://github.com/Azure/iotedge-lorawan-starterkit/runs/4537926427?check_suite_focus=true

Issue #966, opened 2021-12-03
Reuse `FramePort`


atifaziz:
- [ ] Find and update usages in:
`DefaultLoRaDataRequestHandler`, `DownlinkMessageBuilder`, `LoRaCloudToDeviceMessageWrapper`, `LoRaCloudToDeviceMessage`
- [ ] Fix affected tests
- [ ] Add unit tests for byte cast operator in `FramePort`

PR #1088, opened 2021-12-16
Reuse FramePort

MaggieSalak:
# PR for issue #966 

## What is being addressed

Usages of `byte` to represent `FPort` field are being replaced with the `FramePort` primitive.

## How is this addressed

- All usages have been updated to use the dedicate primitive instead of `byte`
- `FramePort` struct has been changed to be an `Enum` extending `byte` to better represent the semantics of the `FPort` field 
- Unit tests for `FramePort` were expanded
- Affected tests failing after the introduced changes have been fixed


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/CommonAPI/LoRaFPort.cs
```diff
@@ -11,7 +11,7 @@ public static class LoRaFPort
         /// <summary>
         /// Starting Fport value reserved for future applications.
         /// </summary>
-        public const byte ReservedForFutureAplications = 224;
+        public const byte ReservedForFutureAplications = 225;
```
MaggieSalak, line 14: According to [LoRaWAN specification](https://lora-alliance.org/wp-content/uploads/2020/11/lorawan1.0.3.pdf), "FPort values 225..255 (0xE1..0xFF) are reserved for future standardized application 
extensions."

atifaziz (approved):
> Most usages have been updated, with the exception of `LoRaPayloadDecoder` and `LoRaPayloadData` classes. These are tracked in story [Reuse `FramePort` in remaining classes #1092](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1092). The story has been split because replacing all usages resulted in ~100 unit tests and a number of integration tests failing, with many failures rather complex to fix. This branch contains previously made changes that address all usages of `FPort`, with the failing tests. Because of the complexity of the changes and lack of clarity how the tests can be fixed, that branch has been abandoned in favor of this PR and the story has been split.

I think we can roll all the changes in this PR without much difficulty, with all tests passing, instead of splitting into separate stories. Please see and review my updates.

Issue #1072, opened 2021-12-15
Test LoraDevice Cache uses too high timestamps


Mandur:
Task.Delay accept a timestamp as input, however capped to max int. We should change the default values of the loradevice cacheused for tests to maxint otherwise we will face the following issue: 
![image](https://user-images.githubusercontent.com/8555833/146149404-1cc8025d-b58c-435f-887b-32a4d98ac74a.png)

e.g messageprocessortestbase

PR #1089, opened 2021-12-16
Changing TimeSpan.MaxValue to TimeSpan.FromMilliseconds(int.MaxValue)

danigian:
# PR for issue #1072 

## What is being addressed

Changing TimeSpan.MaxValue to TimeSpan.FromMilliseconds(int.MaxValue) to avoid the issue reported in bug


Mandur (approved):
Thanks

Issue #1082, opened 2021-12-16
`LnsStationConverter.RegionConfigurationConverter` mutates state in `AS923` region singleton


bastbu:
## Expected Behavior

When creating different `AS923` regions for different concentrators using `RegionConfigurationConverter`, creating a region instance with a certain frequency offset setting should not affect other "in-flight" packets that use the same region on a different concentrator.

## Current Behavior

When creating different `AS923` regions for different concentrators with different `Sx1301Config` settings, having different frequency offsets in one concentrator affects other concentrators through mutation of state in a singleton.

PR #1093, opened 2021-12-16
Introducing Region as record for supporting non-destructive mutations

danigian:
# PR for issue #1082 

## What is being addressed

Region is now a record and all its properties are only settable at initialization. In the case of AS923, the WithFrequencyOffset is creating a copy of the region record for a non-destructive mutation of the frequencyoffset field

### DIFF
--- a/Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
+++ b/Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
@@ -580,5 +580,110 @@ namespace LoRaWan.Tests.Unit.NetworkServer.BasicsStation.JsonHandlers
             Assert.Equal(typeof(RegionAS923), region.GetType());
             Assert.Equal(exptectedOffset, ((RegionAS923)region).FrequencyOffset);
         }
+
+        [Fact]
+        public void RegionConfigurationConverter_CorrectlyReadsAS923Region_WithOffset_WithoutMutatingExistingOnes()
+        {
+            var input = @"{{
+            'msgtype': 'router_config',
+            'NetID': [1],
+            'JoinEui': [[0, 18446744073709551615]],
+	        'region': 'AS923',
+	        'hwspec': 'sx1301/1',
+	        'freq_range': [ 920000000, 925000000 ],
+            'DRs': [ [ 11, 125, 0 ],
+                     [ 10, 125, 0 ],
+                     [ 9, 125, 0 ],
+                     [ 8, 125, 0 ],
+                     [ 7, 125, 0 ],
+                     [ 7, 250, 0 ] ],
+            'sx1301_conf': [
+                {{
+                    'radio_0': {{
+                        'enable': true,
+                        'type': 'SX1257',
+                        'freq': {0},
+                        'rssi_offset': -166.0,
+                        'tx_enable': true,
+                        'tx_freq_min': 920000000,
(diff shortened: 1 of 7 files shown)


Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
```diff
@@ -580,5 +580,110 @@ public void RegionConfigurationConverter_Throws_OnNotSetRegion()
...
+            'nodwell': true}}";
+
+            var config1 = JsonUtil.Strictify(string.Format(CultureInfo.InvariantCulture, input, 923000000, 922000000, 200000, 400000));
+            var region1 = LnsStationConfiguration.GetRegion(config1);
+
+            var config2 = JsonUtil.Strictify(string.Format(CultureInfo.InvariantCulture, input, 923000000, 922000000, -1600000, -1400000));
+            var region2 = LnsStationConfiguration.GetRegion(config2);
+
+            // asserting that reading region2 is not affecting region1
+            Assert.Equal(typeof(RegionAS923), region1.GetType());
```
bastbu, line 683: Consider using `Assert.IsType` and leveraging the return value to continue with further assertions (without needing to cast)

Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
```diff
@@ -580,5 +580,110 @@ public void RegionConfigurationConverter_Throws_OnNotSetRegion()
...
+                     [ 9, 125, 0 ],
+                     [ 8, 125, 0 ],
+                     [ 7, 125, 0 ],
+                     [ 7, 250, 0 ] ],
+            'sx1301_conf': [
+                {{
+                    'radio_0': {{
+                        'enable': true,
+                        'type': 'SX1257',
+                        'freq': {0},
```
bastbu, line 605: [nit] we might want to think about integrating this into the helper methods that we already have to replace tokens in that template. In that case, if the format changes, we don't have to fix it all over the place but just in a single place.

bastbu (approved):

Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
danigian: Done

Tests/Unit/NetworkServer/JsonHandlers/LnsStationConfigurationTests.cs
danigian: Even though I agree with you, I think we should not do this as part of this PR.

Could you please discuss the idea tomorrow in the daily and eventually create a new issue, @bastbu? Thanks 👍

atifaziz (commented):
While I see the rationale/motivation, I think it's dangerous to make `Region` a record if the other objects it holds like `RX1DROffsetTable` are mutable. The object is only partially immutable and might lead to subtle bugs.

danigian:
After discussing with @atifaziz regarding his comment on underlying mutable properties, I decided to put back this PR in draft.

Currently Dictionaries like 'DRtoConfiguration' are exposed as public properties. We should either make this Dictionary readonly collections or we should make them not public and only provide methods for accessing what needs to be accessed.

danigian:
#1119 should fix the issue related to this pr, therefore i am closing this one

PR #1094, opened 2021-12-16
added new NOTICE file LBS_NOTICE 40857

roel4ez:


Mandur:
Here something is wrong I guess ;)

Issue #1099, opened 2021-12-16
Resubmitted unconfirmed first messages should not be processed


spygi:
## Current behavior

Currently resubmitted (same concentrator), unconfirmed messages are handled differently based on if they are the first message or subsequent ones. Specifically we allow first messages to be processed more than once, while this is not the case for subsequent messages.

## Expected behavior

Unconfirmed messages coming from the same concentrator should not be processed as this indicates a faulty concentrator or a replay attack. 

Things to consider:
- ABPRelaxed: I don't see how this setting would explain this scenario where the same concentrator resends the same message.
- Strategy Mark or None: do we want to re-send this message upstream? Again, this message has been seen from the same concentrator before. NB: that currently we send upstream 2 times **confirmed** messages from the same concentrator so we should consider changing this as well in order to be consistent.

spygi:
We decided with Patrick to not do this: we have no reliable way to know whether the device gets restarted after sending the first message and sends the same measurement again within the cache retention window (1 minute) - this would be a new message that needs to be processed upstream and not dropped. Deduplication in this (rather unlikely) scenario is "best effort".

We decided to document this scenario in the test cases.

PR #1100, opened 2021-12-16
Ensures that unconfirmed, resubmissions are sent upstream only once.

spygi:
# PR for issue #1099 

Currently resubmitted (same concentrator), unconfirmed messages are handled differently based on if they are the first message or subsequent ones. Specifically we allow first messages to be processed more than once, while this is not the case for subsequent messages.

This PR addresses this by marking immediately unconfirmed messages that are resubmitted from the same concentrator as duplicates. Integration tests are passing.

## TODO
- [ ] Add a unit test covering this case.


spygi:
We decided with @p-schuler to drop this change: if the device gets restarted after sending the first message and sends the same measurement again within the cache window (1 minute), the message would be dropped with this PR which should not be the case. Closing this PR and will document this scenario in the test cases with a different PR.

PR #1102, opened 2021-12-16
Feature/1091 test otaa device

spygi:
# PR for issue #1091

## What is being addressed

Adds an explicit test for OTAA devices: `When_First_Unconfirmed_Data_Message_With_OTAA_Device_Test_All_Different_DeduplicationModes`.

Main difference from the ABP counterpart `When_Unconfirmed_First_Data_Message_Test_All_Different_DeduplicationModes` is the number of frame counter resets (0 instead of 1).

Also note that in the case of resubmission, bundler calls, messages up and twin saves are 1 (instead of 2 in the Mark and None for ABP). Imo this is more correct behavior as explained in #1099 as well.


Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
```diff
@@ -238,12 +238,55 @@ public ConcentratorDeduplicationDataMessagesIntegrationTests()
...
+        [DeduplicationTestData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", DeduplicationMode.Drop, expectedFrameCounterResets: 0, expectedBundlerCalls: 1, expectedFrameCounterDownCalls: 0, expectedMessagesUp: 1, expectedMessagesDown: 0, expectedTwinSaves: 1)] // duplicate
+        [DeduplicationTestData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", DeduplicationMode.Mark, expectedFrameCounterResets: 0, expectedBundlerCalls: 1, expectedFrameCounterDownCalls: 0, expectedMessagesUp: 2, expectedMessagesDown: 0, expectedTwinSaves: 2)] // soft duplicate
+        [DeduplicationTestData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", DeduplicationMode.None, expectedFrameCounterResets: 0, expectedBundlerCalls: 1, expectedFrameCounterDownCalls: 0, expectedMessagesUp: 2, expectedMessagesDown: 0, expectedTwinSaves: 2)] // soft duplicate
+        public async Task When_First_Unconfirmed_Data_Message_With_OTAA_Device_Test_All_Different_DeduplicationModes(
```
MaggieSalak, line 254: [nit] Should it say When_Unconfirmed_First...`, for consistency with the previous test?

MaggieSalak, line 254: All cases for this test have the same values for `expectedFrameCounterResets`, `expectedBundlerCalls` and  `expectedFrameCounterDownCalls` parameters. Maybe let's remove them and use variables inside the test method? I think it would make the test a bit easier to read.

Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
```diff
@@ -238,12 +238,55 @@ public ConcentratorDeduplicationDataMessagesIntegrationTests()
...
+            int expectedTwinSaves)
+        {
+            var simulatedOTAADevice = new SimulatedDevice(TestDeviceInfo.CreateOTAADevice(0)) { DevAddr = "00000000" };
+
+            var dataPayload = simulatedOTAADevice.CreateUnconfirmedDataUpMessage("payload");
+            var request1 = CreateOTAARequest(dataPayload, station1);
+            var request2 = CreateOTAARequest(dataPayload, station2);
+
+            using var loraOTAADevice = new LoRaDevice(simulatedOTAADevice.DevAddr, simulatedOTAADevice.DevEUI, ConnectionManager);
+            loraOTAADevice.AppKey = "00000000000000000000000000000000";
```
MaggieSalak, line 272: Can we use a variable for this value? I think the same is used on line 283.

MaggieSalak (approved):

MaggieSalak:
Approved with a few small comments.

Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
spygi: Done in 5481898a7b90c02f8cb07fd95362b7d8c885dc19

Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
spygi: I think the opposite actually because you would have to look up the shared value for some of the parameters while for others you would have the number directly on them. Also this would introduce another point of difference with the rest of the test methods that would make hard to spot differences between them.

 I will resolve this but if you disagree we can discuss further.

spygi: changed test to consistent names in 5481898a7b90c02f8cb07fd95362b7d8c885dc19

Issue #1081, opened 2021-12-16
`LoRaDeviceCacheTest.When_Refresh_Fails_It_Is_Retried` intermittent failures


bastbu:
## Expected Behavior

When executing `LoRaDeviceCacheTest.When_Refresh_Fails_It_Is_Retried` multiple times, the test never fails.

## Current Behavior

When executing `LoRaDeviceCacheTest.When_Refresh_Fails_It_Is_Retried`, the unit test sometimes fails.

```
﻿ LoRaWan.Tests.Unit.NetworkServer.LoRaDeviceCacheTest.When_Refresh_Fails_It_Is_Retried
   Source: LoRaDeviceCacheTest.cs line 86
   Duration: 285 ms

  Message: 
    Moq.MockException : 
    Expected invocation on the mock at least 2 times, but was 1 times: x => x.InitializeAsync(It.IsAny<NetworkServerConfiguration>(), It.IsAny<CancellationToken>())
    
    Performed invocations:
    
       Mock<LoRaDevice:6> (x):
    
          LoRaDevice.InitializeAsync(NetworkServerConfiguration, CancellationToken)
    

  Stack Trace: 
    Mock.Verify(Mock mock, LambdaExpression expression, Times times, String failMessage) line 337
    Mock`1.Verify[TResult](Expression`1 expression, Times times) line 835
    LoRaDeviceCacheTest.When_Refresh_Fails_It_Is_Retried() line 98
    --- End of stack trace from previous location ---
```

PR #1106, opened 2021-12-16
fix flaky tests by allowing a more deterministic way to wait for the …

p-schuler:
# PR for issue #1081 

## What is being addressed

Tests were subject to fail in resource constraint environments.

## How is this addressed

Adding a more deterministic way of waiting for refreshes to happen. They can still timeout (5s after the refresh cycle calculation) but should be more reliable, and in cases where the machine is running normally, they should pass quicker.

closing #1081


Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
```diff
@@ -381,6 +384,20 @@ public TestDeviceCache(Action<LoRaDevice> onRefreshDevice)
             internal async Task WaitForRefreshAsync(CancellationToken cancellationToken) =>
                 await this.refreshTick.WaitAsync(cancellationToken);
 
+            internal async Task WaitForRefreshCallsAsync(int numberOfCalls, TimeSpan timeout = default)
```
bastbu, line 387: Maybe this might be a candidate for reusing `MoqExtensions.RetryVerifyAsync`?

bastbu (approved):

Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
p-schuler: I would agree if we didn't had the explicit support in the test class to verify for the refreshes. The semaphore signaling makes it more deterministic. If the machine is not overloaded, the timeouts won't kick in and we have a quicker test run.

PR #1113, opened 2021-12-17
Fix & improve URL handling in `LoRaPayloadDecoder`

atifaziz:
## What is being addressed

Safety and correctness of decoder URL detection, manipulation and handling in `LoRaPayloadDecoder`.

## How is this addressed

Standard .NET types are used to parse the decoder URL, extract and manipulate the query string, and re-encode it.

Some problems fixed are:

- The URL scheme is correctly detected (previously it was a containment check which means `http://` could appear anywhere in the string).
- The trailing `/` in the path should not be removed and could have different semantics on the server.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
@@ -56,7 +56,19 @@ namespace LoRaWan.NetworkServer
             var base64Payload = ((payload?.Length ?? 0) == 0) ? string.Empty : Convert.ToBase64String(payload);
 
             // Call local decoder (no "http://" in SensorDecoder)
-            if (!sensorDecoder.Contains("http://", StringComparison.Ordinal))
+            if (Uri.TryCreate(sensorDecoder, UriKind.Absolute, out var url) && (url.Scheme is "http" or "https"))
+            {
+                // Support decoders that have a parameter in the URL
+                // http://decoder/api/sampleDecoder?x=1 -> should become http://decoder/api/sampleDecoder?x=1&devEUI=11&fport=1&payload=12345
+
+                var query = HttpUtility.ParseQueryString(url.Query);
+                query["devEUI"] = devEUI;
+                query["fport"] = fport.ToString(CultureInfo.InvariantCulture);
+                query["payload"] = base64Payload;
+
+                return await CallSensorDecoderModule(new UriBuilder(url) { Query = query.ToString() }.Uri);
+            }
+            else
             {
                 var decoderType = typeof(LoRaPayloadDecoder);
                 var toInvoke = decoderType.GetMethod(sensorDecoder, BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.IgnoreCase);
@@ -72,31 +84,6 @@ namespace LoRaWan.NetworkServer
                         Error = $"'{sensorDecoder}' decoder not found",
                     };
                 }
-            }
-            else
-            {
-                // Call SensorDecoderModule hosted in seperate container ("http://" in SensorDecoder)
-                // Format: http://containername/api/decodername
-                var toCall = sensorDecoder;
-
-                if (sensorDecoder.EndsWith("/", StringComparison.Ordinal))
-                {
-                    toCall = sensorDecoder[..^1];
-                }
-
-                // Support decoders that have a parameter in the URL
-                // http://decoder/api/sampleDecoder?x=1 -> should become http://decoder/api/sampleDecoder?x=1&devEUI=11&fport=1&payload=12345
-                var queryStringParamSeparator = toCall.Contains('?', StringComparison.Ordinal) ? "&" : "?";
-
-                // use HttpUtility to UrlEncode Fport and payload
-                var payloadEncoded = HttpUtility.UrlEncode(base64Payload);
-                var devEUIEncoded = HttpUtility.UrlEncode(devEUI);
-
-                // Add Fport and Payload to URL
-                var url = new Uri($"{toCall}{queryStringParamSeparator}devEUI={devEUIEncoded}&fport={fport}&payload={payloadEncoded}");
-
-                // Call SensorDecoderModule
-                return await CallSensorDecoderModule(url);
             }
         }
 


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
```diff
@@ -56,7 +56,19 @@ public async ValueTask<DecodePayloadResult> DecodeMessageAsync(string devEUI, by
             var base64Payload = ((payload?.Length ?? 0) == 0) ? string.Empty : Convert.ToBase64String(payload);
 
             // Call local decoder (no "http://" in SensorDecoder)
-            if (!sensorDecoder.Contains("http://", StringComparison.Ordinal))
+            if (Uri.TryCreate(sensorDecoder, UriKind.Absolute, out var url) && (url.Scheme is "http" or "https"))
```
spygi, line 59: Was it intended to allow https now?

atifaziz: I found it harmless to add but for sake of 100% backward compatibility, I could remove it. What's your take?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
```diff
@@ -56,7 +56,19 @@ public async ValueTask<DecodePayloadResult> DecodeMessageAsync(string devEUI, by
...
+            {
+                // Support decoders that have a parameter in the URL
+                // http://decoder/api/sampleDecoder?x=1 -> should become http://decoder/api/sampleDecoder?x=1&devEUI=11&fport=1&payload=12345
+
+                var query = HttpUtility.ParseQueryString(url.Query);
+                query["devEUI"] = devEUI;
+                query["fport"] = fport.ToString(CultureInfo.InvariantCulture);
+                query["payload"] = base64Payload;
```
spygi, line 67: I was checking if the `Query` class does the same encoding by default as this was the previous implementation.  According to [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.query?view=net-6.0) they do RFC2396 - not sure if this the case also for HttpUtility.UrlEncode? Checking for example `HttpUtility.UrlEncode("{hello:world}");` vs `query["foo"] = "{hello:world}";` yields different results. 

Not sure if this could be a problem for the payload (not sure what characters are sent). Do you think we have sufficient test coverage here to trust the tests? Should we run an E2E test for this?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
spygi: Can't see a problem with being more permissive and allow https unless we have trouble handling it. but also not very familiar with different decoders so can't give a very educated opinion here :)

spygi (commented):
Looks ok and is definitely an improvement but I would run E2E tests I think on this one since it affects the message we send upstream to IoTHub

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
atifaziz: Given this test:

```c#
var url = new Uri("http://example.com/");
var query = HttpUtility.ParseQueryString(url.Query);
const string foo = "{hello:world}";
query["foo"] = foo;
Console.WriteLine(query.ToString());
Console.WriteLine(HttpUtility.UrlEncode(foo));
```

I see no difference in how the _value_ of `foo` is encoded in either case. Output of [running the above](https://dotnetfiddle.net/gEQ3EI):

    foo=%7bhello%3aworld%7d
    %7bhello%3aworld%7d

atifaziz:
> I think on this one since it affects the message we send upstream to IoTHub

Do you think that's still the case given that [I am not seeing any difference](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1113#discussion_r771256961)?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
spygi: This is what I was checking
```
var a = "http://hello/world?x=1/";
Uri.TryCreate(a, UriKind.Absolute, out var url);
var query = HttpUtility.ParseQueryString(url.Query);
const string foo = "{hello:world}";
query["foo"] = foo;
new UriBuilder(url) { Query = query.ToString() }.Uri.Dump();

HttpUtility.UrlEncode(foo).Dump();
```

spygi: As discussed with Atif, our understanding is that `query.ToString()` over-escapes and that {} are not reserved characters anyway. Since we are passing a "strong" object Uri and also by trying [the rest of reserved characters](https://datatracker.ietf.org/doc/html/rfc2396#section-3.4), we think this usage is ok.

spygi (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
atifaziz: Reverted to HTTP-only with 75af256a04566d254f0548123e2e7aaf939d99c9.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaPayloadDecoder.cs
atifaziz: > Regarding the trailing slash, we discussed that it would be better to remove here as well in order to be backwards compatible.

Reverted with 2d222b0d0a88142b33dd298f27c787bd47cc88f9.

Issue #1069, opened 2021-12-14
Add Dwell Time support for AS923


bastbu:
## Description

Based on the ADR on dwell time support for AS923, we should implement the mechanism that supports dwell time limitations in that region.

Configuration structure:
  - Concentrator configuration:
    - Default dwell time settings [required for AS923]
    - Desired dwell time settings [required for AS923]
  - Device configuration:
    - Reported dwell time settings

When a concentrator receives an uplink message from an end device, it does the following comparisons:
- If the region does not support dwell time limitations -> noop
- If the region does support dwell time limitations, it compares the desired dwell time settings from the concentrator configuration with the reported properties of the end device.
  - If the reported dwell settings of the device match the desired dwell time settings of the concentrator -> noop
  - If the two settings do not match, it sends down a `TxParamSetupReq` MAC command to communicate the actual dwell time settings with the end device, using the last reported dwell time setting of the device
  - If the dwell time settings are not present on the device properties, it sends down a `TxParamSetupReq` MAC command, using the default dwell time settings of the concentrator

If the concentrator receives a `TxParamSetupAns` MAC answer, it updates the reported properties of the device with the dwell time settings that are present on the concentrator configuration (since they were acknowledged by the end device). Potentially, we have a race there: if the concentrator dwell time configuration was updated between sending the downstream `TxParamSetupReq` and the `TxParamSetupAns`, we could store a false dwell time configuration in the reported properties of the device twin.

## Acceptance criteria

- [x] Each station has an (optional) configuration value for the dwell time limitations settings of that region
- [x] `TxParamSetupReq` MAC Command is supported
- [x] `TxParamSetupAns` responses are tracked in the device twin reported properties
- [x] `TxParamSetupReq` MAC Command is sent automatically downstream if the `TxParamSetupAns` acknowledgement does not match the concentrator dwell time configuration
- [x] Documentation around the Dwell Time support for AS923
- [x] CLI is updated

bastbu:
CLI did not need to be updated, as we do not support AS923 from the CLI yet

PR #1119, opened 2021-12-17
Dwell time support for AS923

bastbu:
# PR for issue #1069

## What is being addressed

With this PR we bring in a mechanism to support dwell time limitations for regions such as AS923.

## How is this addressed

We make use of the concept of a `Region` as a location which contains concentrator-specific configuration. The core changes are:
- Support of configuration (required for regions where we have potential dwell time limitations) of desired dwell time settings of the following format (stored in the device twin of a concentrator):
```
"desiredTxParams": {
    "uplinkDwellLimit": true,
    "downlinkDwellLimit": true,
    "eirp": 4
}
```
Default dwell time settings are defined as a readonly on each dwell time limited region. If a device does not comply with the default regional settings, it needs to reflect its actual dwell time settings in the reported properties.
- Support of downlink dwell time limitation based on a device's reported properties and the default dwell time settings for a concentrator
- Support of a `TxParamSetupReq` MAC command to change the end device dwell time settings and reaction to a subsequent `TxParamSetupAns` MAC command. We send `TxParamSetupReq` MAC commands downstream by reusing the same mechanism we do for decoders.

We do not apply the dwell time settings on regions that are created for the C2D and join request message flow: for C2D we always use RX2 (which has minimum DR2), and join requests are always using DR2 - DR5 (compliant with any dwell time limitations)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -242,6 +244,27 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
...
+                                    _ = await loRaDevice.SaveChangesAsync(force: true);
+                                }
+                                else
+                                {
+                                    this.logger.LogDebug("Received 'TxParamSetupAns' even though reported dwell time settings match desired dwell time settings.");
+                                }
+                            }
+                            else
+                            {
+                                throw new LoRaProcessingException($"Received 'TxParamSetupAns' in region '{request.Region.LoRaRegion}' which does not support dwell limitations.");
```
Mandur, line 265: Is this from the spec that this sis forbidden? otherwise I would simply log

Mandur (approved):
It looks good to me

Mandur:
Would be great to test it with the new GW in january

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: No I didn't find it anywhere in the spec - will do that and replace the exception with a log statement

bastbu: Done in 9cfe612c3fd6145065e3c07c01221603f1038bdf

Tests/Unit/LoRaTools/RegionAS923TestData.cs
```diff
@@ -59,7 +67,7 @@ public static IEnumerable<object[]> TestRegionFrequencyData()
            {
                new object[] { region, 1, 8 },
                new object[] { region, 1, 9 },
-               new object[] { regionWithDwellTime, 1, 10 },
+               new object[] { regionWithDwellTime, 2, 10 },
```
MaggieSalak, line 70: Just curious, was this test case invalid? Similar for deleted lines 46-47.

MaggieSalak (approved):

Tests/Unit/LoRaTools/RegionAS923TestData.cs
bastbu: Nice catch. My original reasoning was that since for downstream data rates, due to dwell time limits enforce DR2 and higher that this would automatically apply for upstream data rates, since dwell time limits both upstream and downstream. Since this is not specified explicitly, I reverted that assumption in fb5d856e5c1bca8aad8d8fd90102f14fbfb7d1ce, but I forgot to bring back the original test cases. Solved this in 033e20937e3c79c335e5bd7c7fd83f6f24c26435

bastbu:
Tested the dwell time settings with @danigian in region AS923-3. We were able to observe the `TxParamSetupReq` MAC command that was sent downstream to the LoRa device, and the LoRa device was acknowledging with a `TxParamSetupAns`. However, the answers were piggybacked on the payload for the device with which we tested it and we need to solve #1120 to handle these cases as well

PR #1122, opened 2021-12-19
Remove RXPK: Round I

Mandur:
# PR for issue #1128

We are now migrating the solution to not use the RXPK class, but a native basicsstation format.
The TXPK is fully cleaned out but the RXPK will stay until the simulator is migrated.

I reverted the changes of RSSI to double and added #1126


LoRaEngine/LoraKeysManagerFacade/CheckPreferredGateway/PreferredGatewayTableItem.cs
```diff
@@ -7,9 +7,9 @@ public class PreferredGatewayTableItem
     {
         public string GatewayID { get; private set; }
 
-        public int Rssi { get; private set; }
+        public double Rssi { get; private set; }
```
Mandur, line 10: I changed this to double as it comes as a double from the basicstation.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
```diff
@@ -33,25 +33,24 @@ internal class DownstreamSender : IPacketForwarder
...
             if (message is null) throw new ArgumentNullException(nameof(message));
             if (message.StationEui == default) throw new ArgumentException($"A proper StationEui needs to be set. Received '{message.StationEui}'.");
 
             if (this.socketWriterRegistry.TryGetHandle(message.StationEui, out var webSocketWriterHandle))
             {
-                var region = await this.basicsStationConfigurationService.GetRegionAsync(message.StationEui, CancellationToken.None);
-                var payload = Message(message, region);
-                LogSendingMessage(this.logger, message.StationEui, message.Txpk.Data, null);
+                var payload = Message(message);
+                LogSendingMessage(this.logger, message.StationEui, ConversionHelper.ByteArrayToString(message.Data), null);
```
Mandur, line 44: Probably there is too much logs, we could merge tis one with the log L76 on the same file

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/RadioMetadata.cs
```diff
@@ -3,35 +3,7 @@
 
 namespace LoRaWan.NetworkServer.BasicsStation
 {
-    public class RadioMetadata
-    {
-        public DataRateIndex DataRate { get; }
-        public Hertz Frequency { get; }
-        public RadioMetadataUpInfo UpInfo { get; }
+    public record RadioMetadata(DataRateIndex DataRate, Hertz Frequency, RadioMetadataUpInfo UpInfo);
```
Mandur, line 6: @atifaziz could you tell me it this codeQL warning is appropriate? am I missusing records?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -221,16 +207,35 @@ internal static class DownlinkMessageBuilder
                 loRaDevice.Supports32BitFCnt ? fcntDown : null);
 
             // todo: check the device twin preference if using confirmed or unconfirmed down
-            logger.LogInformation($"sending a downstream message with ID {ConversionHelper.ByteArrayToString(rndToken)}");
-
-            var downlinkPktFwdMessage = ackLoRaMessage.Serialize(loRaDevice.AppSKey, loRaDevice.NwkSKey, datr, freq, tmst, loRaDevice.DevEUI, lnsRxDelay, rxpk.Rfch, rxpk.Time, request.StationEui);
+            var downlinkPktFwdMessage = BuildDownstreamMessage(loRaDevice, request.StationEui, logger, radioMetadata.UpInfo.Xtime, datr, loRaRegion.GetDownstreamRX2DataRate(configuration.Rx2DataRate, loRaDevice.ReportedRX2DataRate, logger, deviceJoinInfo), freq, loRaRegion.GetDownstreamRX2Freq(configuration.Rx2Frequency, logger, deviceJoinInfo), lnsRxDelay, ackLoRaMessage, radioMetadata.UpInfo.AntennaPreference);
```
Mandur, line 210: remove pktfwd inname

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilderResponse.cs
```diff
@@ -7,12 +7,13 @@ namespace LoRaWan.NetworkServer
 
     public class DownlinkMessageBuilderResponse
     {
-        internal DownlinkPktFwdMessage DownlinkPktFwdMessage { get; set; }
+        internal DownlinkBasicsStationMessage DownlinkPktFwdMessage { get; set; }
```
Mandur, line 10: pkt fwd in name

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -290,23 +282,35 @@ internal async Task ProcessJoinRequestAsync(LoRaRequest request)
                     loraSpecDesiredRxDelay,
                     null);
 
-                var joinAccept = loRaPayloadJoinAccept.Serialize(loRaDevice.AppKey, datr, freq, devEUI, tmst, lnsRxDelay, request.Rxpk.Rfch, request.Rxpk.Time, request.StationEui);
-                if (joinAccept != null)
+                var joinAcceptBytes = loRaPayloadJoinAccept.Serialize(loRaDevice.AppKey);
+                var downlinkPktFwdMessage = new DownlinkBasicsStationMessage(
+                  joinAcceptBytes,
+                  request.RadioMetadata.UpInfo.Xtime,
+                  loraRegion.GetDownstreamDataRate(request.RadioMetadata.DataRate, loRaDevice.ReportedRX1DROffset),
```
Mandur, line 289: this should be changed to datr

Mandur: No that is ok

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -206,35 +205,28 @@ internal async Task ProcessJoinRequestAsync(LoRaRequest request)
...
 
                     // set tmst for the normal case
-                    tmst = request.Rxpk.Tmst + (loraRegion.JoinAcceptDelay1 * 1000000);
                     lnsRxDelay = (ushort)loraRegion.JoinAcceptDelay1;
                 }
                 else
                 {
                     this.logger.LogDebug("processing of the join request took too long, using second join accept receive window");
-                    tmst = request.Rxpk.Tmst + (loraRegion.JoinAcceptDelay2 * 1000000);
                     lnsRxDelay = (ushort)loraRegion.JoinAcceptDelay2;
```
Mandur, line 226: all this block can be simplified, let's just keep the lnsrxdealy

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
```diff
@@ -74,34 +72,15 @@ public void DispatchRequest(LoRaRequest request)
...
-                    this.logger.LogError("There was a problem in decoding the Rxpk");
-                    request.NotifyFailed(LoRaDeviceRequestFailedReason.InvalidRxpk);
-                    return;
-                }
-                request.SetPayload(loRaPayload);
+                throw new LoRaProcessingException(nameof(request.Payload), LoRaProcessingErrorCode.PayloadNotSet);
             }
 
-            if (this.loraRegion == null)
+            // I changed the logic here to make the request immediately fails by throwing instead of returning null
```
Mandur, line 78: let me know if that is ok

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -253,18 +252,13 @@ public LoRaPayloadData(MacMessageType messageType, byte[] devAddr, FrameControlF
         /// </summary>
         /// <param name="appSKey">the app key used for encryption.</param>
         /// <param name="nwkSKey">the nwk key used for encryption.</param>
-        /// <param name="datr">the calculated datarate.</param>
-        /// <param name="freq">The frequency at which to be sent.</param>
         /// <param name="tmst">time stamp.</param>
```
Mandur, line 255: Remove old stuff from param list

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
```diff
@@ -315,6 +314,7 @@ private async Task<SearchLogResult> SearchTcpLogs(Func<SearchLogEvent, bool> pre
                 if (i > 0)
                 {
                     var timeToWait = i * Configuration.EnsureHasEventDelayBetweenReadsInSeconds;
+                    TestLogger.Log($"Searching for TCP logs suiting {predicate.Method.GetMethodBody()} ***");
```
Mandur, line 317: this is wrong should be removed

Tests/Common/SimulatedDevice.cs
```diff
@@ -213,86 +220,95 @@ public async Task SendConfirmedMessageAsync(SimulatedPacketForwarder simulatedPa
         }
 
         // Performs join
+        // THE FOLLOWING METHODS
```
Mandur, line 223: Finish this comment

Tests/Integration/CloudToDeviceMessageTests.cs
```diff
@@ -601,23 +572,22 @@ public async Task When_Device_Prefers_Second_Window_Should_Send_Downstream_In_Se
...
-        [InlineData(2, 0, 1400, 1610, 2)]
-        [InlineData(2, 100, 1300, 1510, 2)]
+        [InlineData(2, 0, 1400, 1610)]
+        [InlineData(2, 100, 1300, 1510)]
         public async Task When_Device_Checks_For_C2D_Message_Uses_Available_Time(
             int preferredWindow,
             int sendEventDurationInMs,
             int checkMinDuration,
-            int checkMaxDuration,
-            int expectedRX)
```
Mandur, line 620: this was removed as we can't decide when to send a message, this decision is made by the basicstation

Tests/Unit/NetworkServer/BasicsStation/DownstreamSenderTests.cs
```diff
@@ -77,33 +83,41 @@ public async Task SendDownstreamAsync_Succeeds_WithValidDownlinkMessage_ClassADe
...
-            }
-            else
-            {
-                Assert.DoesNotContain("rctx", actualMessage, StringComparison.InvariantCulture);
-            }
+           
             Assert.Contains(@"""priority"":0", actualMessage, StringComparison.InvariantCulture);
+
+            // THIS NEED TO BE REWORKED
+            // rctx need to be reworked
```
Mandur, line 92: The current processing we are doing with rctx is wrong we need to rework this as part of  #1125

Tests/Unit/NetworkServer/BasicsStation/DownstreamSenderTests.cs
```diff
@@ -121,21 +135,26 @@ public async Task SendDownstreamAsync_Succeeds_WithValidDownlinkMessage_ClassCDe
...
-            if (rfchHasValue)
-            {
-                Assert.Contains(@"""rctx"":1,", actualMessage, StringComparison.InvariantCulture);
-            }
-            else
-            {
-                Assert.DoesNotContain("rctx", actualMessage, StringComparison.InvariantCulture);
-            }
+            // Will select DR0 as it is the second DR.
+            Assert.Contains(@"""RX2DR"":0,", actualMessage, StringComparison.InvariantCulture);
```
Mandur, line 139: This doesn't match previous values, why?

Tests/Unit/NetworkServer/MessageProcessorJoinTest.cs
```diff
@@ -669,14 +656,15 @@ public async Task When_Join_With_Custom_Join_Update_Old_Desired_Properties()
...
+        // In case the offset is invalid, we rollback to offset 0. for europe that means = to upstream
         [InlineData(0, DR2)]
         [InlineData(1, DR1)]
         [InlineData(2, DR0)]
         [InlineData(3, DR0)]
         [InlineData(4, DR0)]
-        [InlineData(6, DR0)]
-        [InlineData(12, DR0)]
-        [InlineData(-2, DR0)]
+        [InlineData(6, DR2)]
```
Mandur, line 665: Why?

Mandur: All thise offsets are invalid, that means we now threat as if there was no offseett. Previous logic seems to have defaulted to DR0?

Mandur: Actually, the expected values were never used on the left and the test was rather using txpk == rxpk. This is performing as expected

Tests/Unit/NetworkServer/MessageProcessorSingleGatewayTest.cs
```diff
@@ -220,14 +215,17 @@ public void When_Faulty_MAC_Message_Is_Received_Processing_Abort_Without_Infinit
                 FrameCounterUpdateStrategyProvider);
 
             // TODO This will be moved in #1086
```
Mandur, line 217: remove

LoRaEngine/LoraKeysManagerFacade/CheckPreferredGateway/PreferredGatewayTableItem.cs
Mandur: Following feedback from @atifaziz this was removed

Tests/Unit/NetworkServer/BasicsStation/DownstreamSenderTests.cs
Mandur: in the previous implementation we never differentiated between rx1dr and rx2dr, therefore before the payload was set to DR 5 on both rates. Here Rx1 is set to DR5 but rxwin2 is set to DR0 on the constructor

Tests/Unit/NetworkServer/BasicsStation/DownstreamSenderTests.cs
```diff
@@ -121,21 +135,26 @@ public async Task SendDownstreamAsync_Succeeds_WithValidDownlinkMessage_ClassCDe
...
+            Assert.Contains(@"""RX2Freq"":869500000,", actualMessage, StringComparison.InvariantCulture);
```
Mandur, line 140: As before we now have a frequency for rx2 as weel

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/RadioMetadata.cs
atifaziz: Sorry, I don't see the warning. Where should I be looking?

Tests/Common/LoRaPayloadExtensions.cs
```diff
@@ -8,16 +8,6 @@ namespace LoRaWan.Tests.Common
 
     public static class LoRaPayloadExtensions
```
atifaziz, line 9: Seems this whole class/file can be removed since in nothing in here used now?

Tests/Common/TestUtils.cs
```diff
@@ -302,18 +304,39 @@ public static void StartBasicsStation(TestConfiguration config, Dictionary<strin
...
-        public static DataRateIndex GetDataRateIndex(this Region region, DataRate datr) =>
-             region.DRtoConfiguration.FirstOrDefault(x => x.Value.DataRate == datr) is (var index, (not null, _)) ? index : throw new KeyNotFoundException("");
-
         public static RadioMetadata GenerateTestRadioMetadata(
                 DataRateIndex dataRate = DataRateIndex.DR2,
                 Hertz? frequency = null,
                 uint antennaPreference = 1,
                 ulong xtime = 100000,
-                uint gpstime = 0,
+                uint gpstime = 100000,
```
atifaziz, line 312: Why this change?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
```diff
@@ -0,0 +1,64 @@
...
+        public Hertz FrequencyRx2 { get; }
+
+#pragma warning disable CA1819 // Properties should not return arrays
+        public byte[] Data { get; }
+#pragma warning restore CA1819 // Properties should not return arrays
+
+
+        public StationEui StationEui { get; }
+
+        public DownlinkBasicsStationMessage()
```
atifaziz, line 39: Why is an empty constructor allowed? Wouldn't that give an invalid message object?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace LoRaTools.LoRaPhysical
+{
+    using System;
+    using LoRaWan;
+
+    /// <summary>
+    /// JSON of a Downlink message for the Basic Station.
+    /// </summary>
+    public class DownlinkBasicsStationMessage
```
atifaziz, line 12: Why not just call this `DownlinkMessage`?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaTools.LoRaPhysical
+{
+    using System;
+    using LoRaWan;
+
+    /// <summary>
+    /// JSON of a Downlink message for the Basic Station.
```
atifaziz, line 10: I don't see anything JSON-specific in this class. I believe the real conversion is being done in `DownstreamSender.Message`?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
```diff
@@ -0,0 +1,64 @@
...
+        public DataRateIndex DataRateRx1 { get; }
+
+        public DataRateIndex DataRateRx2 { get; }
+
+        public Hertz FrequencyRx1 { get; }
+
+        public Hertz FrequencyRx2 { get; }
+
+#pragma warning disable CA1819 // Properties should not return arrays
+        public byte[] Data { get; }
```
atifaziz, line 33: Like the warning says, can we not avoid exposing bare arrays?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    using System;
+    using LoRaWan;
+
+    /// <summary>
+    /// JSON of a Downlink message for the Basic Station.
+    /// </summary>
+    public class DownlinkBasicsStationMessage
+    {
+        public string DevEui { get; }
```
atifaziz, line 14: Use `DevEui` for new code?

Tests/Common/TestUtils.cs
Mandur: image
As it's time after Epoch, I imagined 0 was very unlikely but so is any other values

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
Mandur: Was like this in the previous class, I tried to replace as less as possible given size of the PR. This empty constructor is used by tests

Mandur: Trying to replace it quickly led to ~20 integration tests failing, if that is fine with you, I propose we tackle this as a fturue work item

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
Mandur: I changed the comment,

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
Mandur: What would you suggest? Shall I put it here in Hex already (we do the conversion in the message atm)? Could you explain the problem that we could have here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
Mandur: This is not really new code, this is the DownlinkPktFwdMessage being renamed... 
I am fine migrating this to the primitives, however that would cascade modification to a larger amount of files.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/RadioMetadata.cs
atifaziz: I believe those are false positives and can be ignored. With positional records, you are declaring the primary constructor with arguments at the same time as the properties of the record and CodeQL is complaining that your constructor arguments are named the same as the properties and thus shadowing them. There is by-design of the language.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -385,9 +373,9 @@ private static ushort ValidateAndConvert16bitFCnt(uint fcntDown)
                         case Cid.Zero:
                         case Cid.One:
                         case Cid.LinkADRCmd:
-                            if (rxpk != null)
+                            if (loRaRequest != null)
```
atifaziz, line 376: FYI, noticed this code is a duplicate of the above case that I consolidated in #1130.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
atifaziz: I would change this to `ReadOnlyMemory<byte>`, and align the type of `payload` in the constructor accordingly. This way, the whole type is read-only and the byte array cannot be manipulated by anyone else. This change should have minimal impact:

```diff
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
index 98f46c8a..5428030f 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
@@ -29,10 +29,7 @@ namespace LoRaTools.LoRaPhysical
 
         public Hertz FrequencyRx2 { get; }
 
-#pragma warning disable CA1819 // Properties should not return arrays
-        public byte[] Data { get; }
-#pragma warning restore CA1819 // Properties should not return arrays
-
+        public ReadOnlyMemory<byte> Data { get; }
 
         public StationEui StationEui { get; }
 
@@ -45,11 +42,9 @@ namespace LoRaTools.LoRaPhysical
         /// This method is used in case of a response to a upstream message.
         /// </summary>
         /// <returns>DownlinkPktFwdMessage object ready to be sent.</returns>
-        public DownlinkMessage(byte[] payload, ulong xtime, DataRateIndex datrRx1, DataRateIndex datrRx2, Hertz freqRx1, Hertz freqRx2, string devEui, ushort lnsRxDelay = 0, StationEui stationEui = default, uint antennaPreference = 0)
+        public DownlinkMessage(ReadOnlyMemory<byte> payload, ulong xtime, DataRateIndex datrRx1, DataRateIndex datrRx2, Hertz freqRx1, Hertz freqRx2, string devEui, ushort lnsRxDelay = 0, StationEui stationEui = default, uint antennaPreference = 0)
         {
-            if (payload is null) throw new ArgumentNullException(nameof(payload));
             Data = payload;
-
             DevEui = devEui;
             LnsRxDelay = lnsRxDelay;
             AntennaPreference = antennaPreference;
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
index 04ede78c..a014b6eb 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
@@ -67,7 +67,7 @@ namespace LoRaWan.NetworkServer.BasicsStation
             // Getting and writing payload bytes
             var pduBytes = message.Data;
             var pduChars = new char[pduBytes.Length * 2];
-            Hexadecimal.Write(pduBytes, pduChars);
+            Hexadecimal.Write(pduBytes.Span, pduChars);
             writer.WriteString("pdu", pduChars);
 
 #pragma warning disable CA5394 // Do not use insecure randomness. This is fine as not used for any crypto operations.
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public DownlinkMessage()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DownlinkMessage"/> class.
+        /// This method is used in case of a response to a upstream message.
+        /// </summary>
+        /// <returns>DownlinkPktFwdMessage object ready to be sent.</returns>
```
atifaziz, line 47:
```suggestion
        /// <returns><see cref="DownlinkMessage"/> object ready to be sent.</returns>
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
```diff
@@ -0,0 +1,64 @@
...
+    using LoRaWan;
+
+    /// <summary>
+    /// Model class for a Downlink message for the Basic Station.
+    /// </summary>
+    public class DownlinkMessage
+    {
+        public string DevEui { get; }
+
+        /// This rx delay control the time we need to wait before sending.
```
atifaziz, line 16: What's this delay expressed in? Can we not use `TimeSpan` so no one has to guess?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
atifaziz: It's used in only one place in tests (`TestDefaultLoRaRequestHandler`), so why not replace there with a call to the normal constructor with `default` for all arguments? It would have the same effect and your tests should pass.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
Mandur: it's either 1 or 2 (seconds). I can use a timespan if you prefer.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
```diff
@@ -67,32 +65,30 @@ private string Message(DownlinkPktFwdMessage message, Region region)
...
             {
                 writer.WriteNumber("RxDelay", message.LnsRxDelay);
-                writer.WriteNumber("RX1DR", (int)dataRate);
-                writer.WriteNumber("RX1Freq", (ulong)(message.Txpk.Freq * 1e6));
-                writer.WriteNumber("RX2DR", (int)region.GetDefaultRX2ReceiveWindow().DataRate);
-                writer.WriteNumber("RX2Freq", (ulong)region.GetDefaultRX2ReceiveWindow().Frequency);
+                writer.WriteNumber("RX1DR", (int)message.DataRateRx1);
+                writer.WriteNumber("RX1Freq", message.FrequencyRx1.AsUInt64);
+                writer.WriteNumber("RX2DR", (int)message.DataRateRx2);
+                writer.WriteNumber("RX2Freq", message.FrequencyRx2.AsUInt64);
```
atifaziz, line 85: Prefer casting (frequency to `ulong`) consistently:

```suggestion
                writer.WriteNumber("RX1Freq", (ulong)message.FrequencyRx1);
                writer.WriteNumber("RX2DR", (int)message.DataRateRx2);
                writer.WriteNumber("RX2Freq", (ulong)message.FrequencyRx2);
```

I think we'll want to get rid of `AsUInt64` eventually.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
```diff
@@ -67,32 +65,30 @@ private string Message(DownlinkPktFwdMessage message, Region region)
...
                 writer.WriteNumber("xtime", message.Xtime);
             }
             else if (deviceClassType is LoRaDeviceClassType.C)
             {
-                writer.WriteNumber("RX2DR", (int)dataRate);
-                writer.WriteNumber("RX2Freq", (ulong)(message.Txpk.Freq * 1e6));
+                writer.WriteNumber("RX2DR", (int)message.DataRateRx2);
+                writer.WriteNumber("RX2Freq", message.FrequencyRx2.AsUInt64);
```
atifaziz, line 91: Case as in the DR case:

```suggestion
                writer.WriteNumber("RX2Freq", (ulong)message.FrequencyRx2);
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -138,7 +138,7 @@ public static NetworkServerConfiguration CreateFromEnvironmentVariables()
             config.EnableGateway = envVars.GetEnvVar("ENABLE_GATEWAY", config.EnableGateway);
             config.GatewayID = envVars.GetEnvVar("IOTEDGE_DEVICEID", string.Empty);
             config.HttpsProxy = envVars.GetEnvVar("HTTPS_PROXY", string.Empty);
-            config.Rx2DataRate = envVars.GetEnvVar("RX2_DATR", string.Empty);
+            config.Rx2DataRate = envVars.GetEnvVar("RX2_DATR") is { } someDr ? (DataRateIndex)someDr : null;
```
atifaziz, line 141: Nice, but then `GetEnvVar` returns a `double` here. Don't think we want that? How about the following:

```suggestion
            config.Rx2DataRate = envVars.GetEnvVar("RX2_DATR", -1) is var datrNum && (DataRateIndex)datrNum is var datr && Enum.IsDefined(datr) ? datr : null;
```

What's our policy on misconfiguration? I think we should throw if, for example, the data rate is in correct (outside of the 0..15 range). If so, should we open a bug for that and potentially other possible cases of misconfiguration?

Tests/Unit/NetworkServer/LnsProtocolMessageProcessorTests.cs
```diff
@@ -334,7 +323,7 @@ public async Task InternalHandleDataAsync_ShouldProperlyCreateLoraPayloadForUpdf
 
             // assert
             Assert.NotNull(loRaRequest);
-            Assert.True(AreRxpkEqual(loRaRequest.Rxpk, expectedRxpk));
+            Assert.True(RadioMetadata(loRaRequest.RadioMetadata, expectedRadioMetadata));
```
atifaziz, line 326: Since you made `RadioMetadata` a record, you get equality comparison for free:

```suggestion
            Assert.Equal(loRaRequest.RadioMetadata, expectedRadioMetadata);
```

Tests/Unit/NetworkServer/LnsProtocolMessageProcessorTests.cs
```diff
@@ -284,29 +283,19 @@ internal async Task InternalHandleDataAsync_ShouldNotThrow_OnNonHandledMessageTy
...
-            && subject.Time == other.Time
-            && subject.Tmms == other.Tmms
-            && subject.Tmst == other.Tmst;
+        private static bool RadioMetadata(RadioMetadata subject, RadioMetadata other) =>
+            subject.DataRate == other.DataRate
+            && subject.Frequency == other.Frequency
+            && subject.UpInfo.Xtime == other.UpInfo.Xtime
+            && subject.UpInfo.SignalNoiseRatio == other.UpInfo.SignalNoiseRatio
+            && subject.UpInfo.GpsTime == other.UpInfo.GpsTime
+            && subject.UpInfo.ReceivedSignalStrengthIndication == other.UpInfo.ReceivedSignalStrengthIndication;
```
atifaziz, line 299: Since you made `RadioMetadata` a record, you get equality comparison for free so you can completely remove this helper:

```suggestion
```

See and commit together with [my other suggestion](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1122/files#r772525778).

Tests/Unit/NetworkServer/LnsProtocolMessageProcessorTests.cs
```diff
@@ -284,29 +283,19 @@ internal async Task InternalHandleDataAsync_ShouldNotThrow_OnNonHandledMessageTy
                                                                        CancellationToken.None);
         }
 
-        private static Rxpk GetExpectedRxpk()
+        private static RadioMetadata GetExpectedRadioMetadata()
         {
             var radioMetadataUpInfo = new RadioMetadataUpInfo(0, 68116944405337035, 0, -53, (float)8.25);
```
atifaziz, line 288: nit: Prefer type specifier over cast:

```suggestion
            var radioMetadataUpInfo = new RadioMetadataUpInfo(0, 68116944405337035, 0, -53, 8.25f);
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/RadioMetadata.cs
atifaziz, line 6: I would recommend making this a regular record as opposed to a positional record (the latter is good when there's a natural order to fields, but I don't think we can assume that here):

```suggestion
    public sealed record RadioMetadata
    {
        public RadioMetadata(DataRateIndex dataRate, Hertz frequency, RadioMetadataUpInfo upInfo)
        {
            DataRate = dataRate;
            Frequency = frequency;
            UpInfo = upInfo;
        }

        public DataRateIndex DataRate { get; init; }
        public Hertz Frequency { get; init; }
        public RadioMetadataUpInfo UpInfo { get; init; }
    }
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/RadioMetadata.cs
```diff
@@ -3,35 +3,7 @@
...
-            SignalNoiseRatio = signalNoiseRatio;
-        }
-
-        public uint AntennaPreference { get; }
-        public ulong Xtime { get; }
-        public uint GpsTime { get; }
-        public double ReceivedSignalStrengthIndication { get; }
-        public float SignalNoiseRatio { get; }
-    }
+    public record RadioMetadataUpInfo(uint AntennaPreference, ulong Xtime, uint GpsTime, double ReceivedSignalStrengthIndication, float SignalNoiseRatio);
```
atifaziz, line 8: I would recommend making this a regular record as opposed to a positional record (the latter is good when there's a natural order to fields, but I don't think we can assume that here):

```suggestion
    public sealed record RadioMetadataUpInfo
    {
        public RadioMetadataUpInfo(uint antennaPreference, ulong xtime, uint gpsTime,
                                   double receivedSignalStrengthIndication, float signalNoiseRatio)
        {
            AntennaPreference = antennaPreference;
            Xtime = xtime;
            GpsTime = gpsTime;
            ReceivedSignalStrengthIndication = receivedSignalStrengthIndication;
            SignalNoiseRatio = signalNoiseRatio;
        }

        public uint AntennaPreference { get; init; }
        public ulong Xtime { get; init; }
        public uint GpsTime { get; init; }
        public double ReceivedSignalStrengthIndication { get; init; }
        public float SignalNoiseRatio { get; init; }
    }
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
```diff
@@ -85,29 +77,27 @@ public LoRaDeviceTelemetry()
...
-            Tmst = rxpk.Tmst;
             Fcnt = upstreamPayload.GetFcnt();
             Port = upstreamPayload.Fport.Value;
+            Freq = radioMetadata.Frequency.InMega;
+            Datr = datr.ToString();
+            Rssi = radioMetadata.UpInfo.ReceivedSignalStrengthIndication;
+            Rfch = radioMetadata.UpInfo.AntennaPreference;
+            Lsnr = radioMetadata.UpInfo.SignalNoiseRatio;
+            Time = radioMetadata.UpInfo.Xtime.ToString(CultureInfo.InvariantCulture); // This field was just used for telemetry in rxpk. Now it's being used for bringing unaltered the Xtime.
+            Tmst = unchecked((uint)radioMetadata.UpInfo.Xtime); // This is used by former computation only. Should go away when we drop PktFwd support.
```
atifaziz, line 97: Why not change the type of `Tmst` to `ulong` instead of casting here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
```diff
@@ -85,29 +77,27 @@ public LoRaDeviceTelemetry()
...
+            Chan = checked((uint)radioMetadata.DataRate); // This is not used in any computation. It is only reported in the device telemetry.
```
atifaziz, line 98: Checked here but unchecked above? Is there a reason for doing the two differently?

Mandur: I copied this from previous code, we can check tommorow

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
Mandur: As discussed I am not sure timespan is appropriate as only 4 values are possible, let's discuss

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
Mandur: I am wondering if we even want to keep this field as we have both time and xtime... I think we might want to rework that as for example might be better to surface the gpstime...

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
Mandur: this is an optional property, therefore I think setting it to fail when not set I think is not a good idea

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -221,15 +207,33 @@ internal static class DownlinkMessageBuilder
...
-            downlinkPktFwdMessage.DeviceJoinInfo = deviceJoinInfo;
+                logger.LogDebug($"{ackLoRaMessage.MessageType} {JsonConvert.SerializeObject(downlinkMessage)}");
+
+            return new DownlinkMessageBuilderResponse(downlinkMessage, isMessageTooLong, receiveWindow);
+        }
+
+        private static DownlinkMessage BuildDownstreamMessage(LoRaDevice loRaDevice, StationEui stationEUI, ILogger logger, ulong xTime, DataRateIndex rx1Datr, DataRateIndex rx2Datr, Hertz freqRx1, Hertz freqRx2, ushort lnsRxDelay, LoRaPayloadData loRaMessage, uint antennaPreference = 0)
+        {
+            var messageBytes = loRaMessage.Serialize(loRaDevice.AppSKey, loRaDevice.NwkSKey);
+            var downlinkPktFwdMessage = new DownlinkMessage(
```
atifaziz, line 221: nit: rename `downlinkPktFwdMessage` to `downlinkMessage` or just `message`.

Tests/Unit/NetworkServer/MessageProcessorSingleGatewayTest.cs
```diff
@@ -221,14 +215,17 @@ public void When_Faulty_MAC_Message_Is_Received_Processing_Abort_Without_Infinit
...
-
+            //var request = new LoRaRequest(
+            //    new Rxpk
+            //    {
+            //        Data = "QDDaAAGxfh0FAI6wAENHbvgt1UK5Je1uPo/bLPB9HlnOXLGlLRUrTtA0KOHrZhusGl+L4g=="
+            //    },
+            //    null,
+            //    DateTime.Now);
+
+            var payload = new LoRaPayloadDataLns(DevAddr.Parse("0100DA30"), new MacHeader(40), FrameControl.Decode(177).Flags, 7550, "05", "8EB00043476EF82DD542B925ED6E3E8FDB2CF07D1E59CE5CB1A52D152B4ED03428E1EB661BAC", new FramePort(), new Mic(Convert.ToUInt32("1A5F8BE2", 16)), NullLogger.Instance);
```
atifaziz, line 226: Let's use the new types as much as possible:

```suggestion
            var payload = new LoRaPayloadDataLns(new DevAddr(0x0100DA30),
                                                 new MacHeader(MacMessageType.JoinAccept),
                                                 FrameControlFlags.ClassB | FrameControlFlags.Ack | FrameControlFlags.Adr,
                                                 counter: 7550,
                                                 options: "05",
                                                 "8EB00043476EF82DD542B925ED6E3E8FDB2CF07D1E59CE5CB1A52D152B4ED03428E1EB661BAC",
                                                 FramePort.MacCommand,
                                                 new Mic(0x1A5F8BE2),
                                                 NullLogger.Instance);
```

Tests/Unit/NetworkServer/MessageProcessorJoinTest.cs
```diff
@@ -768,24 +755,19 @@ public async Task When_Getting_RX1_Offset_From_Twin_Returns_JoinAccept_With_Corr
...
-#pragma warning restore CS0618 // #655 - This Rxpk based implementation will go away as soon as the complete LNS implementation is done
-            }
-            else
-            {
-#pragma warning disable CS0618 // #655 - This Rxpk based implementation will go away as soon as the complete LNS implementation is done
-                Assert.Equal(confirmedRequest.Region.GetDRFromFreqAndChan(confirmedRequest.Rxpk.Datr), confirmedRequest.Region.GetDRFromFreqAndChan(downstreamMessage.Txpk.Datr));
-#pragma warning restore CS0618 // #655 - This Rxpk based implementation will go away as soon as the complete LNS implementation is done
-            }
+            var euRegion = RegionManager.EU868;
+
```
atifaziz, line 761:
```suggestion
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
```diff
@@ -85,29 +77,27 @@ public LoRaDeviceTelemetry()
...
-            Tmms = rxpk.Tmms;
-            Tmst = rxpk.Tmst;
             Fcnt = upstreamPayload.GetFcnt();
             Port = upstreamPayload.Fport.Value;
+            Freq = radioMetadata.Frequency.InMega;
+            Datr = datr.ToString();
+            Rssi = radioMetadata.UpInfo.ReceivedSignalStrengthIndication;
+            Rfch = radioMetadata.UpInfo.AntennaPreference;
+            Lsnr = radioMetadata.UpInfo.SignalNoiseRatio;
+            Time = radioMetadata.UpInfo.Xtime.ToString(CultureInfo.InvariantCulture); // This field was just used for telemetry in rxpk. Now it's being used for bringing unaltered the Xtime.
```
atifaziz, line 96: Why convert to a string? Why not keep the original `ulong` type?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
atifaziz: Okay then let's just keep one of them for now in this PR.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
atifaziz: What I meant is validating that it's correct _when it is set_.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
Mandur: We could actually use a time 0 -15 would be possible values here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
Mandur: I will convert this in ulong

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
Mandur: I will add gps time here

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
Mandur: I will revmove the check

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
Mandur: @bastbu

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
Mandur: To be elaborated in #1137

Tests/Integration/CloudToDeviceMessageTests.cs
Mandur: To be verified

Mandur: Add assert lnsrxdelay = 1

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaTools.LoRaPhysical
```
atifaziz, line 4: Add nullability checks for new code:

```suggestion
#nullable enable

namespace LoRaTools.LoRaPhysical
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkBasicsStationMessage.cs
Mandur: As discussed it creates more changes required, therefore we will do that in subsequent PR, tracked in #1138

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceTelemetry.cs
Mandur: Gpstime is already there

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
Mandur: If ok with you I will do it as separate element, I added a comment to specify more the possible values in the meanwhile

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
bastbu: I don't think I have the necessary context to give a good answer here. Without knowing the history of that code, I wonder if we need the null-check/throw at all? If we don't have the region set in a path, surely we will fail somewhere later in the processing chain, and our tests would catch that (since it's only invoked in the join request and uplink data frame cases, which are covered)? In any case, I would probably remove the comment on L78.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs
Mandur: I think you are already looking at it on another branch

atifaziz (approved):

PR #1123, opened 2021-12-20
added new NOTICE file Universal_Decoder_NOTICE 40942

roel4ez:


Mandur:
@roel4ez  It looks like the version are the same, but just swapped out?

roel4ez:
> @roel4ez It looks like the version are the same, but just swapped out?

yep, it seems the NOTICE generator doesn't always use the same order. Will close this PR.

Issue #1003, opened 2021-12-07
Unit test the inner implementation of key creation


spygi:
Context regarding the unit testing: https://github.com/Azure/iotedge-lorawan-starterkit/pull/937/files/7d7d04a1b48d186903bb1e9878715632d2f2dd0c#r763824329

PR #1136, opened 2021-12-21
Feat/1003 unit test cache keys

spygi:
# PR for issue #1003

## What is being addressed

Adds unit tests for the deduplication cache key creation, for data and join requests.

## How is this addressed

- We are not checking what is the exact hash algorithm used as this can change (plus we already have a test checking if an expected hash changes that would detect a change in the algorithm)
- What we are checking is whether the fields we expect are part of the input which will be hashed at the right position. 

Cleanup:
- For the data messages only the payload can be empty, so we could simplify a bit the code (was previously checking if Mic is empty too)
- We could also remove a method that was previously used only from the tests.


Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
```diff
@@ -100,7 +112,32 @@ public void CreateKeyMethod_Should_Produce_Expected_Key_For_UpstreamDataFrames()
...
+            mock.Verify(x => x.HashKey(It.IsAny<byte[]>()), Times.Once);
+            Assert.True(MemoryExtensions.SequenceEqual(this.dataPayload.DevAddr.Span, actualBuffer.Span[0..4]));
+            Assert.True(MemoryExtensions.SequenceEqual(this.dataPayload.Mic.Span, actualBuffer.Span[4..8]));
+            var index = 8;
+            if (this.dataPayload.RawMessage?.Length > 0)
+            {
+                index += this.dataPayload.RawMessage.Length;
+                Assert.True(MemoryExtensions.SequenceEqual(this.dataPayload.RawMessage.AsSpan(), actualBuffer.Span[8..index]));
+            }
+            Assert.True(MemoryExtensions.SequenceEqual(this.dataPayload.Fcnt.Span, actualBuffer.Span[index..])); // implicitly asserts that the length is correct as well
```
atifaziz, line 140: Why not invoke extensions methods more in naturally as extensions?

```suggestion
            Assert.True(this.dataPayload.DevAddr.Span.SequenceEqual(actualBuffer.Span[0..4]));
            Assert.True(this.dataPayload.Mic.Span.SequenceEqual(actualBuffer.Span[4..8]));
            var index = 8;
            if (this.dataPayload.RawMessage?.Length > 0)
            {
                index += this.dataPayload.RawMessage.Length;
                Assert.True(this.dataPayload.RawMessage.AsSpan().SequenceEqual(actualBuffer.Span[8..index]));
            }
            Assert.True(this.dataPayload.Fcnt.Span.SequenceEqual(actualBuffer.Span[index..])); // implicitly asserts that the length is correct as well
```

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
```diff
@@ -156,26 +193,35 @@ public void When_Join_Request_Encountered_Should_Find_Duplicate(string station1,
...
 
-            // act/assert
-            Assert.Throws<ArgumentException>(() => ConcentratorDeduplication.CreateCacheKey(this.dataRequest));
+            // act
+            _ = mock.Object.CreateCacheKey(this.joinPayload);
+
+            // assert
+            mock.Verify(x => x.HashKey(It.IsAny<byte[]>()), Times.Once);
+            Assert.True(MemoryExtensions.SequenceEqual(this.joinPayload.AppEUI.Span, actualBuffer.Span[0..8]));
+            Assert.True(MemoryExtensions.SequenceEqual(this.joinPayload.DevEUI.Span, actualBuffer.Span[8..16]));
```
atifaziz, line 221: Same as [earlier comment](url):

```suggestion
            Assert.True(this.joinPayload.AppEUI.Span.SequenceEqual(actualBuffer.Span[0..8]));
            Assert.True(this.joinPayload.DevEUI.Span.SequenceEqual(actualBuffer.Span[8..16]));
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -12,7 +12,7 @@ namespace LoRaWan.NetworkServer
     using Microsoft.Extensions.Caching.Memory;
     using Microsoft.Extensions.Logging;
 
-    public sealed class ConcentratorDeduplication :
+    public class ConcentratorDeduplication :
```
atifaziz, line 15: Any reason why this is unsealed now?

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
```diff
@@ -20,24 +22,34 @@ public sealed class ConcentratorDeduplicationTest : IDisposable
...
         private readonly WaitableLoRaRequest dataRequest;
         private readonly WaitableLoRaRequest joinRequest;
 
+        public static TheoryData<byte[]?> TestRawMessages =>
+            new TheoryData<byte[]?>
+            {
+                null,
+                Array.Empty<byte>(),
+                new byte[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 }
+            };
```
atifaziz, line 37: nit: Can be made a static field and the constructed type can be omitted for brevity:

```suggestion
        public static readonly TheoryData<byte[]?> TestRawMessages = new()
        {
            null,
            Array.Empty<byte>(),
            new byte[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 }
        };
```

atifaziz: I would recommend moving this field closer to the `CreateKeyMethod_Should_Include_All_Properties_For_Different_RawMessages` test where it's used.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Because of the virtual method `HashKey` which needs to be virtual in order to verify calls on it. Can we avoid leaving the class unsealed?

spygi: I wonder if we can utilize the interface we have and mock on that. But then I think we would have to make `CreateCacheKey` (both implementations) and `HashKey` members of the interface which is also not ideal (why would users of the interface need to implement these methods anyway?)

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
spygi: Thanks (should have realised these are real extension methods not just by name :D) Resolving though because this code was removed.

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
spygi: Thanks, this exact code was removed but we have 2 other instances that do the same where I took your suggestion about making it a static field and omitting the constructed type. Didn't move it closer to the test though because I think it reads better to have all the tests together and last the supporting data if you need to look up/amend them.

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
spygi: See comment above, code was removed.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Since we don't need the virtual method anymore for the tests we can seal the class again. I will resolve this but interested in your answer how we could avoid unsealing :)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -88,15 +88,7 @@ private bool EnsureFirstMessageInCache(string key, LoRaRequest loRaRequest, out
...
-        internal static string CreateCacheKey(LoRaRequest loRaRequest)
-            => loRaRequest.Payload switch
-            {
-                LoRaPayloadData asDataPayload => CreateCacheKey(asDataPayload),
-                LoRaPayloadJoinRequest asJoinPayload => CreateCacheKey(asJoinPayload),
-                _ => throw new ArgumentException($"Provided request is of type {loRaRequest.GetType()} which is not valid for deduplication.")
-            };
-
-        private static string CreateCacheKey(LoRaPayloadData payload)
+        internal string CreateCacheKey(LoRaPayloadData payload)
```
atifaziz, line 91: This can be static.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -105,21 +97,20 @@ private static string CreateCacheKey(LoRaPayloadData payload)
...
             BinaryPrimitives.WriteUInt16LittleEndian(buffer[index..], BinaryPrimitives.ReadUInt16LittleEndian(payload.Fcnt.Span));
 
-            var key = Sha256.ComputeHash(buffer.ToArray());
+            var key = HashKey(buffer.ToArray());
 
             return BitConverter.ToString(key);
         }
 
-        private static string CreateCacheKey(LoRaPayloadJoinRequest payload)
+        internal string CreateCacheKey(LoRaPayloadJoinRequest payload)
```
atifaziz, line 113: This can be static.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -134,9 +125,11 @@ private static string CreateCacheKey(LoRaPayloadJoinRequest payload)
             var devNonce = payload.DevNonce;
             _ = devNonce.Write(buffer);
 
-            var key = Sha256.ComputeHash(head.ToArray());
+            var key = HashKey(head.ToArray()); ;
 
             return BitConverter.ToString(key);
         }
+
+        private static byte[] HashKey(byte[] input) => Sha256.ComputeHash(input);
```
atifaziz, line 133: Is this really needed anymore? Seems to be doing exactly the same as `Sha256.ComputeHash(input)`.

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
```diff
@@ -156,25 +155,56 @@ public void When_Join_Request_Encountered_Should_Find_Duplicate(string station1,
...
+        [Theory, MemberData(nameof(TestJoinRequestData.DataGenerator), MemberType = typeof(TestJoinRequestData))]
+        public void CreateKeyMethod_Should_Return_Expected_Keys_For_Different_JoinRequests(LoRaPayloadJoinRequest joinRequestPayload, string expectedKey)
+        {
+            Assert.Equal(expectedKey, this.concentratorDeduplication.CreateCacheKey(joinRequestPayload));
         }
         #endregion
 
-        [Fact]
-        public void CreateKeyMethod_Should_Throw_When_Used_With_Wrong_Type()
+        private static class TestPayloadData
```
atifaziz, line 174: This seems like it can be simplified to very simple tests with inline data:

```c#
[Theory]
[InlineData("60daa3a5f7dbfa200f8c82840ecf5b42640b70f3b7218a4c6bbd67db542e75a4", 0, 0, 0)]
[InlineData("bf6781db771aef1c14559e2c22e7d1cf4f577777656e2cd6e3d41a6ea16a1786", 0x101010101010101UL, 0, 0)]
[InlineData("4025818beac7accd324e2a02ce15c89b72a481329d919ffe7a62d8ba3ac4e400", 0, 0x101010101010101UL, 0)]
[InlineData("b0ef53e122b3c40b579355219695430329f46c3b24936cbd734967780a609be2", 0, 0, 1)]
public void CreateCacheKey_JoinRequests(string expectedKey, ulong joinEui, ulong devEui, ushort devNonce)
{
    var payload = new LoRaPayloadJoinRequestLns(new MacHeader(MacMessageType.JoinRequest),
                                                new JoinEui(joinEui), new DevEui(devEui),
                                                new DevNonce(devNonce), new Mic(0));
#pragma warning disable CA1308 // Normalize strings to uppercase
    var key = this.concentratorDeduplication.CreateCacheKey(payload)
                                            .ToLowerInvariant()
                                            .Replace("-", string.Empty, StringComparison.Ordinal);
#pragma warning restore CA1308 // Normalize strings to uppercase
    Assert.Equal(expectedKey, key);
}
```

The benefits are:

- No need for separate data classes with byte arrays.
- Test data lives closer to or with the actual test method (inline).
- SHA hashes are shorter to read (though ideally `CreateCacheKey` methods can be modified to simply use `Hexadecimal` class to produce the hashes like that in the first place).
- Good to have expectations comes first, like in assertions.

In diff-speak, the changes would look as follows:

```diff
diff --git a/Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs b/Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
index 9cbfbccd..00dd504f 100644
--- a/Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
+++ b/Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
@@ -164,10 +164,22 @@ namespace LoRaWan.Tests.Unit.NetworkServer
             Assert.Equal(expectedKey, this.concentratorDeduplication.CreateCacheKey(this.joinPayload));
         }
 
-        [Theory, MemberData(nameof(TestJoinRequestData.DataGenerator), MemberType = typeof(TestJoinRequestData))]
-        public void CreateKeyMethod_Should_Return_Expected_Keys_For_Different_JoinRequests(LoRaPayloadJoinRequest joinRequestPayload, string expectedKey)
+        [Theory]
+        [InlineData("60daa3a5f7dbfa200f8c82840ecf5b42640b70f3b7218a4c6bbd67db542e75a4", 0, 0, 0)]
+        [InlineData("bf6781db771aef1c14559e2c22e7d1cf4f577777656e2cd6e3d41a6ea16a1786", 0x101010101010101UL, 0, 0)]
+        [InlineData("4025818beac7accd324e2a02ce15c89b72a481329d919ffe7a62d8ba3ac4e400", 0, 0x101010101010101UL, 0)]
+        [InlineData("b0ef53e122b3c40b579355219695430329f46c3b24936cbd734967780a609be2", 0, 0, 1)]
+        public void CreateCacheKey_JoinRequests(string expectedKey, ulong joinEui, ulong devEui, ushort devNonce)
         {
-            Assert.Equal(expectedKey, this.concentratorDeduplication.CreateCacheKey(joinRequestPayload));
+            var payload = new LoRaPayloadJoinRequestLns(new MacHeader(MacMessageType.JoinRequest),
+                                                        new JoinEui(joinEui), new DevEui(devEui),
+                                                        new DevNonce(devNonce), new Mic(0));
+#pragma warning disable CA1308 // Normalize strings to uppercase
+            var key = this.concentratorDeduplication.CreateCacheKey(payload)
+                                                    .ToLowerInvariant()
+                                                    .Replace("-", string.Empty, StringComparison.Ordinal);
+#pragma warning restore CA1308 // Normalize strings to uppercase
+            Assert.Equal(expectedKey, key);
         }
         #endregion
 
@@ -191,22 +203,6 @@ namespace LoRaWan.Tests.Unit.NetworkServer
             };
         }
 
-        private static class TestJoinRequestData
-        {
-            private static readonly Memory<byte> BaseAppEui = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0 };
-            private static readonly Memory<byte> BaseDevEui = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0 };
-            private static readonly DevNonce BaseDevNonce = new DevNonce(0);
-
-            public static readonly TheoryData<LoRaPayloadJoinRequest, string> DataGenerator = new()
-            {
-                { new LoRaPayloadJoinRequest() { AppEUI = BaseAppEui, DevEUI = BaseDevEui, DevNonce = BaseDevNonce }, "60-DA-A3-A5-F7-DB-FA-20-0F-8C-82-84-0E-CF-5B-42-64-0B-70-F3-B7-21-8A-4C-6B-BD-67-DB-54-2E-75-A4" },
-                { new LoRaPayloadJoinRequest() { AppEUI = BaseAppEui, DevEUI = BaseDevEui, DevNonce = BaseDevNonce, Mhdr = new byte[] { 1 } }, "60-DA-A3-A5-F7-DB-FA-20-0F-8C-82-84-0E-CF-5B-42-64-0B-70-F3-B7-21-8A-4C-6B-BD-67-DB-54-2E-75-A4" },
-                { new LoRaPayloadJoinRequest() { AppEUI = new byte[] { 1, 1, 1, 1, 1, 1, 1, 1 }, DevEUI = BaseDevEui, DevNonce = BaseDevNonce }, "BF-67-81-DB-77-1A-EF-1C-14-55-9E-2C-22-E7-D1-CF-4F-57-77-77-65-6E-2C-D6-E3-D4-1A-6E-A1-6A-17-86" },
-                { new LoRaPayloadJoinRequest() { AppEUI = BaseAppEui, DevEUI = new byte[] { 1, 1, 1, 1, 1, 1, 1, 1 }, DevNonce = BaseDevNonce }, "40-25-81-8B-EA-C7-AC-CD-32-4E-2A-02-CE-15-C8-9B-72-A4-81-32-9D-91-9F-FE-7A-62-D8-BA-3A-C4-E4-00" },
-                { new LoRaPayloadJoinRequest() { AppEUI = BaseAppEui, DevEUI = BaseDevEui, DevNonce = new DevNonce(1) }, "B0-EF-53-E1-22-B3-C4-0B-57-93-55-21-96-95-43-03-29-F4-6C-3B-24-93-6C-BD-73-49-67-78-0A-60-9B-E2" },
-            };
-        }
-
         public void Dispose()
         {
             this.loRaDevice.Dispose();
```

I would recommend doing the same for the other test.

atifaziz (approved):
LGTM on the condition that [the test & data are simplified as recommended](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1136/files#r773843836) (especially because we should be using primitives and newer types to avoid regressing to byte arrays).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Changed e987b029bcbb7d57399eac92758876fe11d6a2f1

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Changed, e987b029bcbb7d57399eac92758876fe11d6a2f1

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
spygi: Removed, fd4c1e27b19e609e5f5a0aeceac20683473b4cb5

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
spygi: Took the majority of your recommendations:
- Inline data (closer to test) without the need for a separate type (no extra mental overhead).
- Using new primitives instead of byte arrays (safer).
- Expectations in the beginning: debatable (I think most of our tests do not conform to this) but ok.

I did not take the recommendation about SHA as we would assert on something different than what the actual code uses (and also it's debatable whether it's more readable to have a continuous long string imo)

spygi: Resolving due to inaction.

Issue #812, opened 2021-11-16
EFLOW VM can be turned on by the CI


Mandur:
The EFLOW VM turns off at 2am automatically.
Ideally the e2e tests should turn on the VM.

We could reduce costs with this.

Nice-to-have.

PR #1140, opened 2021-12-21
Feature/812 turn eflow on off in ci

roel4ez:
# PR for issue #812 

## What is being addressed

This PR adds the ability for the e2e workflow to turn on and off the EFLOW VM.
In this case, `turn off` means `deallocate` so it does not generate any cost.

## How is this addressed

It adds:
- composite action for reusing the login and power on/off logic
- a workflow that can be triggered manually to turn on/off VM
- two jobs to the e2e workflow which turns on the VM at the start, and turns it off at the end. The power-on job is parallelized with the env-var setup.

It requires 3 repository secrets to be able to use `az login`
- AZURE_SP_CLIENTID
- AZURE_SP_SECRET
- AZURE_TENANTID


.github/workflows/e2e-ci.yaml
```diff
@@ -86,6 +86,23 @@ jobs:
...
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Power OFF Azure VM
+        uses: ./.github/actions/power-azure-vm
+        with:
+          AZURE_VM_NAME: eflow
+          AZURE_RG_NAME: lorae2etest
```
roel4ez, line 101:
```suggestion
          AZURE_RG_NAME: lorae2etests
```

.github/workflows/e2e-ci.yaml
```diff
@@ -86,6 +86,23 @@ jobs:
...
       CertRemotePath: "/var/lorastarterkit/certs"
+  
+  power_on_azure_vm:
+    name: Power ON EFLOW
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Power OFF Azure VM
```
roel4ez, line 97:
```suggestion
      - name: Power ON Azure VM
```

.github/workflows/e2e-ci.yaml
```diff
@@ -501,5 +518,22 @@ jobs:
...
+      - e2e_tests_job
+    steps:
+      - uses: actions/checkout@v2
 
-
+      - name: Power OFF Azure VM
+        uses: ./.github/actions/power-azure-vm
+        with:
+          AZURE_VM_NAME: eflow
+          AZURE_RG_NAME: lorae2etest
```
roel4ez, line 535:
```suggestion
          AZURE_RG_NAME: lorae2etests
```

.github/actions/power-azure-vm/action.yml
```diff
@@ -0,0 +1,43 @@
...
+    
+    - name: "Deallocate VM"
+      shell: bash
+      if: ${{ !inputs.POWER_SWITCH }}
+      run: | 
+        az vm deallocate --name ${{ inputs.AZURE_VM_NAME }} --resource-group ${{ inputs.AZURE_RG_NAME }} --no-wait 
+
+    - name: "Start VM"
+      shell: bash
+      if: ${{ inputs.POWER_SWITCH }}
```
Mandur, line 41: nit on if clauses you don't need the {{

.github/workflows/e2e-ci.yaml
```diff
@@ -501,5 +518,22 @@ jobs:
       with:
         name: e2e-test-results
         path: ${{ env.TESTS_RESULTS_FOLDER }}/E2E/
+  
+  power_off_azure_vm:
+    name: Power OFF EFLOW
```
Mandur, line 523: you confirm this call is idempotent right? If the vm is already deallocated this won't fail?

.github/workflows/power_eflow_vm.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Power Azure VM
+        uses: ./.github/actions/power-azure-vm
+        with:
+          AZURE_VM_NAME: ${{ github.event.inputs.vm_name }}
+          AZURE_RG_NAME: ${{ github.event.inputs.rg_name }}
+          POWER_SWITCH: ${{ github.event.inputs.power }}
+          AZURE_SP_CLIENTID: ${{ secrets.AZURE_SP_CLIENTID }}
+          AZURE_SP_SECRET: ${{ secrets.AZURE_SP_SECRET }}
+          AZURE_TENANTID: ${{ secrets.AZURE_TENANTID }}
```
Mandur, line 35: I think you need to add a carriage return at the end of each file, otherwise Atif will get angry :D

Mandur (approved):

.github/workflows/e2e-ci.yaml
roel4ez: YesYepGIF

Issue #1143, opened 2021-12-21
Add and use primitive for RX delay


atifaziz:
Add a type to capture the strict notion of receive (RX) delay, which is always seconds between 0 and 15 but where 0 and 1 _both_ mean 1 second.

Mandur:
I guess @atifaziz you will also tackle #1139 in this one right?

PR #1144, opened 2021-12-21
Add & use a primitive for RX delay

atifaziz:
# PR for issue #1143

## What is being addressed

The strict definition of RX delays (as per spec) is not captured into a type. For example, the spec says the value must be between 0 and 15 and where 0 and 1 both mean a delay of 1 second (all other values represent the delay in seconds).

## How is this addressed

- A new primitive (an `enum`) called `RxDelay` has been added.
- All code has been adapted to use the primitive including helper extensions that help map the delay value to integral seconds and `TimeSpan` instances where needed.


danigian (approved):
LGTM, please run fullci (when possible) to validate changes before merging

Issue #1149, opened 2021-12-23
`DevAddr` is lost from responses provided by functions in `LoraKeysManagerFacade`


atifaziz:
In PR #1073, `IoTHubDeviceInfo.DevAddr` was marked with `[JsonIgnore]`. This means that the property is never part of a JSON-serialized response from a function in `LoraKeysManagerFacade` that returns `IoTHubDeviceInfo`.

p-schuler:
@atifaziz @bastbu - what endpoint are we specifically talking here and what calls?

bastbu:
This affects every endpoint that uses the `LoraKeysManagerFacade.IoTHubDeviceInfo` for as the type for its result (which is the `GetDevice` and `GetDeviceByDevEUI` endpoint of the facade)

PR #1150, opened 2021-12-23
Reuse `DevAddr`

atifaziz:
# PR for issue #1077

## What is being addressed

Representations of LoRaWAN DevAddr using weak types like strings and bytes.

## How is this addressed

Replace with uses of `DevAddr`.

This completes what was started in #962.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
```diff
@@ -56,6 +75,21 @@ public Span<byte> Write(Span<byte> buffer)
             return buffer[Size..];
         }
 
+        /// <summary>
+        /// Creates a <see cref="DevAddr"/> where its <see cref="NetworkId"/> is set to 0,
+        /// representing a private network address.
+        /// </summary>
+        public static DevAddr Private0(int address) => new DevAddr(0, address);
```
bastbu, line 82: Why add static members to production code which are only used from test code?

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
```diff
@@ -108,8 +108,6 @@ public void CreateKeyMethod_Should_Return_Expected_Keys_For_Different_Payloads(s
             var options = fieldNotUsedInKey ?? string.Empty;
             var payload = new LoRaPayloadDataLns(new DevAddr(devAddr), new MacHeader(MacMessageType.ConfirmedDataUp),
                                                  frameCounter, options, rawPayload, new Mic(mic));
-            _ = Hexadecimal.TryParse(rawPayload, out uint raw);
-            payload.RawMessage = BitConverter.GetBytes(raw);
```
bastbu, line 112: I don't understand this change - why do we not need to set the raw message any longer and modify the expected key of only one test?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
```diff
@@ -10,19 +10,15 @@ namespace LoRaTools
...
-            var devAddr = new byte[4];
-
-            RndKeysGenerator.GetBytes(devAddr);
-
-            devAddr[0] = (byte)((nwkPart & 0b11111110) | (devAddr[0] & 0b00000001));
-            return ConversionHelper.ByteArrayToString(devAddr);
+            Span<byte> bytes = stackalloc byte[4];
+            RandomNumberGenerator.Fill(bytes);
+            var id = netId[0] & 0b0111_1111;
+            var address = BitConverter.ToInt32(bytes) & 0x01ff_ffff;
```
bastbu, line 20: Instead of relying on "magical" bit masks here, should we consider making the `NetworkAddressMask` and `NetworkIdMask` `public` or at least internal? It would be easier to figure out what's going on in such scenarios then

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
```diff
@@ -10,19 +10,15 @@ namespace LoRaTools
...
-            var nwkPart = netId[0] << 1;
-            var devAddr = new byte[4];
-
-            RndKeysGenerator.GetBytes(devAddr);
-
-            devAddr[0] = (byte)((nwkPart & 0b11111110) | (devAddr[0] & 0b00000001));
-            return ConversionHelper.ByteArrayToString(devAddr);
+            Span<byte> bytes = stackalloc byte[4];
+            RandomNumberGenerator.Fill(bytes);
+            var id = netId[0] & 0b0111_1111;
```
bastbu, line 19: Don't we have to shift the `netId` by 1 left here?

bastbu (commented):
I think it might make sense to run fullci on this, what do you think?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -90,18 +90,18 @@ private bool EnsureFirstMessageInCache(string key, LoRaRequest loRaRequest, out
...
-            index += payload.DevAddr.Length;
+            _ = payload.DevAddr.Write(buffer);
+            index += DevAddr.Size;
 
             BinaryPrimitives.WriteUInt32LittleEndian(buffer[index..], BinaryPrimitives.ReadUInt32LittleEndian(payload.Mic.Span));
             index += payload.Mic.Length;
 
-            payload.RawMessage?.CopyTo(buffer[index..]);
-            index += payload.RawMessage?.Length ?? 0;
+            payload.Frmpayload.Span.CopyTo(buffer[index..]);
```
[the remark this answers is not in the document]
atifaziz: This a bug being worked on by @spygi in PR #1152. This change was needed to make the tests pass, but I think it's best to disable the test instead and re-enable with #1152.

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
atifaziz: That's being done by `DevAddr` when the network ID gets passed to it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
atifaziz: This is not the same mask. Here, the 7 LSB of the NetID become the 7 MSB of the `DevAddr`. Perhaps the mask can be introduced when we get round to using `NetId` as part of #972.

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
atifaziz: I've reverted this with 2622525262a7d3b993c7b8da8cbe1f884c9736fa as it should be addressed by bug #1152 that was identified as part of this PR.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -312,17 +305,10 @@ private static ushort ValidateAndConvert16bitFCnt(uint fcntDown)
...
-            for (var i = reversedDevAddr.Length - 1; i >= 0; --i)
-            {
-                reversedDevAddr[i] = payloadDevAddr[^(1 + i)];
-            }
-
             var msgType = cloudToDeviceMessage.Confirmed ? MacMessageType.ConfirmedDataDown : MacMessageType.UnconfirmedDataDown;
             var ackLoRaMessage = new LoRaPayloadData(
                 msgType,
-                reversedDevAddr,
+                loRaDevice.DevAddr.Value,
```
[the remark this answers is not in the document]
atifaziz: Because by the time you can send data, the device must have an address assigned.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/DevAddr.cs
atifaziz: This creates a private network address, which is perfectly valid for production too.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
```diff
@@ -111,7 +111,7 @@ private async Task<SearchDevicesResult> SearchDevicesAsync(string gatewayID = nu
             {
                 ["code"] = AuthCode,
                 ["GateWayId"] = gatewayID,
-                ["DevAddr"] = devAddr,
+                ["DevAddr"] = devAddr is { } someDevAddr ? someDevAddr.ToString() : null,
```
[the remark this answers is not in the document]
atifaziz: Yep, silly me. Simplified with 56b8203316d6ac9040d01bdc4ac464e91d133969.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
atifaziz: Reverted with 2622525262a7d3b993c7b8da8cbe1f884c9736fa.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubDeviceInfo.cs
```diff
@@ -3,9 +3,18 @@
 
 namespace LoRaWan.NetworkServer
 {
+    using Newtonsoft.Json;
+
     public class IoTHubDeviceInfo
     {
-        public string DevAddr { get; set; }
+        [JsonProperty("DevAddr")]
+        public string DevAddrString
```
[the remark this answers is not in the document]
atifaziz: In fact, in the _other_ `IoTHubDeviceInfo`, we want to probably use the same strategy as this when addressing #1149.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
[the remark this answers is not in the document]
atifaziz: Test case disabled with ac0ad5cee67d349f49a19cbf4e5273598f376087.

Issue #497, opened 2021-10-12
Required load-tests and refactor/implement as needed in the `Simulation` project


spygi:
Are load tests in scope? If yes, what do we want to test. 

## Acceptance criteria
- ADR regarding load testing
- Refactor/implement accordingly in `Simulation` project (rename as well for extra clarity)

roel4ez:
duplicate #1160

PR #1153, opened 2021-12-23
Remove RXPK & LoRa Physical (and add simulator)

Mandur:
# PR for issue #1127

## What is being addressed

* Remove the RXPK class and LoRaPhysical
* Provide a very barebone simulator implementation working for ABP and OTAA test (single device only, other test will need to be carried in other work item)
* Proposed creation method will enable simulator to dynamically simulate both payloads, but also physical elements, allowing the tests of further regions.

** The current implementation of the simulator is by no mean final, there are still a lot to be done to improve stability, reduce exceptions and change current tests to reduce high flakiness **


Tests/Simulation/IntegrationTestFixtureSim.cs
```diff
@@ -88,37 +88,37 @@ public override void SetupTestDevices()
...
-                    {
-                        DeviceID = deviceID.ToString("0000000000000000", CultureInfo.InvariantCulture),
-                        AppEUI = deviceID.ToString("0000000000000000", CultureInfo.InvariantCulture),
-                        AppKey = deviceID.ToString("00000000000000000000000000000000", CultureInfo.InvariantCulture),
-                        GatewayID = gatewayID,
-                        IsIoTHubDevice = true,
-                        SensorDecoder = "DecoderValueSensor",
-                    });
-            }
+            //// Range of 1000 ABP devices from 2000 to 2999: Used for load testing
```
Mandur, line 91: These tests create 1000 devices and it will be used in #497

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,176 @@
...
+            this.stationEUI = stationEUI;
+            this.LnsUri = lnsUri;
+        }
+
+        public async Task StartAsync()
+        {
+            try
+            {
+                var routerReceivedMessage = new List<string>();
+                using var routerWebsocketClient = new WebsocketClient(new Uri(LnsUri, "router-info"))
```
Mandur, line 47: This code here does the initial handshake by the basicstation, I realized it is unnessary at the moment as we don't formaly check a basicstation went through it to accept its messages. still it seems to do the job.

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,173 @@
...
+    using LoRaTools.LoRaMessage;
+    using LoRaTools.Utils;
+    using System.Net.WebSockets;
+    using Xunit;
+    using System.Linq;
+
+    public sealed class SimulatedBasicsStation : IDisposable
+    {
+        private readonly string stationEUI;
+        private WebsocketClient DataWebsocketClient { get; set; }
```
bastbu, line 25: Why use a library for this and not something that [.NET gives us](https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket?view=net-6.0)?

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,173 @@
...
+                await routerWebsocketClient.Start();
+                using var ms = new MemoryStream();
+                using var writer = new Utf8JsonWriter(ms);
+
+                await routerWebsocketClient.SendInstant(JsonSerializer.Serialize(new
+                {
+                    router = this.stationEUI
+                }));
+
+                await Task.Delay(5000);
```
bastbu, line 69: It would be good if we could figure out a way to wait for the response instead of having the static timeout here.

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        //// Sends unconfirmed message
+        public async Task SendDataMessageAsync(LoRaRequest LoRaRequest)
+        {
+            var payload = (LoRaPayloadData)LoRaRequest.Payload;
+
+            // catch me if you can.
+            payload.DevAddr.Span.Reverse();
+
+            var msg = JsonSerializer.Serialize(new
```
bastbu, line 130: It would be great if we could run the simulator tests from now on (or at least part of them) to ensure that we don't break such logic.

Tests/Common/SimulatedDevice.cs
```diff
@@ -184,128 +185,105 @@ public LoRaPayloadData CreateConfirmedDataUpMessage(string data, uint? fcnt = nu
...
 
         // Performs join
-        // THE FOLLOWING METHODS ARE LEFT HERE FOR REFERENCE
-        // THEY ACT AS DOCUMENTATION UNTIL WE MIGRATE THE SIMULATOR
-#pragma warning disable CA1822 // Mark members as static
-#pragma warning disable IDE0060 // Remove unused parameter
-        public async Task<bool> JoinAsync(SimulatedPacketForwarder packetForwarder, int timeoutInMs = 30 * 1000)
-#pragma warning restore IDE0060 // Remove unused parameter
-#pragma warning restore CA1822 // Mark members as static
+        public async Task<bool> JoinAsync(LoRaRequest joinRequest, SimulatedBasicsStation basicsStation, int timeoutInMs = 30 * 1000)
```
bastbu, line 233: I would advocate to use `TimeSpan? timeout = null` as a parameter and then initialize it as needed, e.g. as `timeout ??= TimeSpan.FromSeconds(30)`. This would make it easier to not need to keep track of the units for the timeouts.

bastbu (approved):
I have a couple of suggestions, none of them I would require since the scope of this PR is more to enable us to use the tests that were previously broken. We can also extract some of the suggestions to a separate story if we decide to implement them.

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        //// Sends unconfirmed message
+        public async Task SendDataMessageAsync(LoRaRequest LoRaRequest)
+        {
+            var payload = (LoRaPayloadData)LoRaRequest.Payload;
+
+            // catch me if you can.
+            payload.DevAddr.Span.Reverse();
```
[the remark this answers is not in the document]
Mandur: What type do you mean here?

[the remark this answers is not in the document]
bastbu: `DevAddr`

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,173 @@
...
+                    rctx = 10,
+                    rssi = LoRaRequest.RadioMetadata.UpInfo.ReceivedSignalStrengthIndication,
+                    xtime = LoRaRequest.RadioMetadata.UpInfo.Xtime,
+                    snr = LoRaRequest.RadioMetadata.UpInfo.SignalNoiseRatio
+                }
+            });
+
+            await SendMessageAsync(msg);
+
+            // TestLogger.Log($"[{LoRaDevice.DevAddr}] Sending data: {BitConverter.ToString(header).Replace("-", "")}{Encoding.UTF8.GetString(gatewayInfo)}");
```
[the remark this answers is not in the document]
Mandur: Uncommented

Tests/Common/SimulatedBasicsStation.cs
Mandur: Created #1221 to track the issue, this is a good idea but would require too much changes in this PR

Tests/Common/SimulatedBasicsStation.cs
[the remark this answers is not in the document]
Mandur: I understood ;)

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        internal async Task SendMessageAsync(string vs)
+        {
+            await DataWebsocketClient.SendInstant(vs);
+        }
+
+        public async Task StopAsync()
+        {
+            await DataWebsocketClient.Stop(System.Net.WebSockets.WebSocketCloseStatus.NormalClosure, "closing WS");
```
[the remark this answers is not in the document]
Mandur: Let's discuss this

Tests/Common/SimulatedBasicsStation.cs
Mandur: Agree - I added a work item for this. I think the tests are not yet in shape to bring enough values . #1223

Tests/Simulation/SimulatorTestCollection.cs
```diff
@@ -32,60 +31,59 @@ public SimulatorTestCollection(IntegrationTestFixtureSim testFixture)
...
             {
-                await simulatedDevice.SendUnconfirmedMessageAsync(simulatedPacketForwarder, i.ToString(CultureInfo.InvariantCulture));
-                // await simulatedDevice.SendConfirmedMessageAsync(simulatedPacketForwarder, i.ToString());
+                using var request = WaitableLoRaRequest.CreateWaitableRequest(simulatedDevice.CreateConfirmedDataUpMessage(i.ToString(CultureInfo.InvariantCulture)));
+                await simulatedBasicsStation.SendDataMessageAsync(request);
                 await Task.Delay(this.intervalBetweenMessages);
             }
 
-            await simulatedPacketForwarder.StopAsync();
+            await simulatedBasicsStation.StopAsync();
```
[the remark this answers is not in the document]
Mandur: What kind of assertion do you have in mind?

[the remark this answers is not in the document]
bastbu: Maybe that the LNS delivered the message to IoT Hub?

Tests/E2E/SensorDecodingTest.cs
```diff
@@ -4,6 +4,7 @@
 namespace LoRaWan.Tests.E2E
 {
     using System.Threading.Tasks;
+    using LoRaWan.Tests.Common;
```
bastbu, line 7: [nit] remove

Tests/Simulation/SimulatorTestCollection.cs
```diff
@@ -21,6 +20,7 @@ public sealed class SimulatorTestCollection : IntegrationTestBaseSim
     {
         private readonly TimeSpan intervalBetweenMessages;
         private readonly TimeSpan intervalAfterJoin;
+        private const string StationEUI = "B8-27-EB-FF-FE-A3-BE-42";
```
bastbu, line 23: [nit] why not save it as type `StationEui` already?

Tests/Simulation/SimulatorTestCollection.cs
```diff
@@ -32,60 +32,62 @@ public SimulatorTestCollection(IntegrationTestFixtureSim testFixture)
...
 
             for (var i = 1; i <= MessageCount; i++)
             {
-                await simulatedDevice.SendUnconfirmedMessageAsync(simulatedPacketForwarder, i.ToString(CultureInfo.InvariantCulture));
-                // await simulatedDevice.SendConfirmedMessageAsync(simulatedPacketForwarder, i.ToString());
+                using var request = WaitableLoRaRequest.CreateWaitableRequest(simulatedDevice.CreateConfirmedDataUpMessage(i.ToString(CultureInfo.InvariantCulture)));
+                await simulatedBasicsStation.SendDataMessageAsync(request);
                 await Task.Delay(this.intervalBetweenMessages);
             }
+            var msgsFromDevice = TestFixture.IoTHubMessages.Events.Where(x => x.GetDeviceId() == simulatedDevice.LoRaDevice.DeviceID);
```
bastbu, line 86: [nit] instead of filtering and then counting, we can count directly

```csharp
var actualAmountOfMsgs = TestFixture.IoTHubMessages.Events.Count(x => x.GetDeviceId() == simulatedDevice.LoRaDevice.DeviceID && x.Properties.ContainsKey("iothub-message-schema"));
```

Tests/Simulation/SimulatorTestCollection.cs
```diff
@@ -94,25 +96,27 @@ public async Task Single_OTAA_Simulated_Device()
             const int MessageCount = 5;
 
             var device = TestFixtureSim.Device1002_Simulated_OTAA;
-            var simulatedDevice = new SimulatedDevice(device);
-            var networkServerIPEndpoint = CreateNetworkServerEndpoint();
+            var simulatedDevice = new Common.SimulatedDevice(device);
```
bastbu, line 99:
```suggestion
            var simulatedDevice = new SimulatedDevice(device);
```

Tests/Common/SimulatedBasicsStation.cs
Mandur: SHould be done now

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            // we want to ensure we received a uri message
+            await AssertUtils.ContainsWithRetriesAsync(x => x.Contains("uri", StringComparison.OrdinalIgnoreCase), routerReceivedMessage, interval: TimeSpan.FromSeconds(5d));
+
+            await routerWebsocketClient.Stop(WebSocketCloseStatus.NormalClosure, "closing WS");
+            var factory = new Func<ClientWebSocket>(() => new ClientWebSocket
+            {
+                Options =
+                        {
+                            KeepAliveInterval = TimeSpan.FromSeconds(5),
```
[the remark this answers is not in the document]
Mandur: Good point, I will have a look

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
```diff
@@ -135,7 +135,7 @@ private LoRaPayloadJoinAccept(byte[] inputMessage, AppKey appKey)
             Array.Copy(decryptedPayload, 0, inputMessage, 1, decryptedPayload.Length);
             // ( MACPayload = AppNonce[3] | NetID[3] | DevAddr[4] | DLSettings[1] | RxDelay[1] | CFList[0|15] )
             AppNonce = AppNonce.Read(inputMessage.AsSpan(1));
-            NetId = NetId.Read(inputMessage.AsSpan(3));
+            NetId = NetId.Read(inputMessage.AsSpan(4));
```
Mandur, line 138: @bastbu

Tests/Common/SimulatedBasicsStation.cs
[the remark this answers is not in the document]
Mandur: Removed it, makes it much better.

Tests/Common/SimulatedBasicsStation.cs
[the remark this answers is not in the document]
Mandur: tracked in #1254

Tests/Common/LnsVersionRequest.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Simulation.Models
+{
+    using System.Text.Json.Serialization;
+
+    public sealed record LnsVersionRequest
+    {
+        [JsonPropertyName("msgtype")]
+        public string Msgtype { get; set; } = "version";
```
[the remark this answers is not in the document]
Mandur: Tracked in #1253

Tests/Common/JoinAcceptResponse.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Common
+{
+    using System.Text.Json.Serialization;
+
+    internal class JoinAcceptResponse
+    {
+        [JsonPropertyName("pdu")]
+        public string Pdu { get; set; }
```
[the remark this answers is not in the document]
Mandur: Tracked in #1253

Tests/Simulation/SimulatorTestCollection.cs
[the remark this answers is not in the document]
Mandur: Let's discuss this

PR #1157, opened 2021-12-24
Remove byte array for class

Mandur:
# PR for issue #1138

## What is being addressed

Replace byte array by ReadonlyMemory to avoid code style warnings


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
```diff
@@ -65,7 +65,7 @@ private string Message(DownlinkMessage message)
             writer.WriteNumber("dC", (int)deviceClassType);
 
             // Getting and writing payload bytes
-            var pduBytes = message.Data;
+            var pduBytes = message.Data.ToArray();
```
p-schuler, line 68: I think you can simply use the Data directly?

p-schuler: you would have to add an overload to Hexadecimal to deal with the memory or use the Span (exposed from ReadOnlyMemory<>).

In Hexadecimal for example:

`public static void Write(ReadOnlyMemory<byte> buffer, Span<char> output) => Write(buffer.Span, output, null);`

Mandur: Good point!

atifaziz: I think what @p-schuler meant (as a better alternative) is that you can just use `Span` on the memory and use the existing methods of `Hexadecimal` instead of needing to add an overload for `Write`:

```diff
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
index 6ce70588..d5b409c1 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
@@ -67,7 +67,7 @@ namespace LoRaWan.NetworkServer.BasicsStation
             // Getting and writing payload bytes
             var pduBytes = message.Data;
             var pduChars = new char[pduBytes.Length * 2];
-            Hexadecimal.Write(pduBytes, pduChars);
+            Hexadecimal.Write(pduBytes.Span, pduChars);
             writer.WriteString("pdu", pduChars);
 
 #pragma warning disable CA5394 // Do not use insecure randomness. This is fine as not used for any crypto operations.
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Hexadecimal.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Hexadecimal.cs
index 0498e4a1..77b6abb8 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Hexadecimal.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Hexadecimal.cs
@@ -19,8 +19,6 @@ namespace LoRaWan
                 throw new ArgumentException(InsufficientBufferSizeErrorMessage, paramName);
         }
 
-        public static void Write(ReadOnlyMemory<byte> buffer, Span<char> output) => Write(buffer.Span, output, null);
-
         public static Span<char> Write(byte value, Span<char> output, LetterCase letterCase = LetterCase.Upper)
         {
             ValidateSufficientlySizedBuffer(output.Length, sizeof(byte), nameof(output));
```

I wouldn't favour adding an overload because we'd have to technically maintain more code and unit tests for those overloads (missing in this PR).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -102,6 +102,9 @@ public LoRaPayloadData()
         {
         }
 
+        public LoRaPayloadData(ReadOnlyMemory<byte> inputMessage) : this(inputMessage.ToArray())
```
atifaziz, line 105: The [constructor that this delegates to](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1157/files#diff-434a94aed910f6c38befd5bd0e31ceac7a1ea711b9ac654b3a05b0b71d1fc964L110) should be made `private` to avoid two different ways of constructing a class instance where one constructor makes a copy of the data and the other doesn't. Having two different policies could lead to subtle bugs since a `ReadOnlyMemory<>` doesn't by itself guarantee that the underlying array cannot change. It's up to the class implementation to take a copy of the input and expose that in an immutable manner.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
```diff
@@ -98,6 +98,9 @@ public LoRaPayloadJoinAccept(string netId, byte[] devAddr, byte[] appNonce, byte
             }
         }
 
+        public LoRaPayloadJoinAccept(ReadOnlyMemory<byte> inputMessage, AppKey appKey) : this(inputMessage.ToArray(), appKey)
+        { }
+
         public LoRaPayloadJoinAccept(byte[] inputMessage, AppKey appKey)
```
atifaziz, line 104:
```suggestion
        private LoRaPayloadJoinAccept(byte[] inputMessage, AppKey appKey)
```

This constructor should be made `private` to avoid two different ways of constructing a class instance where one constructor makes a copy of the data and the other doesn't. Having two different policies could lead to subtle bugs since a `ReadOnlyMemory<>` doesn't by itself guarantee that the underlying array cannot change. It's up to the class implementation to take a copy of the input and expose that in an immutable manner.

atifaziz (approved):
If we can address the two comments about making the constructor overloads private then this LGTM.

Issue #1152, opened 2021-12-23
Empty field is used on the cache keys for deduplication causing more messages to potentially be detected as duplicates


spygi:
## Current Behavior on Dev

Should fix ❌

When creating the cache keys, we are making use of the `RawPayload` field. The problem is that this is set only for the (old) `LoRaPayloadData` not the newer one `LoRaPayloadDataLns` which is actually used from the code. This means that we don't take into consideration the payload of the messages leading to more cache hits than what it should (false positives).

Since the default configuration is Deduplication None it means that we skip sending confirmation for more messages than what we should. The severity is worse with Deduplication drop because then we drop the messages immediately.

## Current behavior on latest release

Not affected ✔

According to Mik the last release happened on December 1. Checking what the code looked like on the next day a97a078d55d5e89eac59279c66731e21232761ba

- We are using `UpstreamDataFrameReader.Payload` which is populated (correctly) from the `Frmpayload`
- We are checking if the concentrator used is the same as on the previous message and if yes (which will be the case in the current deployment since they have only 1 concentrator : 1 gateway) we are *not* dropping messages (considered resubmissions)
  
## Current behavior on latest Github release

Not affected ✔

[According to the release list](https://github.com/Azure/iotedge-lorawan-starterkit/releases) the last one happened on November 15. The deduplication service was integrated after that in November 16 with this commit 853c2668e22cd07d02d67f9a90cacb0f3712a3c9.

p-schuler:
@spygi - let us know if we need to triage this.

spygi:
To sum up our discussion with @p-schuler :
- the reason why we had the payload there is to avoid collisions from the DevAddr (not unique)
- if we switch to using the DevEui we could completely drop the payload from the key.

PR #1158, opened 2021-12-24
Bug/1152 use correct payload

spygi:
# PR for issue #1152 

## What is being addressed

On creating the key for the deduplication cache we were utilizing the RawPayload field which is present only in the previous LoRaPayloadData primitive, not the newer that is actually used from the code. This means that in production we are getting more cache hits than should be (false positives). 

## How is this addressed

By using the newer primitive `LoRaPayloadDataLns` and the correct field in it: `Frmpayload`. Unit and integration tests related to deduplication are passing and so is the [E2E test for multiconcentrator in the CI.](https://github.com/Azure/iotedge-lorawan-starterkit/runs/4626900165?check_suite_focus=true)

## TODO
- [ ] Fix the rest of the integration tests that fail (>200). Most (all?) rely on `SimulatedDevice` methods which create a `LoRaPayloadData` payload: we get an exception when we cast it to a `LoRaPayloadDataLns` in L.53 of the `ConcentratorDeduplication`. Trying to switch the `SimulatedDevice` methods to return a `LoRaPayloadDataLns` was not trivial (results in the device not found in the device registry - not sure why).


spygi:
[We decided](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1152#issuecomment-1008702693) to remove the payload all-together from the computation of the cache key and rely on the unique DevEui instead. Hence the issue will be solved in a different PR and this one will be closed.

Issue #821, opened 2021-11-16
A new AzFunction endpoint should be added to send a class C message.


roel4ez:
If we[ currently want](https://azure.github.io/iotedge-lorawan-starterkit/dev/quickstart/#class-a-devices) to send a message to a class c device, we need to send a cloud to device message from IoT Hub targeting the right LNS, in addition with the recent changes we will need also to provide the correct concentrator.

expected parameters are the one of the LoRaCloudToDeviceMessage class. Fport, (payload || rawpayload) and DEVeui are mandatory.

The Idea here is to have an Azure function endpoint handling the class c message handling, the function should be able to:
 - [ ] PreferredGatewayID reported property need to be set to the current gateway id (=gateway ID) (in case there are no desired property GatewayID set) in case of upstream device message at the [same time as we set the stationEUI on upstream ](https://github.com/Azure/iotedge-lorawan-starterkit/blob/3043ae1c900390b84f0689fceb4a3478d56e537a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs#L430)
 - [ ] Connect to IoT Hub to see what the device's assigned IoT Edge is and send the method to it (send it either to the desired property GatewayID or reported prop PreferredGatewayID)
 - [ ] Resolve to which concentrator to send the message (could be done on the LNS, the information is available [in the stationeui reported prop](https://github.com/Azure/iotedge-lorawan-starterkit/blob/ca99c726a6cf97d54118b825859067ec799cbe3f/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs#L173))
 - [ ] New code is unit/integration tested

MaggieSalak:
Closing this story as it appears that the requested functionality is already implemented:
- `SendCloudToDeviceMessage` endpoint [already exists](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs) in the Facade project, which also supports class C devices and uses direct method to send the message to the correct gateway (retrieved from device twin). The endpoint [was added](https://github.com/Azure/iotedge-lorawan-starterkit/pull/147) back in March 2019. 
[...]

PR #147, opened 2019-03-11
Preferred Gateway and Function to send c2d message

fbeltrao:
Added features:
- Azure Function Service Facade has endpoint to send cloud to device messages, removing the complexity existing with class C devices

Changes
- Functions bundling is being called even for 1 execution item. Removed Adr and Dedup endpoints. Pumped Azure function to 2019-03-08-preview, not backward compatible
- Mac commands in c2d is tested with devStatus
- Updated documentation on cloud to device messaging
- Class C devices without a fix gateway assigned call Service Facade when an upstream message is received. Facade decides the closest gateway based on received signal strength indicator (RSSI)
- Last known region for class C devices are stored in twin. Handles the scenario where network server first task is to send a class C device message (w/o receiving a single rxpk to resolve region)
- LoRa device changes can be triggered by change tracking enabled properties
- Moved LoRaCloudToDeviceMessage to Tools project (used by tests, function service facade and network server)
- Added NotSet value to LoRaRegion enum, indicating a not associated variable. 
- Split c2d message size tests to reduce test duration

Fixes AB#1082

PR #1163, opened 2021-12-31
Preferred gateway ID is set for class C devices

MaggieSalak:
# PR for issue #821 

## What is being addressed

`PreferredGatewayID` reported twin property is updated for class C devices.

## How is this addressed

- Preferred gateway ID reported property is updated on upstream message, in case device is a class C device (this is required when retrieving the gateway ID in order to send messages to a class C device via direct method)
- Affected integration tests were updated 

Originally user story #821 was about adding a new AzFunction endpoint for sending messages to class C devices. While working on the story I discovered that the `LoRaKeysManagerFacade` project already implements a `SendCloudToDeviceMessage` endpoint (added in [#147](https://github.com/Azure/iotedge-lorawan-starterkit/pull/147)) which appears to already have all required features, including retrieving of the Gateway ID which is then used to send message via direct method if target device is a class C device.
The only missing functionality was updating of the preferred gateway ID in case it is not set on the LoRa device.


MaggieSalak:
Abandoning this PR as the preferred gateway ID is already set for class C devices in `DefaultLoRaDataRequestHandler` after the Function Bundler result is obtained.

Issue #1129, opened 2021-12-20
Rssi should be a double


Mandur:
RSSI is coming from the basicstation as a double, we are casting it to an int in FunctionBundlerprovider, that should be corrected

PR #1164, opened 2022-01-03
Ensure that RSSI is not casted to an int

bastbu:
# PR for issue #1129 

## What is being addressed

Ensures that RSSI is not casted to an int.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
@@ -47,7 +47,7 @@ namespace LoRaTools.LoRaPhysical
         public string Codr { get; set; }
 
         [JsonProperty("rssi")]
-        public double Rssi { get; set; }
+        public int Rssi { get; set; }
 
         [JsonProperty("lsnr")]
         public float Lsnr { get; set; }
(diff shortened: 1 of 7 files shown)


danigian:
What about LoraKeysManagerFacade\FunctionsBundler\FunctionBundlerRequest.cs and related logic for ComputePreferredGateway?

danigian (changes requested):
What about LoraKeysManagerFacade\FunctionsBundler\FunctionBundlerRequest.cs and related logic for ComputePreferredGateway?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
```diff
@@ -47,7 +47,7 @@ public class Rxpk
         public string Codr { get; set; }
 
         [JsonProperty("rssi")]
-        public int Rssi { get; set; }
+        public double Rssi { get; set; }
```
danigian, line 50: Please revert this change as it's not needed and will be removed anyways with #1153

danigian (approved):
Looks good to me now. Please run a full ci before merging

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaPhysical/Rxpk.cs
bastbu: Done in aab1bb53029445d5dfe49891416cf0b458224419

Issue #1038, opened 2021-12-10
[Story] Set role name in telemetry client


bastbu:
## Description

We should initialize the Application Insights telemetry client with the role name (and other metadata that is useful)

Suggestion: 
RoleName: NetworkServer / Facade
RoleInstance: GatewayId / -

PR #1165, opened 2022-01-03
Set RoleName and RoleInstance on each LNS telemetry item

bastbu:
# PR for issue #1038 

## What is being addressed

Sets the `cloud_RoleName` and `cloud_RoleInstance` on each LNS telemetry item that is sent to Application Insights


danigian (approved):
What about adding a Unit Test for the TelemetryInitializer? 
I would test that RoleName is actually "NetworkServer" (as a sanity check forcing anyone changing it to really understand what's doing) and test that the gateway id is being set as RoleInstance

PR #1171, opened 2022-01-04
added new NOTICE file LNS_NOTICE 41335

roel4ez:


roel4ez:
no changes, only order.

PR #1176, opened 2022-01-04
Revert "Ensure that RSSI is not casted to an int"

bastbu:
Reverts Azure/iotedge-lorawan-starterkit#1164


danigian (approved):
Approving the revert because of failing test

PR #1177, opened 2022-01-04
Reuse `AppSessionKey`, `AppKey` and `NetworkSessionKey`

bastbu:
# PR for issue #968

## What is being addressed

With this PR we use the `AppSessionKey`, `AppKey` and `NetworkSessionKey` types wherever possible.

## How is this addressed

We use the types wherever possible. In addition to introducing the types, we fix a bug in the reading of `Data128` from a buffer.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -212,7 +212,7 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
                     {
                         try
                         {
-                            decryptedPayloadData = loraPayload.GetDecryptedPayload(loRaDevice.AppSKey);
+                            decryptedPayloadData = loraPayload.GetDecryptedPayload(loRaDevice.AppSKey.Value);
```
p-schuler, line 215: The code scanning is right. The AppSKey, might be null?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -217,7 +217,7 @@ internal static class DownlinkMessageBuilder
 
         private static DownlinkMessage BuildDownstreamMessage(LoRaDevice loRaDevice, StationEui stationEUI, ILogger logger, ulong xTime, DataRateIndex rx1Datr, DataRateIndex rx2Datr, Hertz freqRx1, Hertz freqRx2, ushort lnsRxDelay, LoRaPayloadData loRaMessage, uint? antennaPreference = null)
         {
-            var messageBytes = loRaMessage.Serialize(loRaDevice.AppSKey, loRaDevice.NwkSKey);
+            var messageBytes = loRaMessage.Serialize(loRaDevice.AppSKey.Value, loRaDevice.NwkSKey.Value);
```
p-schuler, line 220: same here: I don't think we can assume the AppSKey and NwkSKey are not null here? I thin in theory they should never be, as we process the message, but the type does indicate it might be null.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubDeviceInfo.cs
```diff
@@ -13,7 +15,15 @@ public class IoTHubDeviceInfo
 
         public string GatewayId { get; set; }
 
-        public string NwkSKey { get; set; }
+        [JsonProperty("NwkSKey")]
+        public string NwkSKeyString
+        {
+            get => NwkSKey?.ToString();
+            set => NwkSKey = value is { Length: 32 } some ? LoRaWan.NetworkSessionKey.Parse(some) : null;
```
p-schuler, line 22: if we set an invalid key ( != 32 length), this will simply ignore it. Should we throw in that case?

p-schuler: I think this would never happen in our case, as the serialization on the function side will make sure it's the right format - hence I think it's safe to throw and avoid any other paths to provide an invalid format.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: Yes, I have that in other places as well so might as well track the whole discussion here. I think that's a direct consequence of not having two different types for `ABP`/`OTAA` devices and incrementally updating properties on these devices after creation.

My thinking was that we have tests to verify what our types do not guarantee - in this case that the value is not `null`. I could add a manual check and throw an exception with a custom message if the value is `null`, but at the end of the day that's more or less the same as the default exception that gets thrown by accessing `.Value` on a `null`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -98,7 +98,7 @@ internal async Task ProcessJoinRequestAsync(LoRaRequest request)
                     return;
                 }
 
-                if (!joinReq.CheckMic(loRaDevice.AppKey))
+                if (!joinReq.CheckMic(loRaDevice.AppKey.Value))
```
p-schuler, line 101: nit: I would cache it. We keep on repeating loRaDevice.AppKey.Value

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -266,21 +266,21 @@ public virtual async Task<bool> InitializeAsync(NetworkServerConfiguration confi
             if (twin.Properties.Desired.Contains(TwinProperty.AppSKey))
             {
                 // ABP Case
-                AppSKey = twin.Properties.Desired[TwinProperty.AppSKey].Value as string;
+                AppSKey = AppSessionKey.TryParse(twin.Properties.Desired[TwinProperty.AppSKey].Value as string, out var someAppSessionKey) ? someAppSessionKey : null;
```
p-schuler, line 269: those will have to be changed, either with mine or your change.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubDeviceInfo.cs
bastbu: Agree - I will change the logic to shortcut if there's an empty string, but always try to parse if the string is non-empty

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Data128.cs
```diff
@@ -39,8 +39,8 @@ public override string ToString()
 
         public static Data128 Read(ReadOnlySpan<byte> buffer)
         {
-            var lo = BinaryPrimitives.ReadUInt64BigEndian(buffer);
-            var hi = BinaryPrimitives.ReadUInt64BigEndian(buffer[8..]);
+            var hi = BinaryPrimitives.ReadUInt64BigEndian(buffer);
```
p-schuler, line 42: can we add unit tests, I don't see coverage for this?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -326,15 +328,33 @@ public void ChangeEndianess()
         /// </summary>
         /// <remarks>
         /// src https://github.com/jieter/python-lora/blob/master/lora/crypto.py.</remarks>
-        public byte[] GetDecryptedPayload(string appSkey)
```
p-schuler, line 329: the name suggests we use the appSKey? How come we change this to use the NetworkSessionKey?

p-schuler: sorry, I see that we have 2 overloads. all good.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Data128.cs
bastbu: There is a unit test for this, added it in b0701987576f4c9f723053cbbbaa7b8dc34cedfb

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
```diff
@@ -393,13 +413,21 @@ public byte[] GetDecryptedPayload(string appSkey)
...
+        public override byte[] PerformEncryption(NetworkSessionKey key) =>
+            PerformEncryption(GetDecryptedPayload(key));
+
+        /// <summary>
+        ///  Replaces the <see cref="Frmpayload"/>, encrypting the values.
+        /// </summary>
+        public override byte[] PerformEncryption(AppSessionKey key) =>
+            PerformEncryption(GetDecryptedPayload(key));
+
+        private byte[] PerformEncryption(byte[] rawDecryptedPayload)
```
p-schuler, line 425: I don't get the name of this method. The method does copy the passed in byte buffer to the RawMessage. That's an operation that won't do encryption, as far as I understand.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
p-schuler: We had a discussion on this. We agree the proper fix would be to have dedicated types, but that would be a very large change. Hence, we rely on the tests to guarantee the key won't be null here.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
p-schuler: same here: we can't process a message when we don't have the session key. Should always be valid.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Data128.cs
p-schuler: that's not unit testing Data128, is it? I would assume to have unit tests for the implementation of that type?

bastbu: As of now we don't have any unit tests of `Data128` at all, it's all implicitly covered through the tests of the different key types. I will add that test for `Data128` there, but I won't write the tests for the existing functionality if that's OK

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
bastbu: Done in a77060bca2caa825e0a345fdb02250e3ddad242e

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IoTHubDeviceInfo.cs
bastbu: Done in c5f2b47adb25c0d58b985d6cc6a983e3b3450e23

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Data128.cs
bastbu: Done in 5d3a17d6dbf5c7b706164cd525c16f21aed01c50

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadData.cs
p-schuler: Ok - is there an overload that does? Otherwise I would suggest to change the name. You can resolve when you checked this.

bastbu: Renamed it to the same as its calling method in 7855f2d8eb479e2239f24ffc4c35bac85e380440. Please let me know if that's better.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -292,7 +292,7 @@ public virtual async Task<bool> InitializeAsync(NetworkServerConfiguration confi
                 // OTAA
                 try
                 {
-                    AppKey = desiredTwin.ReadRequiredString(TwinProperty.AppKey);
+                    AppKey = desiredTwin.TryRead(TwinProperty.AppKey, out AppKey someAppKey) ? someAppKey : null;
```
p-schuler, line 295: I think this is wrong. We need to make sure we have an AppKey and throw if there is none for OTAA. Otherwise the device could never join.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -276,8 +276,8 @@ public virtual async Task<bool> InitializeAsync(NetworkServerConfiguration confi
                 // ABP Case
                 try
                 {
-                    AppSKey = desiredTwin.ReadRequiredString(TwinProperty.AppSKey);
-                    NwkSKey = desiredTwin.ReadRequiredString(TwinProperty.NwkSKey);
+                    AppSKey = desiredTwin.TryRead(TwinProperty.AppSKey, out AppSessionKey someAppSessionKey) ? someAppSessionKey : throw new InvalidLoRaDeviceException($"'{TwinProperty.AppSKey}' is empty");
```
p-schuler, line 279: nit: should we add a ReadRequiredProperty instead?

p-schuler (approved):
LGTM

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
bastbu: Done in 6f3c2197acaa8a6f83a14fe9e4b34e7056b33bf4

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
bastbu: Done in 6f3c2197acaa8a6f83a14fe9e4b34e7056b33bf4

PR #1178, opened 2022-01-04
Bug/1129 rssi double

bastbu:
# PR for issue #1129 

## What is being addressed

Ensures that all occurrences of RSSI are typed as `double` instead of a mixture between `double` and `int`.


bastbu:
Flakey OTAATest finally succeeded here: https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1657553687

danigian (approved):

Issue #1142, opened 2021-12-21
End to end tests testing class C should use the new azure function Endpoint


Mandur:
- [ ] Class C test should call the azure function instead of direct method
- [ ] Device24_OTAA should be changed to remove gateway Id assignement in the IntegrationtestfixtureCI

PR #1180, opened 2022-01-04
Class C E2E test uses function endpoint

MaggieSalak:
# PR for issue #1142 

## What is being addressed

Class C E2E test now calls the Facade endpoint `SendCloudToDeviceMessage` instead of calling direct method from the test code.

Successful E2E CI run for the ClassC E2E test: https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1657729247

## How is this addressed

- Device 24 used in Class C E2E test no longer has gateway assigned at the provisioning time - it's now determined by the LNS.
- Functionality to call `SendCloudToDeviceMessage` endpoint has been added to `LoRaApiHelper` in the E2E test project.
- The ClassC E2E test calls the API endpoint `SendCloudToDeviceMessage` (using the `LoRaApiHelper`) instead of calling direct method from the test code.
- `InvokeModuleDirectMethodAsync` helper method in `IntegrationTestFixtureBase` is now unused and has been deleted.
- `SendCloudToDeviceMessage` has been improved by adding API version and DevEUI validation.
- `SendCloudToDeviceMessage` endpoint returns InternalServerError instead of BadRequest in case no gateway has been found for the device in either reported or desired twin properties.
- `SendCloudToDeviceMessage` unit tests have been expanded to cover API version validation, invalid/missing devEUI and unknown gateway ID for the device.

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs
@@ -45,6 +45,15 @@ namespace LoraKeysManagerFacade
             [HttpTrigger(AuthorizationLevel.Function, "post", Route = "cloudtodevicemessage/{devEUI}")] HttpRequest req,
             string devEUI)
         {
+            try
+            {
+                VersionValidator.Validate(req);
+            }
+            catch (IncompatibleVersionException ex)
+            {
+                return new BadRequestObjectResult(ex.Message);
+            }
+
             EUIValidator.ValidateDevEUI(devEUI);
 
             var requestBody = await req.ReadAsStringAsync();
--- a/LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs
@@ -137,7 +137,10 @@ namespace LoraKeysManagerFacade
                         }
 
                         // class c device that did not send a single upstream message
-                        return new ObjectResult("Class C devices must sent at least one message upstream. None has been received") { StatusCode = (int)HttpStatusCode.InternalServerError };
+                        return new ObjectResult("Class C devices must sent at least one message upstream. None has been received")
+                        {
+                            StatusCode = (int)HttpStatusCode.InternalServerError
+                        };
                     }
 
                     // Not a class C device? Send message using sdk/queue
--- a/LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs
@@ -48,13 +48,16 @@ namespace LoraKeysManagerFacade
             try
             {
                 VersionValidator.Validate(req);
+                EUIValidator.ValidateDevEUI(devEUI);
             }
             catch (IncompatibleVersionException ex)
             {
                 return new BadRequestObjectResult(ex.Message);
             }
-
-            EUIValidator.ValidateDevEUI(devEUI);
+            catch (ArgumentException ex)
+            {
+                return new BadRequestObjectResult(ex.Message);
+            }
 
             var requestBody = await req.ReadAsStringAsync();
             if (string.IsNullOrEmpty(requestBody))
@@ -134,7 +137,7 @@ namespace LoraKeysManagerFacade
                         }
 
                         // class c device that did not send a single upstream message
-                        return new BadRequestObjectResult("Class C devices must sent at least one message upstream. None has been received");
+                        return new ObjectResult("Class C devices must sent at least one message upstream. None has been received") { StatusCode = (int)HttpStatusCode.InternalServerError };
                     }
 
                     // Not a class C device? Send message using sdk/queue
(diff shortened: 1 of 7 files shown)


Tests/E2E/LoRaAPIHelper.cs
```diff
@@ -43,6 +44,15 @@ public static async Task<bool> ResetADRCache(string devEUI)
             return response.IsSuccessStatusCode;
         }
 
+        public static async Task<bool> SendCloudToDeviceMessage(string devEUI, LoRaCloudToDeviceMessage c2dMessage)
+        {
+            var url = new Uri($"{baseUrl}cloudtodevicemessage/{devEUI}?code={authCode}");
```
bastbu, line 49: [nit] we could extract the ceremony of the url creation, `authCode` appending and posting etc in a method. I had something of the following signature in mind `static async Task<bool> PostFunctionEndpointAsync(string path, HttpContent content, CancellationToken cancellationToken)`

Tests/Unit/LoraKeysManagerFacade/SendCloudToDeviceMessageTest.cs
```diff
@@ -366,6 +385,47 @@ public async Task When_Querying_Devices_And_Finds_Single_Gateway_Class_C_Should_
...
+        [Fact]
+        public async Task When_Querying_Devices_And_Finds_No_Gateway_For_Class_C_Should_Return_InternalServerError()
+        {
+            const string devEUI = "0123456789";
+
+            var deviceTwin = new Twin
+            {
+                Properties = new TwinProperties()
+                {
+                    Desired = new TwinCollection($"{{\"DevAddr\": \"03010101\", \"ClassType\": \"C\"}}"),
```
bastbu, line 397: [nit] I would suggest to use the `DevAddr` type to ensure that we always have a valid `DevAddr` here (same for the `devEUI` on L391

Tests/Unit/LoraKeysManagerFacade/SendCloudToDeviceMessageTest.cs
```diff
@@ -366,6 +385,47 @@ public async Task When_Querying_Devices_And_Finds_Single_Gateway_Class_C_Should_
...
+            {
+                Fport = TestPort,
+                Payload = "hello",
+            };
+
+            var actual = await this.sendCloudToDeviceMessage.SendCloudToDeviceMessageImplementationAsync(
+                devEUI,
+                actualMessage);
+
+            Assert.IsType<ObjectResult>(actual);
```
bastbu, line 420: [nit] we can use the return value of the assertion to avoid a cast on the next line

Tests/E2E/ClassCTest.cs
```diff
@@ -54,25 +55,11 @@ public async Task Test_ClassC_Send_Message_From_Direct_Method_Should_Be_Received
                 RawPayload = Convert.ToBase64String(new byte[] { 0xFF, 0x00 }),
             };
 
-            TestLogger.Log($"[INFO] Using service client to call direct method to {TestFixture.Configuration.LeafDeviceGatewayID}/{TestFixture.Configuration.NetworkServerModuleID}");
+            TestLogger.Log($"[INFO] Using service API to send C2D message to device {device.DeviceID}");
             TestLogger.Log($"[INFO] {JsonConvert.SerializeObject(c2d, Formatting.None)}");
 
-            for (var i = 0; i < MAX_MODULE_DIRECT_METHOD_CALL_TRIES; ++i)
```
bastbu, line 60: As per our offline discussion: I would argue that if we do not want to deprecate the direct method path for the user (and always force the user to go through the function endpoint), we should keep the old test to validate the direct method path directly.

The benefits I see from this are:
- The E2E test explicitly asserts a guarantee a functionality that we provide for the user (the direct method may still be used without going through the function)
- The direct method path will not only be implicitly covered, but also explicitly covered. If we ever decide to change the function implementation, we will still cover the direct method path.

bastbu (approved):

Tests/E2E/ClassCTest.cs
MaggieSalak: I see the point of verifying that the direct method path works but at the same time I don't really like the idea of having the test execute essentially the same logic twice (the endpoint path is also using the direct method under the hood). How about instead having another assertion in the test that LNS received the direct method call (e.g. asserting the log line: `[TCPLOG] [itestarm1] 01F5A90000000024: received cloud to device message from direct method: {"DevEUI":"01F5A90000000024","Fport":23,"RawPayload":"/wA=","Payload":null,"Confirmed":false,"MessageId":"7c210b2e-4bb9-43c3-8fea-1a1eec03c8f4","MacCommands":[]}` which is logged by the LNS?
Does anyone else have an opinion here - @p-schuler @danigian

danigian: From my point of view the E2E tests should test the user behavior. If, in the documentation, we specify that the way for sending the C2D messages is through the endpoint, that's what we should test. Adding an additional assertion as Maggie is suggesting would be nice.

If we want to keep the explicit testing of the Direct Method, I would rather write an integration test to actually verify that it's properly invoked and handled (if that does not already exists).

No strong opinion here 👍

LoRaEngine/LoraKeysManagerFacade/SendCloudToDeviceMessage/SendCloudToDeviceMessage.cs
```diff
@@ -125,7 +137,10 @@ public async Task<IActionResult> SendCloudToDeviceMessageImplementationAsync(str
                         }
 
                         // class c device that did not send a single upstream message
-                        return new BadRequestObjectResult("Class C devices must sent at least one message upstream. None has been received");
+                        return new ObjectResult("Class C devices must sent at least one message upstream. None has been received")
```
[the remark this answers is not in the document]
MaggieSalak: My reasoning was that BadRequest would mean that there was an issue with data provided by the user. In this case the request is valid but we can't determine the correct gateway. The function doesn't know what the root cause of that is, just that the device twin has no gateway id set so from that point of view I think server error is a slightly better fit.

Tests/E2E/ClassCTest.cs
MaggieSalak: In that case let's leave the E2E test simply using the function endpoint. I added the assertion checking that the direct method usage is logged in the LNS.

Issue #774, opened 2021-11-10
Remove unnecessary/accidental dynamic bindings


atifaziz:
## Expected Behavior

Most `dynamic` bindings can be statically bound or resolved.

## Current Behavior

Lots of member accesses are being dynamically bound due to use of `dynamic`, which can lead to run-time errors and degraded performance.

PR #1181, opened 2022-01-04
Remove dynamic bindings

p-schuler:
# PR for issue #774 

## What is being addressed

The use of dynamics has a negative performance impact when working with dynamics. There are a lot of places, where this was not necessary and the runtime evaluations could be removed from the IL. 

## How is this addressed
All dynamic use is removed except for the use in `CreateEdgeDevice` (which is used to setup edge devices from the starter kit template and is not perf critical). 

The wrapping in a reader was done to have easy access to a logger. This way we can now log all parsing failures in a single place.

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
@@ -82,23 +82,30 @@ namespace LoraKeysManagerFacade
             }
 
             var twin = await this.registryManager.GetTwinAsync(stationEui, cancellationToken);
-            var twinReader = new TwinCollectionReader(twin.Properties.Desired, this.logger);
-            this.logger.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui.ToString());
-            try
+            if (twin != null)
             {
-                var cupsProperty = twinReader.ReadRequiredString(CupsPropertyName);
-                var parsedJson = JObject.Parse(cupsProperty);
-                var url = credentialType is ConcentratorCredentialType.Lns ? parsedJson[LnsCredentialsUrlPropertyName].ToString()
-                                                                            : parsedJson[CupsCredentialsUrlPropertyName].ToString();
-                var result = await GetBase64EncodedBlobAsync(url, cancellationToken);
-                return new OkObjectResult(result);
+                this.logger.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui.ToString());
+                try
+                {
+                    var cupsProperty = ((JObject)(object)twin.Properties.Desired[CupsPropertyName]).ToString();
+                    var parsedJson = JObject.Parse(cupsProperty);
+                    var url = credentialType is ConcentratorCredentialType.Lns ? parsedJson[LnsCredentialsUrlPropertyName].ToString()
+                                                                               : parsedJson[CupsCredentialsUrlPropertyName].ToString();
+                    var result = await GetBase64EncodedBlobAsync(url, cancellationToken);
+                    return new OkObjectResult(result);
+                }
+                catch (Exception ex) when (ex is ArgumentOutOfRangeException
+                                              or JsonReaderException
+                                              or InvalidOperationException)
+                {
+                    this.logger.LogError(ex, "'{PropertyName}' desired property was not found or misconfigured.", CupsPropertyName);
+                    return new UnprocessableEntityResult();
+                }
             }
-            catch (Exception ex) when (ex is ArgumentOutOfRangeException
-                                          or JsonReaderException
-                                          or InvalidOperationException)
+            else
             {
-                this.logger.LogError(ex, "'{PropertyName}' desired property was not found or misconfigured.", CupsPropertyName);
-                return new UnprocessableEntityResult();
+                this.logger.LogInformation($"Searching for {stationEui} returned 0 devices");
+                return new NotFoundResult();
             }
         }
 
--- a/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
@@ -11,6 +11,7 @@ namespace LoraKeysManagerFacade
     using Azure.Storage.Blobs;
     using Azure.Storage.Blobs.Models;
     using LoRaTools.CommonAPI;
+    using LoRaTools.Utils;
     using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Mvc;
     using Microsoft.Azure.Devices;
@@ -29,17 +30,19 @@ namespace LoraKeysManagerFacade
         internal const string LnsCredentialsUrlPropertyName = "tcCredentialUrl";
         private readonly RegistryManager registryManager;
         private readonly IAzureClientFactory<BlobServiceClient> azureClientFactory;
+        private readonly ILogger<ConcentratorCredentialsFunction> logger;
 
         public ConcentratorCredentialsFunction(RegistryManager registryManager,
-                                               IAzureClientFactory<BlobServiceClient> azureClientFactory)
+                                               IAzureClientFactory<BlobServiceClient> azureClientFactory,
+                                               ILogger<ConcentratorCredentialsFunction> logger)
         {
             this.registryManager = registryManager;
             this.azureClientFactory = azureClientFactory;
+            this.logger = logger;
         }
 
         [FunctionName(nameof(FetchConcentratorCredentials))]
         public async Task<IActionResult> FetchConcentratorCredentials([HttpTrigger(AuthorizationLevel.Function, "get", Route = null)] HttpRequest req,
-                                                                      ILogger log,
                                                                       CancellationToken cancellationToken)
         {
             if (req is null) throw new ArgumentNullException(nameof(req));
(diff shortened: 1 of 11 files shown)


LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
```diff
@@ -50,56 +53,62 @@ public class ConcentratorCredentialsFunction
...
             if (twin != null)
             {
-                log.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui.ToString());
+                this.logger.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui.ToString());
                 try
                 {
-                    var cupsProperty = (string)twin.Properties.Desired[CupsPropertyName].ToString();
+                    const string cupsKey = "cups";
+                    if (!twin.Properties.Desired.TryReadJsonBlock(cupsKey, this.logger, out var cupsProperty))
+                        throw new ArgumentOutOfRangeException(cupsKey, "failed to parse cups config");
```
bastbu, line 92: [nit] why do we throw an `ArgumentOutOfRangeException` here? Maybe it would be a bit more accurate to mention that it's not the parsing that goes wrong, but more the retrieval of the JSON string.

Tests/Unit/LoRaTools/TwinCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            Assert.Throws<InvalidOperationException>(() => reader.ReadRequiredString(key));
+        }
+
+        [Fact]
+        public void ReadRequiredString_Reads_Valid_String()
+        {
+            const string key = "test";
+            const string expectedValue = "expected";
+
+            var tc = CreateTwinCollectionReader(key, QuoteJsonString(expectedValue));
```
bastbu, line 60: [nit] maybe it would be a bit simpler to use a specific overload for strings instead of always needing to think about quoting the JSON strings:

```csharp
private TwinCollectionReader CreateTwinCollectionReader(string key, string someString) =>
            CreateTwinCollectionReader<string>(key, $"'{someString}'");
```

Tests/Unit/LoRaTools/TwinCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            Assert.True(tc.TryRead<DataRateIndex>(key, out var dr));
+            Assert.Equal(DataRateIndex.DR8, dr);
+        }
+
+        [Fact]
+        public void Undefined_Enum_Values_Are_Not_Parsed()
+        {
+            const string key = "test";
+            var tc = CreateTwinCollectionReader(key, 100);
+            Assert.False(tc.TryRead<DataRateIndex>(key, out _));
```
bastbu, line 141: [nit] should we also assert the value of the discarded result here?

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
```diff
@@ -876,5 +876,144 @@ public async Task When_Initialized_With_Class_C_And_Custom_RX2DR_Should_Have_Cor
...
+
+            private LoRaDeviceTest CreateDefault()
+                => new LoRaDeviceTest(this.loRaDeviceClient.Object);
+
+            private class LoRaDeviceTest : LoRaDevice
+            {
+                private readonly ILogger<TwinCollectionReader> logger = NullLogger<TwinCollectionReader>.Instance;
+
+                public LoRaDeviceTest(ILoRaDeviceClient deviceClient)
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
bastbu, line 1006: [nit] maybe a comment that ownership is transferred to the `LoRaDevice` would be helpful, then it's not necessary to look through the code to figure out if it's a bug that it's suppressed.

Side note: apparently we just `Release()` the `connectionManager` inside the loradevice, but we don't `Dispose()` it on disposal. Do we need to change this? (It should execute the same logic as of now, but that might change in the future, no?)

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
```diff
@@ -226,10 +229,8 @@ public async Task When_Initialized_ABP_Device_Should_Have_All_Properties()
         [Theory]
         [InlineData("false")]
         [InlineData("FALSE")]
-        [InlineData("0")]
         [InlineData(0)]
         [InlineData(false)]
-        [InlineData("ture")] // misspelled "true"
```
bastbu, line 232: Why remove this case?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/TwinCollectionExtensions.cs
```diff
@@ -0,0 +1,119 @@
...
+    public static class TwinCollectionExtensions
+    {
+        private static readonly Dictionary<Type, Func<object, object>> customConverters = new Dictionary<Type, Func<object, object>>
+        {
+            [typeof(StationEui)] = (s) => StationEui.Parse(s.ToString()),
+            [typeof(DevNonce)] = (s) => new DevNonce(Convert.ToUInt16(s, CultureInfo.InvariantCulture)),
+            [typeof(DevAddr)] = (s) => DevAddr.Parse(s.ToString()),
+        };
+
+        public static T? SafeRead<T>(this TwinCollection twinCollection, string property, T? defaultValue = default, ILogger? logger = null)
```
bastbu, line 27: [nit] `SafeRead<T>` and `ReadRequiredString` are currently only used from the `TwinCollectionReader`. We could move it there entirely, which would increase encapsulation a bit.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/TwinCollectionExtensions.cs
```diff
@@ -0,0 +1,119 @@
...
+            // quick path for values that can be directly converted
+            if (some is JValue someJValue)
+            {
+                if (someJValue.Value is T someT)
+                {
+                    value = someT;
+                    return true;
+                }
+
+                if (someJValue.Type == JTokenType.Null)
```
bastbu, line 56: I think that this might cause a problem in the following test case (does not yet exist):

`Assert.True(reader.TryRead<ushort?>(null, out var result));`

I would have expected that parsing a `null` into a nullable type should succeed.

LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
```diff
@@ -50,56 +53,58 @@ public class ConcentratorCredentialsFunction
...
             {
-                log.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui.ToString());
+                this.logger.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui.ToString());
                 try
                 {
-                    var cupsProperty = (string)twin.Properties.Desired[CupsPropertyName].ToString();
-                    var parsedJson = JObject.Parse(cupsProperty);
-                    var url = credentialType is ConcentratorCredentialType.Lns ? parsedJson[LnsCredentialsUrlPropertyName].ToString()
-                                                                               : parsedJson[CupsCredentialsUrlPropertyName].ToString();
+                    var someJObj = (JObject)(object)twin.Properties.Desired[CupsPropertyName];
```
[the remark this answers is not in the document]
p-schuler: We need to cast it directly to the object to ensure we can avoid the use of CallSite. The compiler can be sure that we can cast it to the object and from there, we take responsibility to cast it to the (JObject) - btw, the cast was changed, as it turns out the returned type on IoT hub is in fact a different type. But that's solved.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
```diff
@@ -229,23 +230,15 @@ private async Task<List<DevAddrCacheInfo>> GetDeviceTwinsFromIotHub(RegistryMana
...
-                        if (twin.Properties.Desired.Contains(LoraKeysManagerFacadeConstants.TwinProperty_DevAddr))
-                        {
-                            rawDevAddr = twin.Properties.Desired[LoraKeysManagerFacadeConstants.TwinProperty_DevAddr].Value as string;
-                        }
-                        else if (twin.Properties.Reported.Contains(LoraKeysManagerFacadeConstants.TwinProperty_DevAddr))
-                        {
-                            rawDevAddr = twin.Properties.Reported[LoraKeysManagerFacadeConstants.TwinProperty_DevAddr].Value as string;
-                        }
-                        else
+                        if (!twin.Properties.Desired.TryRead(LoraKeysManagerFacadeConstants.TwinProperty_DevAddr, this.logger, out DevAddr devAddr) &&
```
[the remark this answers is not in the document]
p-schuler: before this was an issue. Imagine you have multiple devices and one of them is broken. We would not parse the rest. Now we skip over that and continue parsing the others. The problem is logged as an error.

LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
p-schuler: also this was changed.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/TwinCollectionExtensions.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    return false;
+                }
+            }
+
+            try
+            {
+                var t = typeof(T);
+
+                if (TryGetCustomConverter(t, out var converter))
```
bastbu, line 66: `TryGetCustomConverter` has only one reference and could be inline to avoid indirection. Alternatively, we could also get rid of the `customConverters` entirely and use something similar to:

```csharp
value = value switch
{
    StationEui => (T)(object)StationEui.Parse(some.ToString()),
    DevNonce => (T)(object)new DevNonce(Convert.ToUInt16(some, CultureInfo.InvariantCulture)),
    DevAddr => (T)(object)DevAddr.Parse(some.ToString()),
    _ => (T)Convert.ChangeType(some, t, CultureInfo.InvariantCulture)
};

if (t.IsEnum && !t.IsEnumDefined(value))
{
    value = default;
    LogParsingError(logger, property, some);
    return false;
}
```

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
```diff
@@ -876,5 +876,144 @@ public async Task When_Initialized_With_Class_C_And_Custom_RX2DR_Should_Have_Cor
 #pragma warning disable CA2000 // Dispose objects before losing scope
         private LoRaDevice CreateDefaultDevice() => new LoRaDevice("FFFFFFFF", "0000000000000000", new SingleDeviceConnectionManager(this.loRaDeviceClient.Object));
 #pragma warning restore CA2000 // Dispose objects before losing scope
+
+        public class FrameCounterInitTests
```
bastbu, line 880: Nice to see so many tests added. Just curious, why did you decide to test that logic?

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
```diff
@@ -90,22 +95,29 @@ public async Task<string[]> GetAllowedClientThumbprintsAsync(StationEui stationE
...
-                string thumbprintsArrayJson = desiredProperties[ClientThumbprintPropertyName].ToString();
-                return JsonReader.Array(JsonReader.String()).Read(thumbprintsArrayJson);
+                try
+                {
+                    var thumbprints = (JArray)(object)desiredProperties[ClientThumbprintPropertyName];
+                    return thumbprints.ToObject<string[]>();
+                }
+                catch (Exception ex) when (ex is InvalidCastException)
+                {
+                    throw new LoRaProcessingException($"'{ClientThumbprintPropertyName}' format is invalid. An array is expected.", ex, LoRaProcessingErrorCode.InvalidDeviceConfiguration);
```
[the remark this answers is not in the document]
p-schuler: fixed in cf8bf974f9756d49a40f075f6acd5a80e3fca31f

LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
p-schuler: 64d96951736fe0dc8930d143186c77566166178b

Tests/Unit/LoRaTools/TwinCollectionExtensionsTests.cs
p-schuler: the result is default when the method returns false. That's another test.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
p-schuler: I'll add the comment. I thought it was obvious, as we do not keep a reference anywhere and pass it into the ctor of the item. It's a typical DI pattern. 

The connection manager is a singleton that is managing single device connections. We do not dispose it unless we are shutting down the app. We only release the connection that is managed for that particular device. There are a few issues with the pattern as it is implemented atm, and all of those are tracked as part of the caching improvement feature.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
p-schuler: I think this was wrong. The default is true. When we are getting a value we can't parse and we changed the default? I fixed that to keep the default for values we can't parse. I think that makes sense. Sorry, I wanted to mention that in my notes and forgot about it, but good catch. :)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/TwinCollectionExtensions.cs
p-schuler: The model is to expose everything through the extensions in the TwinCollection and have the reader as convenient to wrap the logging. I would like to keep that.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/TwinCollectionExtensions.cs
p-schuler: I think it's still correct. You can't read anything that was not specified, hence the value you get back is correct (default) but the read was not successful. Anyway, you can't specify twins with null values as far as I know. They are removed when you try to do so.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/TwinCollectionExtensions.cs
p-schuler: I think the case of custom converters is different and we do not want to go through the additional code we are having in there (enum validation). 

The reason for the additional method is the nullable types. The dictionary can't indicate that the out is not null when it returns true, hence the redirection. I'll keep it the way it is.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
p-schuler: because I needed to update it with my refactoring and simplified some of the cases. I used TDD to ensure the behavior of the method wasn't change.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
p-schuler: 0e5a8b092caca1617ffa0117132f6e61f1e9f0c0

Tests/Unit/LoRaTools/TwinCollectionExtensionsTests.cs
p-schuler: aeaea698600306ca283c59cf31a9ae4fdc5e8268

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/TwinCollectionExtensions.cs
p-schuler: When going through the change, I realized that this was wrong. Since IoT hub does sanitize null values and remove them in the twins, we should not have tests that actually assume this would be the case. I removed support for that in the twin extensions completely.

Issue #1185, opened 2022-01-05
Refresh device on re-join


p-schuler:
## Expected Behavior
When rejoining a device in a multi gateway environment, all gateways should be able to process requests from the newly joined device.

## Current Behavior
In multi gateway scenarios, where multiple GW have an already joined OTAA device in cache, a rejoin will only update the GW that processes the request. The other GW will have a stale cache until it is refreshed and won't be able to process requests.

## Steps to Reproduce
1. GW 1 receives the join, creates the device and adds it to the cache
2. GW 2 receives the join, creates the device and adds it to the cache
3. GW 1 wins the race to processes the join, adding session keys and DevAddr
4. GW 2 loses the race and does not process the request
**note**: GW2 now has a cached device in join state. Meaning there is no DevAddr. We do have a bug to clean up that state #1024. 
5. The device starts sending message. 
6. GW 2, detects, it has a join state device in the cache and refreshes the state to fetch the DevAddr and sessionkeys
7. Both GW can process messages
8. The device does re-join the network for some reason
9. GW 1 receives the join, wins the race, updates the session keys and dev addr
10. GW 2 receives the join, loses the race and does not update the state
11. When the device now sends messages, the sate on GW 2 won't allow it to process messages. It needs to refresh the device

p-schuler:
We fixed this in #1197. The reason why we do not want to refresh on each message receive is to avoid DOS attacks from bogus devices.

PR #1188, opened 2022-01-05
Refresh devices in cache when loading

p-schuler:
This ensures that we refresh session keys when a re-join happens

# PR for issue #1185

## What is being addressed

We did not refresh the cache on re-joins on GW that lost the race (did not process the join). Hence those session keys got outdated.

## How is this addressed
We always refresh the twins for existing cached items.


bastbu (approved):

p-schuler:
We fixed this in #1197. The reason why we do not want to refresh on each message receive is to avoid DOS attacks from bogus devices.

Issue #820, opened 2021-11-16
Delivery of class C messages on Network server side


roel4ez:
There are currently extraordinarily few checks being made on the Class C sending. There are two diverse ways of reaching the class c messages:
- From the C2D method as described in #821
- From [the decoders](https://github.com/Azure/iotedge-lorawan-starterkit/blob/3043ae1c900390b84f0689fceb4a3478d56e537a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs#L559) where an upstream message coming from a class A message triggers a response from a class C device.

In case it is coming from the second we would need to add a few safeguards from the processing in the send Async method:
- [x] We try to send the message to the last connected basicstation. If we suceed we complete the message
- [x] In case the transient checks fails, we need to abandon the C2D message.

PR #1191, opened 2022-01-05
Verify completion of sending a C2D message to a class C device

MaggieSalak:
# PR for issue #820 

## What is being addressed

- When C2D message was sent to a class C device, we were not checking whether the operation completed successfully which is the case when the send queue has finished sending the message on the correct WebSocket.

## How is this addressed

- Instead of discarding the result of the send operation, `SendDownstreamAsync()` in `DownstreamSender` now returns a bool equal true if the operation completed successfully.
- `DefaultClassCDevicesMessageSender` completes the C2D message if the send operation completed successfully or abandons it if that's not the case.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
```diff
@@ -33,19 +33,22 @@ internal class DownstreamSender : IPacketForwarder
...
         {
             if (message is null) throw new ArgumentNullException(nameof(message));
             if (message.StationEui == default) throw new ArgumentException($"A proper StationEui needs to be set. Received '{message.StationEui}'.");
 
             if (this.socketWriterRegistry.TryGetHandle(message.StationEui, out var webSocketWriterHandle))
             {
                 var payload = Message(message);
-                await webSocketWriterHandle.SendAsync(payload, CancellationToken.None);
+                var task = webSocketWriterHandle.SendAsync(payload, CancellationToken.None);
+                await task;
```
p-schuler, line 45: if the task does not complete successfully, this line will throw, hence the next line would always be true IMHO?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultClassCDevicesMessageSender.cs
```diff
@@ -112,19 +112,36 @@ public async Task<bool> SendAsync(IReceivedLoRaCloudToDeviceMessage message, Can
...
                 {
-                    this.logger.LogError($"[class-c] failed to reject. Id: {message.MessageId ?? "undefined"}");
+                    this.logger.LogError($"[class-c] failed to reject. Id: {messageIdLog}");
                 }
                 return false;
             }
             else
             {
-                await this.packetForwarder.SendDownstreamAsync(downlinkMessageBuilderResp.DownlinkMessage);
+                if (await this.packetForwarder.SendDownstreamAsync(downlinkMessageBuilderResp.DownlinkMessage))
```
p-schuler, line 129: I think we should let the exception bubble up and Abandon on exceptions + rethrow and complete otherwise.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownstreamSender.cs
MaggieSalak: I don't really understand - if that line would always be true then what would be the point of exposing `IsCompletedSuccessfully()` in `ValueTask`?
I saw we already do something similar in a few places in the project e.g. in [`DeviceLoaderSynchronizer`](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DeviceLoaderSynchronizer.cs#L171)

p-schuler: The example of the DeviceLoaderSynchronizer is different. We do have a WhenAll, if one of them fails and exception is thrown and we catch it (if it is within the condition of the catch). When going through the list, they are in a unknown state, hence the check for the completion. Here we have a single task that we are awaiting and that will throw if there is an exception.

On a related note: what we do in the DLS is not safe either. There might be tasks completing after the fact we checked and that result will be discarded. It's a bit indeterministic. I believe a Task.WhenAny loop would be better there and then handle individual results as they are available and initialize or discard accordingly.

MaggieSalak: Thanks for explaining. I opened a bug #1212 for the DeviceLoaderSynchronizer issue; since it's unrelated to this PR it makes sense to track it separately.

p-schuler: After discussion and validation, we can close 1212 again. It's going to wait for all of them before raising the exception, hence it is deterministic. The only improvement we would get is quicker initialization.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultClassCDevicesMessageSender.cs
```diff
@@ -112,19 +112,39 @@ public async Task<bool> SendAsync(IReceivedLoRaCloudToDeviceMessage message, Can
                 fcntDown,
                 this.logger);
 
+            var messageIdLog = message.MessageId ?? "undefined";
+
             if (downlinkMessageBuilderResp.IsMessageTooLong)
             {
                 this.c2dMessageTooLong?.Add(1);
-                this.logger.LogError($"[class-c] cloud to device message too large, rejecting. Id: {message.MessageId ?? "undefined"}");
+                this.logger.LogError($"[class-c] cloud to device message too large, rejecting. Id: {messageIdLog}");
```
p-schuler, line 120: can we use structured logging instead?

p-schuler (commented):
Let's use structured logging for the MessageId. Otherwise this looks good.

p-schuler:
Sorry, realized, this was already merged. Not a big deal.

Issue #1024, opened 2021-12-09
Clear cache, when receiving valid join request


p-schuler:
## Expected Behavior
When a device re-joins, gateways keeping outdated session keys should invalidate their cache, when receiving the join.

## Current Behavior
1. Device A sends a join
2. Gateway GA receives the join and wins the race
3. Gateway GB receives the join and loses the race
4. Gateway GA loads the device, creates the keys, updates its cache and sends the join accept
5. The join accept is not received by the device (lost, misses window e.g.)
6. Device A sends a new join
7. GB receives the join and wins the race
8. GA receives the join and loses the race
9. Gateway GB loads the device, creates the keys, updates its cache and sends the join accept
10. Device A receives the join accept and creates the session key using the app nonce

In this situation, GA does have a session key for device A that is not matching the device's session keys. When receiving messages from Device A we do not process them but simply ignore the messages as not being for one of our devices - we do have the DevAddr in the cache but can't find a matching key. 

This situation will eventually resolve itself by the cache refresh operation. However, it will take up to that configured time to start processing the messages from that device again.

## Mitigation
We can make sure that if we receive a valid join request from the device (mic check was performed) - we clear the cache for that particular DevEUI. We can optimize further for only doing this on the device that loses the race for that particular join - the winning device will update the key in the cache.

This will not solve the problem for all cases. If the 2nd join is not received by the gateway that has an outdated cache - it will fallback to the cache refresh interval to get fixed. That's a possible scenario in cases where the device is moving for example.

PR #1197, opened 2022-01-06
Clearing the cache on re-join

p-schuler:
# PR for issue #1024

## What is being addressed

When receiving a re-join of an already cached device in a multi gw scenario, the gateways that lose the join processing race, will be left with a stale session key, making it impossible to process messages until the cache is refreshed.

## How is this addressed

We clear that particular device from the cache, when we lose the race during the join. That will make sure the session key will be re-fetched on the next message. Note: there is a chance that the device won't get the request at all and then the session keys will only be refreshed on the periodic cache refresh we perform.

Closing #1024


Tests/Unit/NetworkServer/LoRaDeviceRegistryTest.cs
```diff
@@ -44,6 +44,28 @@ public async Task GetDeviceForJoinRequestAsync_When_Device_Api_Throws_Error_Shou
...
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task GetDeviceForJoinRequestAsync_When_Join_Handled_By_Other_Cache_Is_Updated(bool joinedDevice)
+        {
+            var devNonce = new DevNonce(1);
+            var apiService = new Mock<LoRaDeviceAPIServiceBase>();
+            var otaaDevice = TestDeviceInfo.CreateOTAADevice(1);
+            if (joinedDevice) otaaDevice.AppSKey = "123";
```
bastbu, line 55: [nit] you could create the key here using `TestKeys` (will make the build green again)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
```diff
@@ -215,6 +215,15 @@ public async Task<LoRaDevice> GetDeviceForJoinRequestAsync(string devEUI, DevNon
 
             if (searchDeviceResult.IsDevNonceAlreadyUsed)
             {
+                // another gateway processed the join request. If we have it in the cache
+                // with existing session keys, we need to invalidate that entry, to ensure
+                // it gets re-fetched on the next message
+                if (this.deviceCache.TryGetByDevEui(devEUI, out var someDevice) && someDevice.AppSKey != null)
```
bastbu, line 221: I don't understand the case `someDevice.AppSKey != null`, why would we have it in the cache without an `AppSKey`?

bastbu (approved):

Tests/Unit/NetworkServer/LoRaDeviceRegistryTest.cs
p-schuler: Already fixed. Thanks. Didn't know about the TestKeys. Will use them when I need real keys.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
p-schuler: As discussed: when a join request is received, we do put it into the cache without the session keys yet initialized. That means, we can't handle incoming messages, but we can use it for other join messages (if we miss the join receive window).

PR #1202, opened 2022-01-06
Adding cancellation token to LoRaDeviceClient.UpdateReportedPropertiesAsync

### LINKED ISSUE #1194: Intermittent integration tests failure
## Current Behavior

https://github.com/Azure/iotedge-lorawan-starterkit/runs/4724512417?check_suite_focus=true has the following test failing

```
LoRaWan.Tests.Integration.JoinSlowTwinUpdateTests.When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Attempt_Should_Succeed(deviceGatewayID: null) [10 s]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at LoRaWan.Tests.Integration.JoinSlowTwinUpdateTests.When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Attempt_Should_Succeed(String deviceGatewayID) in /home/runner/work/iotedge-lorawan-starterkit/iotedge-lorawan-starterkit/Tests/Integration/JoinSlowTwinUpdateTests.cs:line 95
--- End of stack trace from previous location ---
```

danigian:
## What is being addressed

This PR is adding a cancellation token for LoRaDeviceClient.UpdateReportedPropertiesAsync method.

The reported properties are either updated when dealing with a join requests or when dealing with normal data frames.
In case a join request is being handled, we want the token to be cancelled before the end of the second window, otherwise we would not be able to be in time for sending a join accept frame back.

In case a normal data frame is being handled, we can be more 'relaxed' and therefore we are setting with this PR a timeout of 10 seconds for these scenarios.

Most of the changes of this PR are dealing with fixing tests.

This PR also fixes #1194 intermittent failures


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -57,6 +58,7 @@ internal async Task ProcessJoinRequestAsync(LoRaRequest request)
             try
             {
                 var timeWatcher = request.GetTimeWatcher();
+                using var joinAcceptCancellationToken = new CancellationTokenSource(timeWatcher.InTimeForJoinAccept() ? timeWatcher.GetRemainingTimeToJoinAcceptSecondWindow() : TimeSpan.Zero);
```
p-schuler, line 61: The cts will be canceled immediately, if we are already outside the join accept window. I think that's ok, but we should make sure the behavior is not changed. Previously, for missing the window, we log, but we do not throw. This would now cause the request to immediately throw and that exception would then abort the request, but logged as an error in the system. 

I would suggest to handle that the same way we used to. If we get the OperationCanceledException that was triggered by our TCS we should handle it the same way as when we missed the window.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -213,6 +213,8 @@ public LoRaRegionType LoRaRegion
 
         public StationEui LastProcessingStationEui => this.lastProcessingStationEui.Get();
 
+        private readonly TimeSpan twinUpdateTimeout = TimeSpan.FromSeconds(10);
```
p-schuler, line 216: nit: we can make this a static. Can you also try to move it to the private fields? I know it's already a bit a mess, but there are some groups in there.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -106,17 +106,15 @@ public async Task<Twin> GetTwinAsync(CancellationToken cancellationToken = defau
...
             try
             {
-                this.deviceClient.OperationTimeoutInMilliseconds = 120000;
-
                 SetRetry(true);
 
                 this.logger.LogDebug("updating twin");
 
-                await this.deviceClient.UpdateReportedPropertiesAsync(reportedProperties);
+                await this.deviceClient.UpdateReportedPropertiesAsync(reportedProperties, cancellationToken);
```
p-schuler, line 117: we kind of rely on the fact that the cancellationToken is not None now, should we enforce it in here?

Tests/Integration/JoinSlowTwinUpdateTests.cs
```diff
@@ -41,30 +41,29 @@ public async Task When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Att
...
-            string afterJoin1DevAddr = null;
             AppSessionKey? afterJoin2AppSKey = null;
             NetworkSessionKey? afterJoin2NwkSKey = null;
             string afterJoin2DevAddr = null;
 
-            LoRaDeviceClient.Setup(x => x.UpdateReportedPropertiesAsync(It.IsNotNull<TwinCollection>()))
-                .ReturnsAsync(true, TimeSpan.FromSeconds(10))
-                .Callback<TwinCollection>((updatedTwin) =>
+            var invocations = 0;
+            LoRaDeviceClient.Setup(x => x.UpdateReportedPropertiesAsync(It.IsNotNull<TwinCollection>(), It.IsAny<CancellationToken>()))
```
p-schuler, line 49: use SetupSequence instead of trying to keep track of the invocations manually.

Tests/Integration/JoinSlowTwinUpdateTests.cs
```diff
@@ -41,30 +41,29 @@ public async Task When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Att
...
-                        .Callback<TwinCollection>((updatedTwin2) =>
-                        {
-                            afterJoin2AppSKey = AppSessionKey.Parse(updatedTwin2[TwinProperty.AppSKey].Value);
-                            afterJoin2NwkSKey = NetworkSessionKey.Parse(updatedTwin2[TwinProperty.NwkSKey].Value);
-                            afterJoin2DevAddr = updatedTwin2[TwinProperty.DevAddr];
-                        });
+                    if (invocations == 0)
+                    {
+                        invocations++;
+                        await Task.Delay(7_000, CancellationToken.None);
```
p-schuler, line 55: shouldn't we pass the token here?

p-schuler: then you don't need the code below.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
p-schuler: to avoid 2 calculations (InTimeForJoinAccept and GetRemaining...) you could get the remaining time and make sure it's not negative. Also, we may want to add some margin to it to allow for cleanup. If we deduct ~80ms we should be able to stop the processing before the next join comes in.

p-schuler (changes requested):
We need to fix the error handling in the join.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
danigian: As discussed offline, no throwing will happen.

I fixed the catched exception in 711c8ee and the timeout in b5208f6 and in b5208f6

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
danigian: Made internal static for reusal with 55b1819

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
danigian: As discussed, I am now enforcing a default timeout if cancellation token is none

Tests/Integration/JoinSlowTwinUpdateTests.cs
danigian: In the test I want to actually assert that the first request is cancelled after 7 seconds while the second join request is just going through properly. Added a comment and the Sequence setup in 9189435

Tests/Integration/JoinSlowTwinUpdateTests.cs
danigian: Added the sequence setup in 9189435

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -31,6 +31,8 @@ public class LoRaDevice : IDisposable, ILoRaDeviceRequestQueue
         /// </summary>
         internal const ushort DefaultJoinValues = 0;
 
+        internal static readonly TimeSpan TwinUpdateTimeout = TimeSpan.FromSeconds(10);
```
p-schuler, line 34: let's move this to the LoRaDeviceClient instead and make it the default timeout. Seems odd, to reference the default timeout in the LoRaDeviceClient from here. The default should be driven by the device client wrapper and callers can adjust it if needed.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -106,29 +106,36 @@ public async Task<Twin> GetTwinAsync(CancellationToken cancellationToken = defau
...
-        public async Task<bool> UpdateReportedPropertiesAsync(TwinCollection reportedProperties)
+        public async Task<bool> UpdateReportedPropertiesAsync(TwinCollection reportedProperties, CancellationToken cancellationToken)
         {
+            CancellationTokenSource cts = default;
             try
             {
-                this.deviceClient.OperationTimeoutInMilliseconds = 120000;
+                if (cancellationToken == default)
+                {
+                    cts = new CancellationTokenSource(LoRaDevice.TwinUpdateTimeout);
```
p-schuler, line 116: as mentioned above, let's move the default into this class.

Tests/Integration/JoinSlowTwinUpdateTests.cs
```diff
@@ -41,31 +41,29 @@ public async Task When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Att
...
-                });
+            var mockSequence = new MockSequence();
+            LoRaDeviceClient.InSequence(mockSequence)
+                            .Setup(x => x.UpdateReportedPropertiesAsync(It.IsNotNull<TwinCollection>(), It.IsAny<CancellationToken>()))
+                            .Returns<TwinCollection, CancellationToken>(async (_, token) =>
+                            {
+                                // No cancellation token specified because we want to wait 7 seconds no matter what, to make the "token" being cancelled by logic
+                                await Task.Delay(7_000, CancellationToken.None);
+                                Assert.True(token.IsCancellationRequested, "First request should have been canceled after timeout.");
+                                return false;
```
p-schuler, line 56: The 7s are hard to understand for anyone not knowing the 2n downtime window is 6s. I would use the token - that also makes the test more resilient for future changes to that that window. Something like that?
```suggestion
                                // No cancellation token specified because we want to wait 7 seconds no matter what, to make the "token" being cancelled by logic
                                try
                                {
                                   await Task.Delay(TimeSpan.FromSeconds(20), token);
                                   Assert.True(false, "Token timeout expected");
                                }
                                catch (OperationCanceledException) { }
                                return false;
```

Tests/Integration/JoinSlowTwinUpdateTests.cs
```diff
@@ -41,31 +41,29 @@ public async Task When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Att
...
+                            });
+            LoRaDeviceClient.InSequence(mockSequence)
+                            .Setup(x => x.UpdateReportedPropertiesAsync(It.IsNotNull<TwinCollection>(), It.IsAny<CancellationToken>()))
+                            .Returns<TwinCollection, CancellationToken>((updatedTwin, _) =>
+                            {
+                                afterJoin2AppSKey = AppSessionKey.Parse(updatedTwin[TwinProperty.AppSKey].Value);
```
p-schuler, line 62: can we use the new extension methods on the TwinCollection to read the keys? Also the DevAddr can be read.

p-schuler (changes requested):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
danigian: Done in f8eb224

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
danigian: Done in f8eb224

Tests/Integration/JoinSlowTwinUpdateTests.cs
danigian: Done in 86963a1

Tests/Integration/JoinSlowTwinUpdateTests.cs
danigian: Done in 766db56

Tests/Integration/JoinSlowTwinUpdateTests.cs
```diff
@@ -41,31 +42,33 @@ public async Task When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Att
...
-                        });
-                });
+            DevAddr? afterJoin2DevAddr = null;
+
+            var mockSequence = new MockSequence();
+            LoRaDeviceClient.InSequence(mockSequence)
+                            .Setup(x => x.UpdateReportedPropertiesAsync(It.IsNotNull<TwinCollection>(), It.IsAny<CancellationToken>()))
+                            .Returns<TwinCollection, CancellationToken>(async (_, token) =>
+                            {
+                                // No cancellation token specified because we want to wait 7 seconds no matter what, to make the "token" being cancelled by logic
```
p-schuler, line 54:
```suggestion
```

p-schuler (approved):
LGTM

Tests/Integration/JoinSlowTwinUpdateTests.cs
```diff
@@ -41,31 +42,32 @@ public async Task When_First_Join_Fails_Due_To_Slow_Twin_Update_Retry_Second_Att
...
+                                    Assert.True(false, "Token timeout expected");
+                                }
+                                catch (OperationCanceledException) { }
+                                return false;
+                            });
+            LoRaDeviceClient.InSequence(mockSequence)
+                            .Setup(x => x.UpdateReportedPropertiesAsync(It.IsNotNull<TwinCollection>(), It.IsAny<CancellationToken>()))
+                            .Returns<TwinCollection, CancellationToken>((updatedTwin, token) =>
+                            {
+                                _ = updatedTwin.TryRead(TwinProperty.AppSKey, null, out afterJoin2AppSKey);
```
p-schuler, line 66: if you don't need the bool result, you can actually use SafeRead instead.

PR #1205, opened 2022-01-07
Fix docs deployment

roel4ez:
[publish dev docs](https://github.com/Azure/iotedge-lorawan-starterkit/runs/4737947744?check_suite_focus=true) pipeline failed because of missing dependencies.

This PR fixes that by ensuring dependencies are installed from `requirements.txt`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,6 @@ mkdocs==1.2.3
 mkdocs-material==8.1.3
 pymdown-extensions==9.1
 markdown==3.3.6
+mike==1.1.2
 mdx_truly_sane_lists==1.2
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -2,5 +2,6 @@ mkdocs==1.2.3
 mkdocs-material==8.1.3
 pymdown-extensions==9.1
 markdown==3.3.6
+mike==1.1.2
```
p-schuler, line 5: Mike is back! ;)

p-schuler (approved):

requirements.txt
roel4ez: HahaMickaelJacksonGIF

PR #1206, opened 2022-01-07
Reuse `Mic`

bastbu:
# PR for issue #971 

## What is being addressed

Use the `Mic` type when possible for MIC calculations. With this PR we also add some unit tests as sanity checks for the MIC calculations, which did previously not have any unit tests and were only covered through integration tests.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Mic.cs
```diff
@@ -42,13 +45,23 @@ public Span<byte> Write(Span<byte> buffer)
         // Source: https://www.thethingsnetwork.org/docs/lorawan/message-types/#calculating-the-message-integrity-code
 
         public static Mic ComputeForJoinRequest(AppKey appKey, MacHeader mhdr, JoinEui joinEui, DevEui devEui, DevNonce devNonce)
+        {
+            var keyBytes = new byte[AppKey.Size];
+            _ = appKey.Write(keyBytes);
+            return ComputeForJoinRequest(keyBytes, mhdr, joinEui, devEui, devNonce);
+        }
+
+        public static Mic ComputeForJoinRequest(NetworkSessionKey networkSessionKey, MacHeader mhdr, JoinEui joinEui, DevEui devEui, DevNonce devNonce)
```
danigian, line 54: I think we should get rid of this. Join Request MIC should not be using NwkSKey (also because it's not yet generated when we are doing a join request, but derived after the join from the AppKey)

danigian (changes requested):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Mic.cs
bastbu: Done in 4bb7e18ab4d52ba2b57bb226be1cf8f5c0da429b

danigian (approved):

PR #1208, opened 2022-01-10
added new NOTICE file LNS_NOTICE 41481

roel4ez:


roel4ez:
only order change, ignore

PR #1215, opened 2022-01-10
Reuse `NetId`

bastbu:
# PR for issue #972

## What is being addressed

Reuses the type `NetId` wherever possible. With this PR we also fix a small bug with respect to range validation in `NetId`.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Hexadecimal.cs
```diff
@@ -105,6 +105,19 @@ public static bool TryParse(ReadOnlySpan<char> chars, out uint value, char? sepa
...
+            const int size = sizeof(int);
+            if (chars.IsEmpty || chars.Length != (separator is null ? size * 2 : (size * 3) - 1))
+                return false;
+            Span<byte> bytes = stackalloc byte[size];
+            if (!TryParse(chars, bytes, separator))
+                return false;
+            value = BinaryPrimitives.ReadInt32BigEndian(bytes);
+            return true;
+        }
+
```
atifaziz, line 120: This method is not used anywhere except in tests. Consider removing?

```suggestion
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/NetId.cs
```diff
@@ -31,5 +31,36 @@ public NetId(int value)
         public int NetworkId => this.value & 0b111_1111;
 
         public override string ToString() => this.value.ToString("X6", CultureInfo.InvariantCulture);
+
+        public static NetId Parse(ReadOnlySpan<char> input) =>
+            TryParse(input, out var result) ? result : throw new FormatException();
+
+        public static bool TryParse(ReadOnlySpan<char> input, out NetId result)
+        {
+            if (int.TryParse(input, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var raw))
```
atifaziz, line 40: `raw` is an `int` but a **NetID** is supposed to be a 24-bit number. If `raw` is too big then `NetId` will throw `ArgumentOutOfRangeException`, which shouldn't happen for a method that's meant to _try and parse_. I would add a test to validate that and the fix should be as follows:

```suggestion
            if (int.TryParse(input, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var raw)
                && unchecked((uint)raw & 0xff00_0000) == 0)
```

You could then introduce a helper static method in the type along the lines of:

```c#
private static bool IsSesquiWord(int n) => unchecked((uint)n & 0xff00_0000) == 0;
```

so that the constructor and this method use a single validation logic.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/NetId.cs
```diff
@@ -31,5 +31,36 @@ public NetId(int value)
...
+                result = default;
+                return false;
+            }
+        }
+
+        public Span<byte> Write(Span<byte> buffer)
+        {
+            buffer[2] = (byte)(this.value >> 16);
+            buffer[1] = (byte)(this.value >> 8);
+            buffer[0] = (byte)this.value;
```
atifaziz, line 56: Consider inverting for readability of order (0…2) and use unchecked conversions:

```suggestion
            unchecked
            {
                buffer[0] = (byte)this.value;
                buffer[1] = (byte)(this.value >> 8);
                buffer[2] = (byte)(this.value >> 16);
            }
```

If someone [changes the default for the project](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/language#checkforoverflowunderflow) then some of the byte conversions could overflow.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/NetId.cs
```diff
@@ -31,5 +31,36 @@ public NetId(int value)
...
+            return buffer[Size..];
+        }
+
+        public static NetId Read(ReadOnlySpan<byte> buffer)
+        {
+            var value = buffer[0] + (buffer[1] << 8) + (buffer[2] << 16);
+            return new NetId(value);
+        }
```
atifaziz, line 64: nit: Logical OR-ing seems more natural for what's fundamentally merging of bits:

```suggestion
        public static NetId Read(ReadOnlySpan<byte> buffer) =>
            new(buffer[0] | (buffer[1] << 8) | (buffer[2] << 16));
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
```diff
@@ -10,31 +10,30 @@ namespace LoRaTools
 
     public static class OTAAKeysGenerator
     {
-        public static DevAddr GetNwkId(uint netId)
+        public static DevAddr GetNwkId(NetId netId)
```
atifaziz, line 13: This method seems to be only used in one place and doesn't have much to do with OTAA key generation. Why not consider inlining as part of this PR?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
```diff
@@ -10,31 +10,30 @@ namespace LoRaTools
...
-        public static AppSessionKey CalculateAppSessionKey(byte[] type, byte[] appnonce, byte[] netid, DevNonce devNonce, AppKey appKey)
+        public static AppSessionKey CalculateAppSessionKey(byte[] type, byte[] appnonce, NetId netid, DevNonce devNonce, AppKey appKey)
         {
             var keyString = CalculateKey(type, appnonce, netid, devNonce, appKey);
             return AppSessionKey.Parse(keyString);
         }
 
         // don't work with CFLIST atm
-        private static string CalculateKey(byte[] type, byte[] appnonce, byte[] netid, DevNonce devNonce, AppKey appKey)
+        public static string CalculateKey(byte[] type, byte[] appnonce, NetId netid, DevNonce devNonce, AppKey appKey)
```
atifaziz, line 33: Any reason to change the accessibility of this method? It doesn't look like it's being used beyond this class with this PR.

atifaziz: Also, consider renaming parameter `netid` or `netId` that would be more conventional.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -102,7 +102,7 @@ public class NetworkServerConfiguration
         /// <summary>
         /// Gets or sets the gateway netword id.
         /// </summary>
-        public uint NetId { get; set; } = 1;
+        public NetId NetId { get; set; } = new NetId(1);
```
atifaziz, line 105: I have doubts whether this is meant to be a `NetId`.  I think this might be more the **NwkID** from the specification, the 7 LSB of a **NetID**.

atifaziz (commented):
See comments. Let me know if any of them need further clarification or expanding. Thanks!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -58,7 +58,7 @@ public class LoRaDevice : IDisposable, ILoRaDeviceRequestQueue
 
         public DevNonce? DevNonce { get; set; }
 
-        public string NetID { get; set; }
+        public NetId? NetID { get; set; }
```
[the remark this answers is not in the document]
bastbu: Addressed this with c17ca9ba119cf66e5af87ad19b876a8d4980a446

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/OTAAHelper/OTAAKeysGenerator.cs
bastbu: Done with 5b4bb44c5fccb14409ab6d5e1c264ef9fe5d4073

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/NetId.cs
bastbu: Applied with 892492d5f0db18378c49299e9601f658fcf3866c

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
atifaziz: Okay, it seems that this is indeed the **NetID** from the specification and therefore using our `NetId` type is the right call here. Sorry, but the confusion was on my end (thus my doubt) that stemmed from not being familiar with the specification in its entirety. Reading the relevant passages and other documents such as the [NetID FAQ](https://lora-alliance.org/wp-content/uploads/2021/01/20200714_NetID_FAQ.pdf) cleared up some of that confusion.

atifaziz (approved):

bastbu:
Since `OTAATest` and `OTAAJoinTest` were successful on 73c00c0ac490d8497250a879a0dd213c14c04cd4, will merge this PR without rerunning the full CI.

PR #1216, opened 2022-01-10
[ADR] CUPS firmware upgrade

MaggieSalak:
# PR for issue #1189 

## What is being addressed

ADR for CUPS firmware upgrade, explaining the flow and required changes in different components.


docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Context
+
+The CUPS request described in [The CUPS protocol - documentation][cupsproto]
+contains a number of fields which are related to the current firmware version of
+the Basics Station. Specifically, `station`, `model`, `package` and `keys`
+fields are dependent on the Station version and need to be updated whenever a
+firmware upgrade of the Basics Station is performed.
+
+### Firmware upgrade flow
```
p-schuler, line 47: We need a good understanding of the flow for the signing keys/signatures and how those are accessed when ingesting. I would tend to get them directly on the command line when ingesting a new firmware - alternatively, key vault could be supported.

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Out-of-scope
+
+## Decisions
+
+### Context
+
+The CUPS request described in [The CUPS protocol - documentation][cupsproto]
+contains a number of fields which are related to the current firmware version of
+the Basics Station. Specifically, `station`, `model`, `package` and `keys`
```
p-schuler, line 43: I think we can keep it simple and only use the package as an indication for the update requirement. If there is a change in station name, we can indicate that with an update of the package version (a new package, will have to be created anyway to deploy a new station, hence, that's somewhat redundant).

docs/adr/007_cups_firmware_upgrade.md
danigian: From my point of view the flow should be the following:

- When provisioning a device that will need firmware upgrades the end user should generate and store on the device and in a centralized repository of his own choice a "sig-0.key" (example on how to generate it [here](https://github.com/lorabasics/basicstation/blob/master/examples/cups/prep.sh))
- When a firmware upgrade should be provided to the device, user should generate a digest of the executable file of the upgrade (as per example above) and retrieve the CRC32 Checksum of the sig-0.key. Then 4 inputs (file, digest, checksum and new version) should be provided to Device Provisioning CLI tool to properly upload the file to a blob storage and store in the concentrator twin the needed information (blob url in storage, digest, checksum and the new version #)

Since this is a starter kit, I would not force users to use KeyVault for storing and retrieving the sig-*.key files, even though we can definitely suggest it in the docs.

What do you think about it @p-schuler, @maggiesalak?

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,129 @@
...
+"cups": {
+  // ...
+  "package": "1.0.1",
+  "fwUrl": "https://...",
+  "fwKeyChecksum": 123456,
+  "fwSignature": ""
+}
+```
+
+- **'package'**: current package version of the Station
```
danigian, line 86:
```suggestion
- **'package'**: desired package version of the Station (matching what will be extracted as version.txt file during update)
```

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,129 @@
...
+
+### LoRa Device Provisioning CLI changes
+
+A new command will be added to the Device Provisioning CLI which will allow the
+user to trigger a firmware upgrade. The command will accept the follwing inputs:
+
+- firmware upgrade file
+- signature
+- CRC32 checksum of the key used for the signature
+- Basics Station package version
```
danigian, line 121:
```suggestion
- new package version
```

danigian: i would omit "basics station package version" but just say new package version, since the update might decide to update not only the basic station but also other components

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,129 @@
...
+
+The CLI tool will:
+
+1. Upload a blob with the firmware file to the storage account.
+1. Update concentrator device twin with the new blob URL, signature and CRC32
+   checksum of the key used to generate the signature.
```
danigian, line 127:
```suggestion
   checksum of the key used to generate the signature and the new package version
```

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,129 @@
...
+### Storage related changes
+
+We will use the same storage solution as the one selected for the general CUPS
+protocol support (storage account). A new container will be used to store the
+firmware upgrade files.
+
+### Azure Function related changes
+
+A new endpoint will be added in the Facade Azure Function which will be used to
+fetch firmware upgrade files from the storage account.
```
danigian, line 102: Let's add some details here.
The endpoint should accept as input the StationEui, then retrieve the twin from IoT Hub, download the FW file from the storage and send it back via the response.

Let's also specify that this mechanism is setting a theoretical limit of 100MB for the firmware upgrade (as we can't process more with Azure Function). This should be enough given the size of the basicstation executable (around 1MB as of now).

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,129 @@
...
+  the upgrade
+- **'fwKeyChecksum'**: checksum of the key used to sign the digest of the
+  firmware upgrade file
+- **'fwSignature'**: signature of the uploaded firmware upgrade file
+
+### Storage related changes
+
+We will use the same storage solution as the one selected for the general CUPS
+protocol support (storage account). A new container will be used to store the
+firmware upgrade files.
```
danigian, line 97: Let's specify that we are going to save the files in a "fwupgrades" container and that the file names will be in the format "{stationEui}-{package}" (without any extension, as anyways these are going to be downloaded as "update.bin" from the basics station executable)

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,129 @@
...
+### Azure Function related changes
+
+A new endpoint will be added in the Facade Azure Function which will be used to
+fetch firmware upgrade files from the storage account.
+
+### Changes in the LoRaWan Network Server
+
+The implementation of `CupsProtocolMessageProcessor` should be extended for
+checking the `package` field from the concentrator device twin. In case there
+the value is different from the one received from the Basics Station in the CUPS
```
danigian, line 108: If the value of 'package' field is different, we have to first retrieve the twin and check whether any of the key in the keys array from the request is actually equal to the fwKeyChecksum field that we have in the twin.

If there is no matching checksum, it means that the concentrator is missing the key required for calculating the digest and verifying the update file. When this happens we should throw an appropriate error

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,129 @@
...
+  "fwSignature": ""
+}
+```
+
+- **'package'**: current package version of the Station
+- **'fwUrl'**: URL pointing to the storage location of the file required to run
+  the upgrade
+- **'fwKeyChecksum'**: checksum of the key used to sign the digest of the
+  firmware upgrade file
+- **'fwSignature'**: signature of the uploaded firmware upgrade file
```
danigian, line 91:
```suggestion
- **'fwSignature'**: signature of the uploaded firmware upgrade file (as base64 encoded string)
```

danigian:
I think we should add an appendix explaining:
- how to generate the signature keys
- how to calculate the checksum of the key
- how to generate the digest of the upgrade files

Even though we are going to automatize it through script, I think it's not a bad idea. What do you think @MaggieSalak?

Generate the signature
```bash
openssl ecparam -name prime256v1 -genkey | openssl ec -out sig-0.pem
openssl ec -in sig-0.pem -pubout -out sig-0.pub
openssl ec -in sig-0.pub -inform PEM -outform DER -pubin | tail -c 64 > sig-0.key
```

Calculate the checksum of the signature
```bash
cat sig-0.key | gzip -1 | tail -c 8 | od -t ${1:-u}4 -N 4 -An --endian=little | xargs echo > sig-0.crc
```

Calculate the digest of an update.sh file
```bash
openssl dgst -sha512 -sign sig-0.pem update.sh > update.sh.sig-0.sha512
```

docs/adr/007_cups_firmware_upgrade.md
MaggieSalak: Makes sense, I update the flow to reflect the steps you mentioned - generating a key and then using the CLI with correct inputs to initiate the upgrade.

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+contains the `package` field which indicates the current firmware version of the
+Basics Station. The value will need to be updated whenever a firmware upgrade of
+the Basics Station is performed by the user.
+
+### Firmware upgrade flow
+
+1. When provisioning a device that will need firmware upgrades, the end user
+   should generate a `sig-0.key` and store it on the device and in a centralized
+   repository of their own choice (example on how to generate it can be found in
+   the appendix, and [LoRa Basics Station repository][lbs-example]).
```
danigian, line 62:
```suggestion
   the appendix).
```
Let's remove the reference to the official repository, the link or the example might change and cause troubles

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+
+Digest of an upgrade file (`upgrade.sh`) can be calculated using the previously
+generated signature key using the command:
+
+```shell
+openssl dgst -sha512 -sign sig-0.pem update.sh > update.sh.sig-0.sha512
+```
+
+[cupsproto]: https://doc.sm.tc/station/cupsproto.html
+[lbs-example]: https://github.com/lorabasics/basicstation/blob/master/examples/cups/prep.sh
```
danigian, line 189:
```suggestion
```
As per previous comment, let's remove this reference

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+
+**Feature**:
+[#1189](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1189)  
+
+**Authors**: Daniele Antonio Maggio, Maggie Salak
+
+**Status**: Proposed
+
+>This ADR is an extension of [006. CUPS Protocol Implementation - Credential
+>management](./docs/adr/006_cups.md) and focuses of firmware upgrades. For
```
danigian, line 11:
```suggestion
>management](./006_cups.md) and focuses of firmware upgrades. For
```
Link was broken

danigian (commented):
Please also add adr/007_cups_firmware_upgrade.md to the navigation list in mkdocs.yml

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,211 @@
...
+    Network Server->>IoT Hub: Retrieve 'Concentrator' twin
+    Network Server->>Network Server: Check if versions of the Station are different
+    alt versions are the same
+        Network Server->>Concentrator: CUPS response without firmware upgrade
+    else versions are different
+      Network Server->>Network Server: Verify if any of the `keys` from CUPS request match the `fwKeyChecksum`
+      alt no match
+        Network Server->>Network Server: Throw as this indicates a misconfiguration
+      else one of the keys matches `fwKeyChecksum`
+        Network Server->>Facade Function: Retrieve firmware upgrade file
```
bastbu, line 94: How big do we expect that these firmware upgrade files can be? Will there be potential restrictions to the size of the firmware upgrade package through the timeout of the Azure Function at lower network speeds?

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Firmware upgrade flow
+
+1. When provisioning a device that will need firmware upgrades, the end user
+   should generate a `sig-0.key` and store it on the device and in a centralized
+   repository of their own choice (example on how to generate it can be found in
+   the appendix).
+
+1. When a firmware upgrade needs to be provided to the device, the user should
+   generate a digest of the executable file of the upgrade (as in examples
```
p-schuler, line 61: nit: is it an executable? or just a binary?

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,211 @@
...
+        Network Server->>Facade Function: Retrieve firmware upgrade file
+        Facade Function->>Storage account: Download firmware file
+        Network Server->>Concentrator: CUPS response with firmware upgrade fields populated
+      end
+    end
+```
+
+### IoT Hub related changes
+
+Only change is related to the concentrator device twin.
```
p-schuler, line 103:
```suggestion
The only change is related to the concentrator device twin.
```

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,211 @@
...
+`"{stationEui}-{package}"` (without any extension, as anyways these are going to
+be downloaded as `update.bin` from the Basics Station executable).
+
+### Azure Function related changes
+
+A new endpoint will be added in the Facade Azure Function which will be used to
+fetch firmware upgrade files from the storage account. The endpoint will accept
+the `StationEui` as input, then retrieve the concentrator twin from IoT Hub,
+download the firmware file from the storage account and send it back in the
+response.
```
p-schuler, line 141: this should be a stream not a download and send. Just making sure we are on the same page. We do not want to put the storage content in memory.

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,211 @@
...
+
+1. The Basics Station sends the CUPS request containing the currently used
+   version (in the `package` field).
+
+1. The LNS compares the values from the device twin and the CUPS request, and
+   determines if an upgrade is required. If that's the case, the LNS will
+   download the upgrade file from storage and send a properly populated response
+   to the Basics Station.
+
+1. The Basics Station will then execute the actual firmware upgrade.
```
p-schuler, line 76: can we consider adding the reported package version? that will allow us to determine, when the concentrator actually updated.

p-schuler (approved):
The reported property update of the package version should be done. The rest LGTM.

docs/adr/007_cups_firmware_upgrade.md
danigian: it must be an executable, as per cups protocol specification.

docs/adr/007_cups_firmware_upgrade.md
```diff
@@ -0,0 +1,211 @@
...
+- **'package'**: desired package version of the Station (matching what will be
+  extracted as version.txt file during update)
+- **'fwUrl'**: URL pointing to the storage location of the file required to run
+  the upgrade
+- **'fwKeyChecksum'**: checksum of the key used to sign the digest of the
+  firmware upgrade file
+- **'fwSignature'**: signature of the uploaded firmware upgrade file (as base64
+  encoded string)
+
+### Storage related changes
```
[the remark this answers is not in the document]
danigian: Absolutely, the function is authenticating to the storage with a connection string, already being used for function runtime itself
@MaggieSalak can you please update this?

Issue #1209, opened 2022-01-10
`LoRaDeviceClient.SetRetry` behavior could be inconsistent when multiple requests are arriving


danigian:
LoRaDeviceClient.SetRetry could set an inconsistent retry policy when concurrent requests are incoming, due to the "singleton" nature of the DeviceClient itself.

While we are processing a request, we employ a single retry policy, since share the DeviceClient instance. Since we do not control the parallelism of the requests, this means that another request could change the retry policy while another is in progress. The outcome is undefined.

Looking at the code, all exposed operations do set the retry to true and reset in the finally block to false. We should simply enable retries in the ctor and keep the setting without tampering with it.

danigian:
@p-schuler please feel free to add details here

PR #1217, opened 2022-01-10
Configure device client retry at initialisation-time

atifaziz:
# PR for issue #1209

## What is being addressed

The issue described in bug #1209.

## How is this addressed

The retry policy is set to the exponential back-off policy at initialisation-time instead of per-operation.


danigian (commented):
LGTM, but would also like to double check with @p-schuler if he sees something I am not considering

p-schuler (approved):
LGTM

PR #1151, opened 2021-12-23
ADR on deduplication strategies

spygi:
# PR for issue #946 

Join work with @p-schuler 

## What is being addressed

Documents Deduplication strategies used in the starterkit. 
- Deduplication on the Network server level.
- Deduplication on the Function level.

PR #1218, opened 2022-01-10
Simplify the deduplication cache key by using the unique DevEui

spygi:
# PR for issue #1152

## What is being addressed

Previously, for the cache key we were relying among others on the `RawMessage` field that was empty for all messages. This meant potentially more cache hits (if the non-unique `DevAddr`, frame counter and `Mic` were also the same) and potentially more dropped messages.

## How is this addressed

By using the unique `DevEui`, we don't need to rely on the raw message anymore. Tests were adjusted to vary the `DevEui` instead of the payload.

Since the keys now have a constant length (no variable component), we are also removing the SHA256 hashing to not have extra computational overhead.

In order to ensure that there are no key collisions in the MemoryCache we are adding also a unique prefix for the different message types with an explicit test for it.

## TODO
If #1151 gets merged before this one, update accordingly.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
@@ -6,7 +6,6 @@
 namespace LoRaWan.NetworkServer
 {
     using System;
-    using System.Buffers.Binary;
     using LoRaTools.LoRaMessage;
     using Microsoft.Extensions.Caching.Memory;
     using Microsoft.Extensions.Logging;
@@ -14,12 +13,14 @@ namespace LoRaWan.NetworkServer
     public sealed class ConcentratorDeduplication : IConcentratorDeduplication
     {
         private static readonly TimeSpan DefaultExpiration = TimeSpan.FromMinutes(1);
-        internal const string DataMessageCacheKeyPrefix = "datamessage:";
-        internal const string JoinMessageCacheKeyPrefix = "joinmessage:";
 
         private readonly IMemoryCache cache;
         private readonly ILogger<IConcentratorDeduplication> logger;
         private static readonly object cacheLock = new object();
+
+        internal sealed record DataMessageKey(DevEui DevEui, Mic Mic, ushort FCnt);
+
+        internal sealed record JoinMessageKey(JoinEui JoinEui, DevEui DevEui, DevNonce DevNonce);
 
         public ConcentratorDeduplication(
             IMemoryCache cache,
@@ -66,7 +67,7 @@ namespace LoRaWan.NetworkServer
             return result;
         }
 
-        private bool EnsureFirstMessageInCache(string key, LoRaRequest loRaRequest, out StationEui previousStation)
+        private bool EnsureFirstMessageInCache(object key, LoRaRequest loRaRequest, out StationEui previousStation)
         {
             var stationEui = loRaRequest.StationEui;
 
@@ -85,41 +86,12 @@ namespace LoRaWan.NetworkServer
             return false;
         }
 
-        internal static string CreateCacheKey(LoRaPayloadData payload, LoRaDevice loRaDevice)
-        {
-            var someMic = payload.Mic ?? throw new ArgumentException(nameof(payload.Mic));
+        internal static DataMessageKey CreateCacheKey(LoRaPayloadData payload, LoRaDevice loRaDevice) =>
+            payload.Mic is { } someMic
+                ? new DataMessageKey(DevEui.Parse(loRaDevice.DevEUI), someMic, payload.GetFcnt())
+                : throw new ArgumentException(nameof(payload.Mic));
 
-            var totalBufferLength = DevEui.Size + Mic.Size + payload.Fcnt.Length;
-            Span<byte> buffer = stackalloc byte[totalBufferLength];
-            var head = buffer; // keeps a view pointing at the start of the buffer
-
-            buffer = DevEui.Parse(loRaDevice.DevEUI).Write(buffer);
-            buffer = someMic.Write(buffer);
-            BinaryPrimitives.WriteUInt16LittleEndian(buffer, BinaryPrimitives.ReadUInt16LittleEndian(payload.Fcnt.Span));
-
-            return CreateCacheKeyCore(DataMessageCacheKeyPrefix, head);
-        }
-
-        internal static string CreateCacheKey(LoRaPayloadJoinRequest payload)
-        {
-            var totalBufferLength = JoinEui.Size + DevEui.Size + DevNonce.Size;
-            Span<byte> buffer = stackalloc byte[totalBufferLength];
-            var head = buffer; // keeps a view pointing at the start of the buffer
-
-            buffer = JoinEui.Read(payload.AppEUI.Span).Write(buffer);
-            buffer = DevEui.Read(payload.DevEUI.Span).Write(buffer);
-            _ = payload.DevNonce.Write(buffer);
-
-            return CreateCacheKeyCore(JoinMessageCacheKeyPrefix, head);
-        }
-
-        private static string CreateCacheKeyCore(string prefix, ReadOnlySpan<byte> buffer)
-        {
-            var bufferToHexLength = (buffer.Length * 3) - 1;
-            Span<char> hexBuffer = bufferToHexLength <= 128 ? stackalloc char[bufferToHexLength] : new char[bufferToHexLength]; // uses the stack for small allocations, otherwise the heap
-            Hexadecimal.Write(buffer, hexBuffer, separator: '-');
-
-            return string.Concat(prefix, hexBuffer.ToString());
-        }
+        internal static JoinMessageKey CreateCacheKey(LoRaPayloadJoinRequest payload) =>
+            new JoinMessageKey(JoinEui.Read(payload.AppEUI.Span), DevEui.Read(payload.DevEUI.Span), payload.DevNonce);
     }
 }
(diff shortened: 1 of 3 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -88,46 +85,37 @@ private bool EnsureFirstMessageInCache(string key, LoRaRequest loRaRequest, out
...
-            _ = devNonce.Write(buffer);
-
-            var key = Sha256.ComputeHash(head.ToArray());
+            return CreateCacheKeyCore(JoinMessageCacheKeyPrefix, head);
+        }
 
-            return BitConverter.ToString(key);
+        private static string CreateCacheKeyCore(string prefix, ReadOnlySpan<byte> buffer)
+        {
+            Span<char> key = stackalloc char[(buffer.Length * 3) - 1];
```
p-schuler, line 116: I think you are making an assumption here about the length. Since we are using stackalloc it would be good to validate this to ensure we do not start using a larger buffer.

spygi: Agree, changed in f1cb1a1e7011f4e140151315f8a601d7a873c295

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -88,46 +85,39 @@ private bool EnsureFirstMessageInCache(string key, LoRaRequest loRaRequest, out
...
-            payload.RawMessage?.CopyTo(buffer[index..]);
-            index += payload.RawMessage?.Length ?? 0;
-
-            BinaryPrimitives.WriteUInt16LittleEndian(buffer[index..], BinaryPrimitives.ReadUInt16LittleEndian(payload.Fcnt.Span));
+            var totalBufferLength = DevEui.Size + Mic.Size + payload.Fcnt.Length;
+            Span<byte> buffer = stackalloc byte[totalBufferLength];
+            var head = buffer; // keeps a view pointing at the start of the buffer
 
-            var key = Sha256.ComputeHash(buffer.ToArray());
+            buffer = DevEui.Parse(loRaDevice.DevEUI.AsSpan()).Write(buffer);
```
atifaziz, line 94: Conversion to Span<char> is not needed as it's implicit:

```suggestion
            buffer = DevEui.Parse(loRaDevice.DevEUI).Write(buffer);
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
```diff
@@ -88,46 +85,39 @@ private bool EnsureFirstMessageInCache(string key, LoRaRequest loRaRequest, out
...
+            buffer = payload.Mic is { } someMic ? someMic.Write(buffer) : throw new InvalidOperationException("Mic must not be null.");
```
atifaziz, line 95: The fact that the payload must have a MIC should be valid at the beginning of the method and `payload` identified as an invalid argument, as shown below:

```diff
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
index 2e4a08cf..4e155ec0 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ConcentratorDeduplication.cs
@@ -87,13 +86,14 @@ namespace LoRaWan.NetworkServer
 
         internal static string CreateCacheKey(LoRaPayloadData payload, LoRaDevice loRaDevice)
         {
+            var mic = payload.Mic ?? throw new ArgumentException("Payload MIC must be defined.");
             var totalBufferLength = DevEui.Size + Mic.Size + payload.Fcnt.Length;
             Span<byte> buffer = stackalloc byte[totalBufferLength];
             var head = buffer; // keeps a view pointing at the start of the buffer
 
             buffer = DevEui.Parse(loRaDevice.DevEUI.AsSpan()).Write(buffer);
-            buffer = payload.Mic is { } someMic ? someMic.Write(buffer) : throw new InvalidOperationException("Mic must not be null.");
+            buffer = mic.Write(buffer);
             BinaryPrimitives.WriteUInt16LittleEndian(buffer, BinaryPrimitives.ReadUInt16LittleEndian(payload.Fcnt.Span));
 
             return CreateCacheKeyCore(DataMessageCacheKeyPrefix, head);
         }
```

spygi: Fixed in a7594688cac18a80e93e6dd92dbd0d01db0788e7

atifaziz (commented):
I would like to recommend use of records for keys as opposed to plain strings, which I feel would simplify a whole host of things and boil down to simply this:

```c#
private sealed record DataMessageKey(DevEui DevEui, Mic Mic, ushort FCnt);
private sealed record JoinMessageKey(JoinEui JoinEui, DevEui DevEui, DevNonce DevNonce);

internal static object CreateCacheKey(LoRaPayloadData payload, LoRaDevice loRaDevice) =>
    payload.Mic is { } someMic
    ? new DataMessageKey(DevEui.Parse(loRaDevice.DevEUI), someMic, payload.GetFcnt())
    : throw new ArgumentException(nameof(payload.Mic));

internal static object CreateCacheKey(LoRaPayloadJoinRequest payload) =>
    new JoinMessageKey(JoinEui.Read(payload.AppEUI.Span), DevEui.Read(payload.DevEUI.Span), payload.DevNonce);
```

All the string manipulation, encoding, buffers, bytes and prefixes are not needed with this approach.

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
```diff
@@ -91,29 +91,32 @@ public void When_Data_Message_Encountered_Should_Find_Duplicates_For_Different_D
             // act/assert
             Assert.Equal(expectedResult, this.concentratorDeduplication.CheckDuplicateData(this.dataRequest, this.loRaDevice));
             Assert.Equal(1, this.cache.Count);
-            var key = ConcentratorDeduplication.CreateCacheKey(this.dataPayload);
+            var key = ConcentratorDeduplication.CreateCacheKey(this.dataPayload, this.loRaDevice);
             Assert.True(this.cache.TryGetValue(key, out var foundStation));
             Assert.Equal(station1Eui, foundStation);
         }
 
+        public static IEnumerable<object[]> CreateKeyDataMessagesTheoryData =>
```
atifaziz, line 99: I think if you're going to go with theory data this way, you might as well make it strong-typed all the way:

```c#
internal static TheoryData<DevEui, Mic, ushort, string?> CreateKeyDataMessagesTheoryData =>
    TheoryDataFactory.From((new DevEui(0), new Mic(0), (ushort)0, null),
                           (new DevEui(0), new Mic(0), (ushort)0, "1"), // a non-relevant field should not influence the key
                           (new DevEui(0x1010101010101010UL), new Mic(0), (ushort)0, null),
                           (new DevEui(0), new Mic(1), (ushort)0, null),
                           (new DevEui(0), new Mic(0), (ushort)1, null));

[Theory]
[MemberData(nameof(CreateKeyDataMessagesTheoryData))]
internal void CreateKeyMethod_Should_Return_Expected_Keys_For_Different_Data_Messages(DevEui devEui, Mic mic, ushort frameCounter, string? fieldNotUsedInKey = null)
{
    var options = fieldNotUsedInKey ?? string.Empty;
    using var testDevice = new LoRaDevice(this.simulatedABPDevice.DevAddr, devEui.ToString(), this.connectionManager);

    var payload = new LoRaPayloadDataLns(this.dataPayload.DevAddr, new MacHeader(MacMessageType.ConfirmedDataUp),
                                         frameCounter, options, "payload", mic);

    var expectedKey = new ConcentratorDeduplication.DataMessageKey(devEui, mic, frameCounter);
    Assert.Equal(expectedKey, ConcentratorDeduplication.CreateCacheKey(payload, testDevice));
}
```

However, it seems odd to me that by exposing the key records (which I didn't in my original suggestion), you see that the last line is essentially _testing the implementation_ by re-creating it in the test code.

atifaziz (approved):

Tests/Unit/NetworkServer/ConcentratorDeduplicationTest.cs
spygi: There is 1 issue and 1 style-preference I would do differently from the committed sample: 
- Issue: MemberData need to be public otherwise we get a runtime exception when running the test
- I would rather we have all the test data as parameterized data and not only the inputs. 

Instead and in order to go more type-safe, see 656944045ea91c722afb20031f7e1915691a18fc
	- I added an overload in the `TheoryDataFactory` that takes 5 parameters 
	- When the `MemberData` is public, then the record type also needs to be public as well (otherwise CS0053 inconsistent accessibility) This would make the type globally visible which I think we should not do only for a test. Instead I am using object just for this parameter and still have type safety for the rest of the parameters + an implicit conversion at runtime for the record type. I think this is the best way in order not to make the record public and provide more type-safety.

PR #1220, opened 2022-01-11
Reuse `StationEui` and `JoinEui`

bastbu:
# PR for issue #964

## What is being addressed

Reuses `StationEui` and `JoinEui`.

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
@@ -4,6 +4,7 @@
 namespace LoraKeysManagerFacade
 {
     using System;
+    using System.Globalization;
     using System.IO;
     using System.Security.Cryptography;
     using System.Threading;
@@ -12,6 +13,7 @@ namespace LoraKeysManagerFacade
     using Azure.Storage.Blobs.Models;
     using LoRaTools.CommonAPI;
     using LoRaTools.Utils;
+    using LoRaWan;
     using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Mvc;
     using Microsoft.Azure.Devices;
@@ -62,12 +64,13 @@ namespace LoraKeysManagerFacade
 
         internal async Task<IActionResult> RunFetchConcentratorCredentials(HttpRequest req, CancellationToken cancellationToken)
         {
-            var stationEui = req.Query["StationEui"];
-            if (StringValues.IsNullOrEmpty(stationEui))
+            var rawStationEui = req.Query["StationEui"];
+            if (StringValues.IsNullOrEmpty(rawStationEui))
             {
                 this.logger.LogError("StationEui missing in request");
                 return new BadRequestObjectResult("StationEui missing in request");
             }
+            var stationEui = StationEui.Parse((string)rawStationEui);
 
             var credentialTypeQueryString = req.Query["CredentialType"];
             if (StringValues.IsNullOrEmpty(credentialTypeQueryString))
@@ -81,10 +84,10 @@ namespace LoraKeysManagerFacade
                 return new BadRequestObjectResult($"Could not parse desired concentrator credential type '{credentialTypeQueryString}'.");
             }
 
-            var twin = await this.registryManager.GetTwinAsync(stationEui, cancellationToken);
+            var twin = await this.registryManager.GetTwinAsync(stationEui.ToString("N", CultureInfo.InvariantCulture), cancellationToken);
             if (twin != null)
             {
-                this.logger.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui.ToString());
+                this.logger.LogInformation("Retrieving '{CredentialType}' for '{StationEui}'.", credentialType.ToString(), stationEui);
                 try
                 {
                     const string cupsKey = "cups";
--- a/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
@@ -64,13 +64,11 @@ namespace LoraKeysManagerFacade
 
         internal async Task<IActionResult> RunFetchConcentratorCredentials(HttpRequest req, CancellationToken cancellationToken)
         {
-            var rawStationEui = req.Query["StationEui"];
-            if (StringValues.IsNullOrEmpty(rawStationEui))
+            if (!StationEui.TryParse((string)req.Query["StationEui"], out var stationEui))
             {
-                this.logger.LogError("StationEui missing in request");
-                return new BadRequestObjectResult("StationEui missing in request");
+                this.logger.LogError("StationEui missing in request or invalid");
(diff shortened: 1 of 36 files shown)


Tests/Simulation/IntegrationTestFixtureSim.cs
```diff
@@ -53,7 +53,7 @@ public override void SetupTestDevices()
             Device1002_Simulated_OTAA = new TestDeviceInfo()
             {
                 DeviceID = "0000000000001002",
-                AppEUI = "0000000000001002",
+                AppEUI = GetJoinEui(1002),
```
atifaziz, line 56: `0000000000001002` was in hex so shouldn't this read `GetJoinEui(0x1002)`?

Tests/Simulation/IntegrationTestFixtureSim.cs
```diff
@@ -64,7 +64,7 @@ public override void SetupTestDevices()
             Device1003_Simulated_HttpBasedDecoder = new TestDeviceInfo
             {
                 DeviceID = "0000000000001003",
-                AppEUI = "0000000000001003",
+                AppEUI = GetJoinEui(1003),
```
atifaziz, line 67: `0000000000001003` was in hex so shouldn't this read `GetJoinEui(0x1003)`?

Tests/Simulation/IntegrationTestFixtureSim.cs
bastbu: Yes, it is not the same value before/after. I tried to be consistent with some of the other helper methods, where in the past we went away from passing hex numbers to just passing integers in decimal representation. Based on our discussion offline, we should probably revisit how we create these devices entirely.

Tests/E2E/OTAAJoinTest.cs
```diff
@@ -142,7 +142,7 @@ public async Task OTAA_Join_With_Wrong_AppEUI_Fails()
             var device = TestFixtureCi.Device13_OTAA;
             LogTestStart(device);
 
-            var appEUIToUse = "FF7A00000000FCE3";
+            var appEUIToUse = JoinEui.Parse("FF7A00000000FCE3");
```
atifaziz, line 145: nit:

```suggestion
            var appEUIToUse = new JoinEui(0xFF7A00000000FCE3);
```

LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
```diff
@@ -62,12 +64,13 @@ public class ConcentratorCredentialsFunction
...
         {
-            var stationEui = req.Query["StationEui"];
-            if (StringValues.IsNullOrEmpty(stationEui))
+            var rawStationEui = req.Query["StationEui"];
+            if (StringValues.IsNullOrEmpty(rawStationEui))
             {
                 this.logger.LogError("StationEui missing in request");
                 return new BadRequestObjectResult("StationEui missing in request");
             }
+            var stationEui = StationEui.Parse((string)rawStationEui);
```
atifaziz, line 73: How about just going with `StationEui.TryParse` here?

```suggestion
            if (!StationEui.TryParse((string)req.Query["StationEui"], out var stationEui))
            {
                this.logger.LogError("StationEui missing in request or invalid");
                return new BadRequestObjectResult("StationEui missing in request or invalid");
            }
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -48,7 +48,7 @@ public class LoRaDevice : IDisposable, ILoRaDeviceRequestQueue
 
         public AppKey? AppKey { get; set; }
 
-        public string AppEUI { get; set; }
+        public JoinEui? AppEUI { get; set; }
```
atifaziz, line 51: Should we use this as an opportunity to rename `AppEUI` to the more conventional `AppEui`?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinRequest.cs
```diff
@@ -15,7 +15,7 @@ public class LoRaPayloadJoinRequest : LoRaPayload
         /// <summary>
         /// Gets or sets aka JoinEUI.
         /// </summary>
-        public Memory<byte> AppEUI { get; set; }
+        public JoinEui AppEUI { get; set; }
```
atifaziz, line 18: Should we use this as an opportunity to rename `AppEUI` to the more conventional `AppEui`?

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
```diff
@@ -107,11 +107,12 @@ public async Task After_Saving_Frame_Counter_Changes_Should_Not_Have_Pending_Cha
         public async Task When_Initialized_New_OTAA_Device_Should_Have_All_Properties()
         {
             var appKey = TestKeys.CreateAppKey(0xABC0200000000000, 0x09);
+            var joinEui = new JoinEui(0xABC0200000000009);
 
             var twin = TestUtils.CreateTwin(
                 desired: new Dictionary<string, object>
                 {
-                    { "AppEUI", "ABC0200000000009" },
+                    { "AppEUI", joinEui.ToString() },
```
atifaziz, line 115: Shouldn't this be using the `N` format to mean the same thing?

atifaziz (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
bastbu: Done in 6339ab6c5009e7292142037060b8e8c60128f308

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinRequest.cs
bastbu: Done in 6339ab6c5009e7292142037060b8e8c60128f308

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
bastbu: Valid point, done in fd8bc7daed7f63f42291c59ed33ae9bcb17917e3

Tests/Simulation/IntegrationTestFixtureSim.cs
bastbu: Resolving this as per other comment

Tests/Simulation/IntegrationTestFixtureSim.cs
bastbu: Update: This comment caught a valid bug. I thought that devices were recreated on E2E test runs if the properties do not match (for which we have logic, which is apparently not executed on each E2E run). I needed to revert to `JoinEui.Parse("...0027")` for the E2E tests to make sure that the EUIs still match the values of the twin.

PR #1222, opened 2022-01-11
Encapsulate session key type into key calculation methods of `OTAAKeysGenerator`

atifaziz:
This PR encapsulates the session key type into the distinct key calculation methods `CalculateNetworkSessionKey` and `CalculateAppSessionKey` as opposed to it being an argument of those methods.

---
PS It also adds unit tests missing for `OTAAKeysGenerator` to ensure no regression.


p-schuler (approved):
LGTM

Issue #1166, opened 2022-01-03
Improve client certificate validation


danigian:
Errors coming from invalid client certificates are propagated in X509ChainStatusFlags and/or SslPolicyErrors that we can handle in ClientCertificateValidatorService.ValidateAsync.

As of now, we are only logging potential issues, without taking any concrete action on it and we are validating certificates only depending on their thumbprint.

We should increase the security of validation process at least by passing and trusting the root certificate (or full chain of trust) in LoRaWan Network Server Docker image.

- [x] #1238
- [x] #1239

PR #1226, opened 2022-01-11
Improve client certificate validation by validating X509ChainStatus

danigian:
# PR for issue #1166

## What is being addressed

Before this PR, we were only logging potential X509ChainStatus issues, without taking any concrete action on those and we were validating certificates only depending on their thumbprint.

With this PR, we are increasing the security of validation process by requiring users to pass and trust the certificate chain in LoRaWan Network Server image

## How is this addressed

- [x] ClientCertificateValidatorService has been adjusted to both log and return false when checking for X509ChainStatus flags
- [x] A start.sh is being introduced for LoRaWan Network Server images. This script is currently conditionally copying a CLIENT_CA_PATH or a default /var/lorastarterkit/certs/client.ca.crt in the /usr/local/share/ca-certificates path for later using the update-ca-certificates command and then execute the .NET entrypoint
- [x] LoRaWan Network Server Dockerfiles have been adjusted to copy and run the start.sh script
- [x] CI has been adjusted to copy the proper file to "AIO" system "DEFAULT_CLIENT_CA_PATH" folder


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
```diff
@@ -43,13 +43,18 @@ public async Task<bool> ValidateAsync(X509Certificate2 certificate, X509Chain? c
 
             using var scope = this.logger.BeginEuiScope(stationEui);
 
-            // Logging any chain related issue, but not failing on it.
-            foreach (var status in chain.ChainStatus)
+            // Logging any chain related issue that is causing verification to fail
+            if (chain.ChainStatus.Any())
```
Mandur, line 47: Thisd part is not clear to me, I understood it reprensents the certificate chain of the submitted certificate?  Why would the presence of chain status be a sign of error, couldn't that be also correct case? https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509chainstatus?view=net-6.0

Mandur: Would be nice if we could have a chat

LoRaEngine/modules/LoRaWanNetworkSrvModule/start.sh
```diff
@@ -0,0 +1,38 @@
...
+    fi
+fi
+
+if [[ ! -z $certificateCopied ]] && [[ $certificateCopied -eq 0 ]]; then
+    update-ca-certificates --fresh
+else
+    echo "No certifiate copied, therefore no update-ca-certificates needed."
+fi
+
+dotnet LoRaWanNetworkSrvModule.dll
```
Mandur, line 38: I think you need a newline at the end of the file :D

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ClientCertificateValidatorService.cs
danigian: After the offline discussion, I added a condition on the check that is verifying if there is any chainstatus flag that is not corresponding to the "NoError" one.

Issue #1225, opened 2022-01-11
Add `AppNonce`


atifaziz:
Add a strong-typed primitive to represent a 24-bit **AppNonce**.

PR #1227, opened 2022-01-11
Add type for `AppNonce`

atifaziz:
# PR for issue #1225

## What is being addressed

- Lack of encapsulation of details regarding **AppNonce** from the LoRaWAN specification
- Use of byte arrays and memory to represent the nonce

## How is this addressed

A new type `AppNonce` has been added to encapsulate the details and used for properties in composite types and method signatures.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Mic.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Mic.cs
@@ -78,14 +78,14 @@ namespace LoRaWan
             return new Mic(BinaryPrimitives.ReadUInt32LittleEndian(cmac));
         }
 
-        public static Mic ComputeForJoinAccept(AppKey appKey, MacHeader macHeader, Memory<byte> joinNonce, Memory<byte> netId, DevAddr devAddr, Memory<byte> dlSettings, Memory<byte> rxDelay, Memory<byte> cfList)
+        public static Mic ComputeForJoinAccept(AppKey appKey, MacHeader macHeader, AppNonce joinNonce, Memory<byte> netId, DevAddr devAddr, Memory<byte> dlSettings, Memory<byte> rxDelay, Memory<byte> cfList)
         {
-            var algoInput = new byte[MacHeader.Size + joinNonce.Length + NetId.Size + DevAddr.Size + dlSettings.Length + rxDelay.Length + cfList.Length];
+            var algoInput = new byte[MacHeader.Size + AppNonce.Size + NetId.Size + DevAddr.Size + dlSettings.Length + rxDelay.Length + cfList.Length];
             var index = 0;
             _ = macHeader.Write(algoInput.AsSpan(index));
             index += MacHeader.Size;
-            joinNonce.CopyTo(algoInput.AsMemory(index));
-            index += joinNonce.Length;
+            _ = joinNonce.Write(algoInput.AsSpan(index));
+            index += AppNonce.Size;
             netId.CopyTo(algoInput.AsMemory(index));
             index += netId.Length;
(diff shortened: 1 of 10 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Mic.cs
```diff
@@ -78,14 +78,14 @@ private static Mic ComputeForJoinRequest(byte[] keyBytes, MacHeader mhdr, JoinEu
...
+        public static Mic ComputeForJoinAccept(AppKey appKey, MacHeader macHeader, AppNonce joinNonce, Memory<byte> netId, DevAddr devAddr, Memory<byte> dlSettings, Memory<byte> rxDelay, Memory<byte> cfList)
         {
-            var algoInput = new byte[MacHeader.Size + joinNonce.Length + NetId.Size + DevAddr.Size + dlSettings.Length + rxDelay.Length + cfList.Length];
+            var algoInput = new byte[MacHeader.Size + AppNonce.Size + NetId.Size + DevAddr.Size + dlSettings.Length + rxDelay.Length + cfList.Length];
             var index = 0;
             _ = macHeader.Write(algoInput.AsSpan(index));
             index += MacHeader.Size;
-            joinNonce.CopyTo(algoInput.AsMemory(index));
-            index += joinNonce.Length;
+            _ = joinNonce.Write(algoInput.AsSpan(index));
```
bastbu, line 87: Don't we have the order of the bytes reversed compared to what we had previously here (since it's a random value it probably won't matter)?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
```diff
@@ -167,8 +163,11 @@ public override byte[] PerformEncryption(AppKey key)
             var micBytes = new byte[4];
             _ = Mic is { } someMic ? someMic.Write(micBytes) : throw new InvalidOperationException("MIC must not be null.");
 
+            Span<byte> appNonce = stackalloc byte[AppNonce.Size];
```
bastbu, line 166: [nit] why not write it to an array directly and avoid the `appNonce.ToArray()` on L170?

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
atifaziz: Was more work but done with 6b3d9fa2837707a50297f6d54d7cb44161a665a0.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/Mic.cs
atifaziz: Not really. Since we have a strong type that reads it correctly (in little endian) from and to buffers. The reversals [were](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1227/files#diff-2e7394e990e11eedfe539de7db67c6c0ca0423cc09dea7714dfb864c6febbc80L92) [removed](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1227/files#diff-2e7394e990e11eedfe539de7db67c6c0ca0423cc09dea7714dfb864c6febbc80L142). While it's a random value, the `MessageProcessorJoinTest.When_Device_Is_Found_In_Api_Should_Update_Twin_And_Return` test does fail on line if the byte order is incorrect.

PR #1228, opened 2022-01-11
Reuse `DevEui`

bastbu:
# PR for issue #964

## What is being addressed

Use `DevEui` in Facade project. References that extend via the `LoRaTools` project into other projects transiently, we resolve by introducing an intermediate parsing step for the `DevEui` based on every usage. These will be removed once we reuse `DevEui` from the rest of the solution.


LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs
```diff
@@ -107,16 +108,16 @@ public async Task<List<IoTHubDeviceInfo>> GetDeviceList(string devEUI, string ga
...
                     else
                     {
-                        log?.LogWarning("Failed to acquire lock for '{key}'", devEUI);
+                        log?.LogWarning("Failed to acquire lock for '{key}'", someDevEui);
                     }
                 }
                 else
                 {
-                    log?.LogDebug("dev nonce already used. Ignore request '{key}':{gwid}", devEUI, gatewayId);
+                    log?.LogDebug("dev nonce already used. Ignore request '{key}':{gwid}", someDevEui, gatewayId);
```
p-schuler, line 120: The default formatting for the ToString on the Eui is using '-' as separators. That's a change from the previous formatting, which was not using any separators. We should consider keeping the default to not change that. One thing that is easier if we do not have the separators is copying around the value.

LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
```diff
@@ -131,9 +133,9 @@ private async Task<LoRaADRTable> GetADRTableCore(string key)
             return null;
         }
 
-        private static string GetEntryKey(string devEUI)
+        private static string GetEntryKey(DevEui devEUI)
         {
-            return devEUI + CacheToken;
+            return devEUI.ToString("N", CultureInfo.InvariantCulture) + CacheToken;
```
p-schuler, line 138: wouldn't it make sense to change the default to be "N" for the Eui?

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
```diff
@@ -277,7 +279,7 @@ private void BulkSaveDevAddrCache(List<DevAddrCacheInfo> devAddrCacheInfos, bool
         private static IDictionary<string, DevAddrCacheInfo> KeepExistingCacheInformation(HashEntry[] cacheDevEUIEntry, IGrouping<DevAddr, DevAddrCacheInfo> newDevEUIList, bool canDeleteExistingDevice)
         {
             // if the new value are not different we want to ensure we don't save, to not update the TTL of the item.
-            var toSyncValues = newDevEUIList.ToDictionary(x => x.DevEUI);
+            var toSyncValues = newDevEUIList.ToDictionary(x => GenerateKey(x.DevEUI.Value));
```
p-schuler, line 282: why is this additional GenerateKey necessarey?

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
```diff
@@ -14,20 +16,15 @@ public sealed class LoRaDeviceCache : IDisposable
...
         private readonly string cacheKey;
 
         public bool IsLockOwner { get; private set; }
 
         private string lockKey;
 
-        public LoRaDeviceCache(ILoRaDeviceCacheStore cacheStore, string devEUI, string gatewayId)
+        public LoRaDeviceCache(ILoRaDeviceCacheStore cacheStore, DevEui devEUI, string gatewayId)
         {
-            if (string.IsNullOrEmpty(devEUI))
```
p-schuler, line 26: I don't think an "empty" devEUI would be allowed. Is this a case we should validate? Do we have a representation of "empty" (all 0)?

LoRaEngine/LoraKeysManagerFacade/SearchDeviceByDevEUI.cs
```diff
@@ -43,29 +45,30 @@ public async Task<IActionResult> GetDeviceByDevEUI([HttpTrigger(AuthorizationLev
 
         private async Task<IActionResult> RunGetDeviceByDevEUI(HttpRequest req, ILogger log)
         {
-            var devEUI = req.Query["DevEUI"];
-            if (StringValues.IsNullOrEmpty(devEUI))
+            var rawDevEui = req.Query["DevEUI"];
+            if (StringValues.IsNullOrEmpty(rawDevEui))
```
p-schuler, line 49: why not start using TryParse on the Eui for all the parameter conversions? We could also get rid of the EUIValidator, no?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/CommonAPI/LoRaCloudToDeviceMessage.cs
```diff
@@ -5,13 +5,22 @@ namespace LoRaTools.CommonAPI
...
     {
-        public string DevEUI { get; set; }
+        [JsonIgnore]
+        public DevEui? DevEUI { get; set; }
+
+        [JsonProperty("DevEUI")]
+        public string DevEuiString
+        {
+            get => DevEUI?.ToString("N", null);
+            set => DevEUI = string.IsNullOrEmpty(value) ? null : DevEui.Parse(value);
```
p-schuler, line 22: Some of the serialization does plain .Parse, others do use TryParse and others do some minor validation (like this one). Some of them may allow null and that's fine (would be good to check again), but here, it seems we should be using TryParse as well?

p-schuler (changes requested):
I think we need to discuss the default for the ToString formatting on the Eui. Right now it's inconsistent. Some of the log output will be writing using '-' as separators, others will use the existing format ('N'). I see DevEUI.ToString("N", CultureInfo.InvariantCulture) all over the place, which suggest, we should make this the default format, remove the explicit "N" cases and get consistency with the default.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/CommonAPI/LoRaCloudToDeviceMessage.cs
bastbu: I think there's a minor difference here that might be useful - if the `value` is not `null` or empty, we enforce that it is a parseable `DevEui`. In case of something invalid (e.g. `"1234"`) we fail faster compared to just using `TryParse`

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
bastbu: Currently, the underlying `ILoRaDeviceCacheStore` only supports keys of type `string`. We could make the underlying support generic keys, but I would propose to do maybe do that in a separate PR to not increase the amount of diffs further.

LoRaEngine/LoraKeysManagerFacade/SearchDeviceByDevEUI.cs
bastbu: We cannot get rid of the `EuiValidator` entirely since it also protects against `default`, but I refactored to get rid of the duplication in 2fdccb71d6985d01c8d81c1a696958524477f81f

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
bastbu: Since `DevEui` is a value type we could use `default` for that. In this case it is already validated higher up in the stack (at the beginning of each function that calls this code).

LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs
bastbu: Tagging @atifaziz for his input here. It's related to some of your other comments, let's discuss it here.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
atifaziz: > but I would propose to do maybe do that in a separate PR to not increase the amount of diffs further.

I think it's a little too late for that now. 😉.

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
atifaziz: According to the [DevEUI documentation on the LoRa Developer Portal][deveui-doc]:

> Note: Values of all ones or all zeros ( 0xFFFFFFFFFFFF or 0x0) are not valid EUI-64 values and should indicate that a value for the EUI-64 is not available (null). Do not use these values for the DevEUI or JoinEUI.
>
> &mdash; <cite>[DevEUI documentation on the LoRa Developer Portal][deveui-doc]</cite>, &copy; Semtech Corp.

So yes, a `default(DevEui)` would be invalid. However, since it's a value type, we should be using `DevEui?` to represent a DevEUI that's missing.

[deveui-doc]: https://lora-developers.semtech.com/documentation/tech-papers-and-guides/the-book/deveui/

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IotHubStationInfo.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.NetworkServer
+{
+    using Newtonsoft.Json;
+
+    public sealed record IotHubStationInfo
```
atifaziz, line 8: If this is going to be mutable, it's best to make it a `class`.

bastbu: My mistake, wanted to make it immutable but made a copy/paste error. Will fix this.

bastbu: Done in 0ac22dfde69a85374c69cd5555aafcb99b04a8f4

LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs
```diff
@@ -4,40 +4,23 @@
...
+    internal static class EuiValidator
     {
-        private const string InvalidZero = "0000000000000000";
-
         /// <summary>
         /// DevEUI are required to be IEEE EUI-64.
         /// </summary>
-        /// <param name="devEUI">devEUI to validate.</param>
+        /// <param name="devEui">devEUI to validate.</param>
         /// <exception cref="ArgumentException">If an invalid devEUI string was passed.</exception>
```
atifaziz, line 15: Remove as no longer true:

```suggestion
```

LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs
```diff
@@ -4,40 +4,23 @@
...
-        internal static void ValidateDevEUI(string devEUI)
+        internal static bool TryParseAndValidate(string devEui, out DevEui result)
```
atifaziz, line 16: Isn't this doing the same as `DevEui.TryParse` and therefore can be entirely replace with it?

LoRaEngine/LoraKeysManagerFacade/IoTHubDeviceInfo.cs
```diff
@@ -23,7 +23,15 @@ public DevAddr DevAddr
...
 
-        public string DevEUI { get; set; }
+        [JsonIgnore]
+        public DevEui? DevEUI { get; set; }
+
+        [JsonProperty("DevEUI")]
+        public string DevEuiString
+        {
+            get => DevEUI?.ToHex();
+            set => DevEUI = string.IsNullOrEmpty(value) ? null : DevEui.Parse(value);
```
atifaziz, line 33: Converting an _empty string_ to `null` will given inconsistent results:

```c#
var info = new IoTHubDeviceInfo();
info.DevEuiString = string.Empty;
var s = info.DevEuiString; // returns null
````

File: LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs (hunk 9 above)
Comment (bastbu) [reply LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs:15]: Not entirely. It also validates that the result is not equal to the `default`, for which we had a test case. Instead of removing the test case and using `TryParse` directly I decided to keep this method. We will fail faster if we assign a `default` `DevEui` somewhere on the Network Server side this way.

File: LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs (hunk 1 above)
Comment (bastbu) [reply LoRaEngine/LoraKeysManagerFacade/DeviceGetter.cs:127]: This was addressed with 6774e906970a25a27d7e163acf4f545376e454d9

File: LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs (hunk 2 above)
Comment (bastbu) [reply LoRaEngine/LoraKeysManagerFacade/LoRaADRFunction/LoRaADRRedisStore.cs]: This was addressed with 6774e906970a25a27d7e163acf4f545376e454d9

File: LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs (hunk 9 above)
Comment (atifaziz) [reply LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs:15]: I see. Perhaps we can then turn this into a one-liner and use `IsValid` from `DevEui` (which also treats `FFFF:FFFF:FFFF:FFFF` as invalid per spec)?

```c#
internal static bool TryParseAndValidate(string devEui, out DevEui result) =>
    DevEui.TryParse(devEui, out result) && result.IsValid;
```

nit: Eventually I think this could be an option on `TryParse` of `DevEui`. The goal here would be to avoid individual floating methods and have related functionality living closer.

LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs
```diff
@@ -4,40 +4,22 @@
 namespace LoraKeysManagerFacade
 {
     using System;
```
atifaziz, line 6: Unused:

```suggestion
```

LoRaEngine/LoraKeysManagerFacade/IoTHubDeviceInfo.cs
bastbu: Since `DevEui` needs to be null in some cases, how would you propose to solve this? Use the `default(DevEui)` as a representation of an empty string?

LoRaEngine/LoraKeysManagerFacade/EUIValidator.cs
bastbu: Addressed this with 7e8d334d92f7136d27b191026084332193fc5632.

LoRaEngine/LoraKeysManagerFacade/IServiceClient.cs
```diff
@@ -13,6 +14,6 @@ public interface IServiceClient
     {
         Task<CloudToDeviceMethodResult> InvokeDeviceMethodAsync(string deviceId, string moduleId, CloudToDeviceMethod cloudToDeviceMethod);
 
-        Task SendAsync(string deviceId, Message message);
+        Task SendAsync(DevEui deviceId, Message message);
```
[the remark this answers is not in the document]
bastbu: Done with cc2c21d295aaaf86ef0afee39da3f411376c4702

LoRaEngine/LoraKeysManagerFacade/IoTHubDeviceInfo.cs
atifaziz: A nullable `DevEui`?

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
p-schuler: Ok. Could we simply use ToString() as the default then?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/CommonAPI/LoRaCloudToDeviceMessage.cs
p-schuler: My comment was more around consistency in handling that assignment in the other types.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
p-schuler: nit comment - we can also keep that. Up to you.

p-schuler (approved):
LGTM - thanks for the changes.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
bastbu: As discussed, removed with 62362a4ab925d921476280de8601414a3c6ee5f5

LoRaEngine/LoraKeysManagerFacade/LoRaDeviceCache.cs
bastbu: As discussed, the empty and `0xFF..` cases are already validated on all paths that call this method (where we parse the query/path parameters of the Azure Function).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/CommonAPI/LoRaCloudToDeviceMessage.cs
bastbu: As discussed offline, will resolve this. There is a related comment which I will leave open for now.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/ADR/LoRaADRManagerBase.cs
```diff
@@ -32,16 +32,15 @@ public virtual async Task StoreADREntryAsync(LoRaADRTableEntry newEntry)
                 return;
             }
 
-            if (string.IsNullOrEmpty(newEntry.DevEUI) ||
-               string.IsNullOrEmpty(newEntry.GatewayId))
+            if (string.IsNullOrEmpty(newEntry.GatewayId))
```
atifaziz, line 35: To be consistent with the previous semantics this should add back the check `!newEntry.DevEUI.IsValid` and the message should read `"Missing Gateway ID or invalid DevEUI"`.

Tests/Common/SearchByDevEuiContract.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Common
+{
+    public static class SearchByDevEuiContract
+    {
+        public static readonly string Eui = new DevEui(0x1b2c3d).ToString();
+        public static readonly string PrimaryKey = "ABCDEFGH1234567890";
+        public static readonly string Response = @"{""DevEUI"":""00000000001B2C3D"",""PrimaryKey"":""QUJDREVGR0gxMjM0NTY3ODkw""}";
```
atifaziz, line 10: nit: use more readable/maintainable JSON:

```suggestion
        public static readonly string Response = JsonUtil.Strictify(@"{ DevEUI: '00000000001B2C3D',
                                                                        PrimaryKey: 'QUJDREVGR0gxMjM0NTY3ODkw' }");
```

Tests/Common/TestEui.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Common
+{
+    using System.Security.Cryptography;
+
+    public static class TestEui
+    {
+        public static DevEui GenerateDevEui() => new DevEui((uint)RandomNumberGenerator.GetInt32(int.MaxValue));
```
atifaziz, line 10: I would start from 1 so it never generates an invalid DevEUI and also remark that only the lower 31 bits are set:

```suggestion
        /// <remarks>
        /// While DevEUI are 64 bits wide, this generates a random DevEUI with only 31 LSB set.
        /// </remarks>
        public static DevEui GenerateDevEui() => new(unchecked((uint)RandomNumberGenerator.GetInt32(int.MaxValue) + 1));
```

Tests/Common/TestUtils.cs
```diff
@@ -28,7 +28,7 @@ public static class TestUtils
             ILoRaDeviceClientConnectionManager connectionManager,
             DefaultLoRaDataRequestHandler requestHandler = null)
         {
-            var result = new LoRaDevice(simulatedDevice.LoRaDevice.DevAddr, simulatedDevice.LoRaDevice.DeviceID, connectionManager)
+            var result = new LoRaDevice(simulatedDevice.LoRaDevice.DevAddr, DevEui.Parse(simulatedDevice.LoRaDevice.DeviceID), connectionManager)
```
atifaziz, line 31: Use `DevEui` already added to `LoRaDevice`:

```suggestion
            var result = new LoRaDevice(simulatedDevice.LoRaDevice.DevAddr, simulatedDevice.LoRaDevice.DevEui, connectionManager)
```

Tests/Integration/DevAddrCacheTest.cs
```diff
@@ -902,5 +903,6 @@ public async Task Full_Update_Perform_Correctly_On_Non_Empty_Cache_And_Keep_Old_
         }
 
         private static DevAddr CreateDevAddr() => new DevAddr((uint)RandomNumberGenerator.GetInt32(int.MaxValue));
+        private static DevEui CreateDevEui() => TestEui.GenerateDevEui();
```
atifaziz, line 906: nit: why not inline this?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultClassCDevicesMessageSender.cs
```diff
@@ -39,7 +39,7 @@ public async Task<bool> SendAsync(IReceivedLoRaCloudToDeviceMessage message, Can
         {
             if (message is null) throw new ArgumentNullException(nameof(message));
 
-            if (string.IsNullOrEmpty(message.DevEUI))
+            if (message.DevEUI == default)
```
atifaziz, line 42:
```suggestion
            if (!message.DevEUI.IsValid)
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -289,7 +289,7 @@ public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaReques
 
                             if (decodePayloadResult.CloudToDeviceMessage != null)
                             {
-                                if (string.IsNullOrEmpty(decodePayloadResult.CloudToDeviceMessage.DevEUI) || string.Equals(loRaDevice.DevEUI, decodePayloadResult.CloudToDeviceMessage.DevEUI, StringComparison.OrdinalIgnoreCase))
+                                if (decodePayloadResult.CloudToDeviceMessage.DevEUI == default || loRaDevice.DevEUI == decodePayloadResult.CloudToDeviceMessage.DevEUI)
```
atifaziz, line 292: `is null` reads clearer here to mean missing:

```suggestion
                                if (decodePayloadResult.CloudToDeviceMessage.DevEUI is null || loRaDevice.DevEUI == decodePayloadResult.CloudToDeviceMessage.DevEUI)
```

Tests/Integration/DevAddrCacheTest.cs
bastbu: Done with 37e6dc642c7aac384bd0db4b8204c0aaf4c69cb9

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/ADR/LoRaADRManagerBase.cs
bastbu: Done in 0f086f54ffac5a62ee2933d0ef5368314920c7ef

LoRaEngine/LoraKeysManagerFacade/IoTHubDeviceInfo.cs
bastbu: As discussed offline, removed the check for empty strings entirely with 6ad941ed410bb1b31df520f46b8353801fb70cf8

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
```diff
@@ -41,7 +41,7 @@ public Task<LoRaDevice> CreateAndRegisterAsync(IoTHubDeviceInfo deviceInfo, Canc
         {
             _ = deviceInfo ?? throw new ArgumentNullException(nameof(deviceInfo));
 
-            if (string.IsNullOrEmpty(deviceInfo.PrimaryKey) || string.IsNullOrEmpty(deviceInfo.DevEUI))
+            if (string.IsNullOrEmpty(deviceInfo.PrimaryKey) || deviceInfo.DevEUI == default)
```
atifaziz, line 44: Better to check for validity:

```suggestion
            if (string.IsNullOrEmpty(deviceInfo.PrimaryKey) || !deviceInfo.DevEUI.IsValid)
```

Tests/Unit/LoRaTools/LoRaCloudToDeviceMessageTest.cs
```diff
@@ -157,5 +158,45 @@ public void When_Fport_0_And_Payload_Is_Empty_Should_Be_Invalid()
...
+
+        [Fact]
+        public void Can_Serialize_DevEui_NewtonsoftJson() =>
+            Can_Serialize_DevEui(m => Newtonsoft.Json.JsonConvert.SerializeObject(m));
+
+        private static void Can_Serialize_DevEui(Func<LoRaCloudToDeviceMessage, string> serialize)
+        {
+            var c2dMessage = new LoRaCloudToDeviceMessage { DevEUI = new DevEui(1) };
+            var result = serialize(c2dMessage);
+            Assert.Contains(@"""DevEUI"":""0000000000000001""", result, StringComparison.Ordinal);
```
atifaziz, line 199: Consider using a stronger check:

```suggestion
            var devEuiString = System.Text.Json.JsonDocument.Parse(result).RootElement.GetProperty("DevEUI").GetString();
            Assert.Equal("0000000000000001", devEuiString);
```

atifaziz (commented):
Based on our discussion, we want to make two more changes:

- Keep the device ID parameter on `ILoRaDeviceFactory.CreateClient` a string because we don't need to abstract the SDK client further.
- Change the search methods on `LoRaDeviceAPIServiceBase` to simply return the primary key since that's all used for now and would require separate types to represent device & station info.

The changes would be as follows:

```diff
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
index 4de996b9..3faea738 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
@@ -63,15 +63,15 @@ namespace LoRaWan.NetworkServer.BasicsStation
                 return await this.cache.GetOrCreateAsync(cacheKey, async cacheEntry =>
                 {
                     _ = cacheEntry.SetAbsoluteExpiration(CacheTimeout);
-                    var info = await this.loRaDeviceApiService.SearchByEuiAsync(stationEui);
-                    if (info is null)
+                    var primaryKey = await this.loRaDeviceApiService.SearchByEuiAsync(stationEui);
+                    if (primaryKey is null)
                     {
                         throw new LoRaProcessingException($"The configuration request of station '{stationEui}' did not match any configuration in IoT Hub. If you expect this connection request to succeed, make sure to provision the Basics Station in the device registry.",
                                                           LoRaProcessingErrorCode.InvalidDeviceConfiguration);
                     }
 
                     // A Station EUI is always also Dev EUI.
-                    using var client = this.loRaDeviceFactory.CreateDeviceClient(DevEui.Parse(info.StationEui.ToString()), info.PrimaryKey);
+                    using var client = this.loRaDeviceFactory.CreateDeviceClient(stationEui.ToString(), primaryKey);
                     var twin = await client.GetTwinAsync(cancellationToken);
                     return twin.Properties.Desired;
                 });
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ILoRaDeviceFactory.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ILoRaDeviceFactory.cs
index 85646047..e4f6cafe 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ILoRaDeviceFactory.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ILoRaDeviceFactory.cs
@@ -24,8 +24,8 @@ namespace LoRaWan.NetworkServer
         /// <summary>
         /// Creates a device client based on the devEUI and primary key.
         /// </summary>
-        /// <param name="eui">Dev EUI of the device.</param>
+        /// <param name="deviceId"></param>
         /// <param name="primaryKey">Primary key of the device.</param>
-        ILoRaDeviceClient CreateDeviceClient(DevEui eui, string primaryKey);
+        ILoRaDeviceClient CreateDeviceClient(string deviceId, string primaryKey);
     }
 }
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IotHubStationInfo.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IotHubStationInfo.cs
deleted file mode 100644
index aa92a7b5..00000000
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/IotHubStationInfo.cs
+++ /dev/null
@@ -1,28 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-namespace LoRaWan.NetworkServer
-{
-    using Newtonsoft.Json;
-
-    public sealed class IotHubStationInfo
-    {
-        [JsonProperty("DevEUI")]
-        public string StationEuiString
-        {
-            get => StationEui.ToString();
-            set => StationEui = StationEui.Parse(value);
-        }
-
-        [JsonIgnore]
-        public StationEui StationEui { get; set; }
-
-        public string PrimaryKey { get; set; }
-
-        public IotHubStationInfo(string stationEuiString, string primaryKey)
-        {
-            StationEuiString = stationEuiString;
-            PrimaryKey = primaryKey;
-        }
-    }
-}
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
index cfb24fa7..47f3cd85 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
@@ -8,6 +8,7 @@ namespace LoRaWan.NetworkServer
     using System.Diagnostics.Metrics;
     using System.Linq;
     using System.Text;
+    using System.Text.Json;
     using System.Threading;
     using System.Threading.Tasks;
     using LoRaTools.CommonAPI;
@@ -153,14 +154,14 @@ namespace LoRaWan.NetworkServer
         }
 
         /// <inheritdoc />
-        public override Task<IoTHubDeviceInfo> SearchByEuiAsync(DevEui eui) =>
-            SearchByEuiAsync<IoTHubDeviceInfo>(eui.ToString());
+        public override Task<string> SearchByEuiAsync(DevEui eui) =>
+            SearchByEuiAsync(eui.ToString());
 
         /// <inheritdoc />
-        public override Task<IotHubStationInfo> SearchByEuiAsync(StationEui eui) =>
-            SearchByEuiAsync<IotHubStationInfo>(eui.ToString());
+        public override Task<string> SearchByEuiAsync(StationEui eui) =>
+            SearchByEuiAsync(eui.ToString());
 
-        private async Task<T> SearchByEuiAsync<T>(string eui)
+        private async Task<string> SearchByEuiAsync(string eui)
         {
             this.deviceLoadRequests?.Add(1);
 
@@ -185,7 +186,7 @@ namespace LoRaWan.NetworkServer
             }
 
             var result = await response.Content.ReadAsStringAsync();
-            return JsonConvert.DeserializeObject<T>(result);
+            return JsonDocument.Parse(result).RootElement.GetProperty("PrimaryKey").GetString();
         }
 
         internal Uri GetFullUri(string relativePath)
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIServiceBase.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIServiceBase.cs
index 5297b5e9..485b9087 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIServiceBase.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIServiceBase.cs
@@ -44,7 +44,7 @@ namespace LoRaWan.NetworkServer
         /// Searches station devices in IoT Hub.
         /// </summary>
         /// <param name="eui">EUI of the station.</param>
-        public abstract Task<IotHubStationInfo> SearchByEuiAsync(StationEui eui);
+        public abstract Task<string> SearchByEuiAsync(StationEui eui);
 
         /// <summary>
         /// Fetch station credentials in IoT Hub.
@@ -56,7 +56,7 @@ namespace LoRaWan.NetworkServer
         /// Searches LoRa devices in IoT Hub.
         /// </summary>
         /// <param name="eui">EUI of the LoRa device.</param>
-        public abstract Task<IoTHubDeviceInfo> SearchByEuiAsync(DevEui eui);
+        public abstract Task<string> SearchByEuiAsync(DevEui eui);
 
         /// <summary>
         /// Sets the authorization code for the URL.
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
index de453c24..d44919ed 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
@@ -61,7 +61,7 @@ namespace LoRaWan.NetworkServer
                 // even though, we don't own it, to detect ownership
                 // changes.
                 // Ownership is transferred to connection manager.
-                this.connectionManager.Register(loRaDevice, CreateDeviceClient(deviceInfo.DevEUI, deviceInfo.PrimaryKey));
+                this.connectionManager.Register(loRaDevice, CreateDeviceClient(deviceInfo.DevEUI.ToString(), deviceInfo.PrimaryKey));
 
                 loRaDevice.SetRequestHandler(this.dataRequestHandler);
 
@@ -120,12 +120,12 @@ namespace LoRaWan.NetworkServer
             return connectionString;
         }
 
-        public virtual ILoRaDeviceClient CreateDeviceClient(DevEui eui, string primaryKey)
+        public virtual ILoRaDeviceClient CreateDeviceClient(string deviceId, string primaryKey)
         {
             try
             {
                 var partConnection = CreateIoTHubConnectionString();
-                var deviceConnectionStr = FormattableString.Invariant($"{partConnection}DeviceId={eui};SharedAccessKey={primaryKey}");
+                var deviceConnectionStr = FormattableString.Invariant($"{partConnection}DeviceId={deviceId};SharedAccessKey={primaryKey}");
 
                 // Enabling AMQP multiplexing
                 var transportSettings = new ITransportSettings[]
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
index e1ae8f68..0f9ecc84 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceRegistry.cs
@@ -154,11 +154,12 @@ namespace LoRaWan.NetworkServer
                 return cachedDevice;
             }
 
-            var device = await this.loRaDeviceAPIService.SearchByEuiAsync(devEUI);
-            if (device == null)
+            var primaryKey = await this.loRaDeviceAPIService.SearchByEuiAsync(devEUI);
+            if (primaryKey == null)
                 return null;
 
-            var loRaDevice = await this.deviceFactory.CreateAndRegisterAsync(device, CancellationToken.None);
+            var info = new IoTHubDeviceInfo { DevEUI = devEUI, PrimaryKey = primaryKey };
+            var loRaDevice = await this.deviceFactory.CreateAndRegisterAsync(info, CancellationToken.None);
 
             if (this.initializers != null)
             {
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
```diff
@@ -176,17 +177,22 @@ private async Task<SearchDevicesResult> SearchByEuiAsync(string eui)
...
             var result = await response.Content.ReadAsStringAsync();
-            var devices = (List<IoTHubDeviceInfo>)JsonConvert.DeserializeObject(result, typeof(List<IoTHubDeviceInfo>));
-            return new SearchDevicesResult(devices);
+
+            if (string.IsNullOrEmpty(result))
+                return null;
+
+            return JsonDocument.Parse(result).RootElement is { ValueKind: JsonValueKind.Object } someObject
+                ? someObject.TryGetProperty("PrimaryKey", out var propertyValue) ? propertyValue.GetString() : null
+                : null;
```
atifaziz, line 195: nit: Perhaps go one more step further here and make sure that `PrimaryKey` is a string?

```suggestion
            return result is { Length: > 0 } json
                   && JsonDocument.Parse(json).RootElement is { ValueKind: JsonValueKind.Object } someObject
                   && someObject.TryGetProperty("PrimaryKey", out var propertyValue)
                   && propertyValue is { ValueKind: JsonValueKind.String }
                 ? propertyValue.GetString()
                 : null;
```

PR #1264, opened 2022-01-13
Reactivate metricscollector in CI

Mandur:
Reactivate metricscollector in CI


LoRaEngine/deployment.ci.aio.template.json
```diff
@@ -184,6 +184,31 @@
...
+                "value": "AzureMonitor"
+              },
+              "LogAnalyticsWorkspaceId": {
+                "value": "$LOG_ANALYTICS_WORKSPACE_ID"
+              },
+              "LogAnalyticsSharedKey": {
+                "value": "$LOG_ANALYTICS_SHARED_KEY"
+              },
+              "MetricsEndpointsCSV": {
+                "value": "http://edgeHub:9600/metrics,http://edgeAgent:9600/metrics,http://LoRaWanNetworkSrvModule:5000/metrics"
```
bastbu, line 207: Since we have the server pfx set here, will we not run into the issue of the self-signed certs and the metrics collector? (the LNS should listen on port 5001 in that case, no?)

LoRaEngine/deployment.ci.lbs.template.json
```diff
@@ -102,6 +102,31 @@
             "version": "1.0",
             "status": "running",
             "restartPolicy": "always"
+          },
+          "metricscollector": {
```
bastbu, line 106: If we set that on EFLOW and the AIO gateway, why do we need it again in the LBS template?

bastbu (approved):

LoRaEngine/deployment.ci.lbs.template.json
Mandur: Imho it helps keep a baseline with a deployment not deploying everything

LoRaEngine/deployment.ci.aio.template.json
Mandur: Very good point I removed it from here.

Issue #1265, opened 2022-01-13
When E2E TCP logs are not found, the searched message is not printed


spygi:
When a TCP log is not found sometimes we the message we are looking for is not printed out, for example:
```09:08:28.354 TCP log message not found, attempt 1/5, waiting 15 secs```

This happens in OTAA tests in [this run for example](https://github.com/Azure/iotedge-lorawan-starterkit/runs/4786557102?check_suite_focus=true#step:10:953)

PR #1266, opened 2022-01-13
Print out search TCP/IoT message in case of failures

spygi:
# PR for issue #1265 

## What is being addressed

For some E2E tests when the TCP logs are not found, they are not printed out. Printing them would help debugging issues in the E2E tests.

## How is this addressed

- By removing the no-arg constructor for the `SearchLogOptions` the Description is now required. We also enable the nullable context there: although it's not applicable now (the caller needs also to be nullable-enabled) we make it easier to switch to nullable context in the future.
- We make the options as required parameter in the `IntegrationTestFixtureBase` methods. The caller could still pass null which would result in a NullReferenceException but that should be acceptable because it's test code. Plus if we switch to nullable we would not need to remove "?".
- Fixes a bug where we would log "data" instead of the value that we were looking for in the logs (that provides more context) commit: c411ae65e8473c138aacbc44966712974b6a9b8f
- The IoT Hub logs are also fixed.

## Verification

- Errors were introduced to check whether the log messages get printed out in the OTAA test, [see for example the run](https://github.com/Azure/iotedge-lorawan-starterkit/runs/4804571049?check_suite_focus=true#step:10:2443)
```14:32:21.800 TCP log message '01F5A9FF00000004: sending message to station with EUI TYPO1' not found, attempt 1/5, waiting 15 secs``` There were no missed messages without the description. Commit was reverted in 2b01c67c46ad2ae62c75d4eb40118eaa65218625
- The same was done for the `MultiConcentratorTests` in [this run](https://github.com/Azure/iotedge-lorawan-starterkit/runs/4843786145?check_suite_focus=true) to check whether the bug was fixed in c411ae65e8473c138aacbc44966712974b6a9b8f - we see the value that was searched and not just "data"
```17:09:38.270 IoT Hub message '{"value":105 TYPO }' not found, attempt 3/5, waiting 45 secs```

### CODE REVIEW

File: Tests/Common/IntegrationTestFixtureBase.Asserts.cs [hunk 1]
```diff
@@ -103,7 +103,7 @@ public async Task AssertNetworkServerModuleLogStartsWithAsync(string logMessageS
                 new SearchLogOptions(string.Concat(logMessageStart1, " or ", logMessageStart2)));
         }
 
-        public async Task<SearchLogResult> SearchNetworkServerModuleAsync(Func<string, bool> predicate, SearchLogOptions options = null) =>
+        public async Task<SearchLogResult> SearchNetworkServerModuleAsync(Func<string, bool> predicate, SearchLogOptions options) =>
```
bastbu, line 106: I would argue that making the `SearchLogOptions` not `null` is not enough to guarantee that the `Description` is always set. To enforce setting the `Description` always we should disallow the no-args constructor and validate that the `description` is not null or empty when constructing the `SearchLogOptions`.

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
```diff
@@ -350,21 +350,21 @@ private async Task<SearchLogResult> SearchTcpLogs(Func<SearchLogEvent, bool> pre
...
-            var maxAttempts = options?.MaxAttempts ?? Configuration.EnsureHasEventMaximumTries;
+            var maxAttempts = options.MaxAttempts ?? Configuration.EnsureHasEventMaximumTries;
             var processedEvents = new HashSet<SearchLogEvent>();
             for (var i = 0; i < maxAttempts; i++)
             {
                 if (i > 0)
                 {
                     var timeToWait = i * Configuration.EnsureHasEventDelayBetweenReadsInSeconds;
-                    if (!string.IsNullOrEmpty(options?.Description))
+                    if (!string.IsNullOrEmpty(options.Description))
```
bastbu, line 367: [nit] this is a statement that should now never be true, no (there are other occurrences, just commenting this once)?

bastbu (approved):
There are a few occurrences of `string.IsNullOrEmpty(options.Description)` that we could now get rid of. Other than that, if E2E succeeds, this looks good to me.

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
spygi: Fair point, done this in fc631626d0e6c7a52dd6aea24183a4d727d99062

Tests/Common/IntegrationTestFixtureBase.Asserts.cs
spygi: After the changes in fc631626d0e6c7a52dd6aea24183a4d727d99062 and 2a4c194d98605f899b707c9092af81c5535acce7 that is the case yes, so the checks were removed.

Issue #1009, opened 2021-12-07
Fix Universal Decoder CI


Mandur:
## Description

- [ ] Fix Universal Decoder CI by preventing it to push docker container on each build, only when an input is provided on manual push.
- [x] Fix notice file

PR #1268, opened 2022-01-13
Universal Decoder: don't push docker image on CI

roel4ez:
# PR for issue #1009

## What is being addressed

Until now the docker image would be pushed to the repository on each CI run. We only want to push the docker image on request.

## How is this addressed

Introduces a parameter `pushDockerImage` that can be set to conditionally push the docker image when the pipeline is triggered manually. For a CI run, the parameter will be `failse`


.github/workflows/universal_decoder_ci.yaml
```diff
@@ -13,6 +13,12 @@ on:
     paths:
       - 'Samples/UniversalDecoder/**'
   workflow_dispatch:
+    inputs:
+      pushDockerImage:
```
bastbu, line 17: Since the pipeline is not enforced to succeed on every PR run, I would argue that instead of adding complexity by introducing new parameters, we should instead consider just removing the triggers that run the pipeline on each PR (and maybe only run it after a merge). What do you think?

bastbu (approved):

.github/workflows/universal_decoder_ci.yaml
```diff
@@ -71,13 +77,17 @@ jobs:
       - name: Get the image version
         id: vars
         run: echo ::set-output name=tag::0.0.$(echo ${GITHUB_SHA})
+      - name: Login to docker
+        run: docker login -u ${{ secrets.DOCKER_LOGIN }} -p ${{ secrets.DOCKER_PASSWORD }}
```
danigian, line 81: This step should be conditional too

.github/workflows/universal_decoder_ci.yaml
```diff
@@ -13,6 +13,12 @@ on:
     paths:
       - 'Samples/UniversalDecoder/**'
   workflow_dispatch:
+    inputs:
+      pushDockerImage:
+        description: 'Push docker image to the repository'
+        type: boolean
+        required: false
+        default: 'false'
```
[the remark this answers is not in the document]
roel4ez: yeah this is interesting, it's what I thought as well, but:
image

I played around with it and it does not work with an actual bool

.github/workflows/universal_decoder_ci.yaml
roel4ez: moved the login to the task where it is neeed

Issue #1051, opened 2021-12-13
The frontpage should show an up-to-date overview of the current feature set


roel4ez:
Currently the front page of both the documentation and the github repo is outdated and shows duplicate information.
Instead, it should show an up-to-date overview of the current supported feature set, and link into the relevant feature details such as regional support, scenarios, etc ...

It needs to be a good one-pager that can be used as an ["elevator pitch](https://en.wikipedia.org/wiki/Elevator_pitch)"

Mandur:
@roel4ez Did you want to work on this next sprint?

PR #1270, opened 2022-01-13
[docs] Update main image and main page text

roel4ez:
# Issue #1051

- Add updated image of architecture
- Rewording and adding some features


docs/index.md
```diff
@@ -12,27 +12,41 @@ hide:
...
 message packets to Azure IoT Hub for cloud based processing, analytics and other
-workloads. Alternatively, it allows sending commands from the cloud to the end
+workloads.
+
+Alternatively, it allows sending commands from the cloud to the end
 nodes. The goal of the the project is to provide guidance and a reference for
 Azure IoT Edge users to experiment with LoRaWAN technology.
 
-![Architecture](images/EdgeArchitecture.png)
+![Architecture](images/EdgeArchitecture-updated.png)
```
p-schuler, line 30: nit: if you want to rename the file, I would use a version rather than "updated". git is managing the history, so you can also argue that you can overwrite it and the 2.0 version shows the right image and so does < 2.0.

docs/index.md
```diff
@@ -12,27 +12,41 @@ hide:
...
+
+## Key Challenges
+
+- Allows LoRa-enabled field gateways to connect directly to Azure without the need for a network operator as intermediary.
+- Enables Azure IoT Edge capabilities for the LoRaWAN network, e.g.:
+  - Local processing on the Edge gateway
+  - Routing to local storage from the Edge gateway
+  - Offline capabilities of the gateway
+- Homogenous management of devices and concentrators independent of connectivity technology through Azure IoT.
+- Off-the-shelf integration with Azure IoT ecosystem, e.g., Azure IoT Central, Azure Digital Twins, Time Series Insights, etc...
```
p-schuler, line 40: Azure IoT Central is not yet in.

docs/index.md
```diff
@@ -12,27 +12,41 @@ hide:
...
 
 ## Features
 
 - LoRaWAN 1.0.2 implementation
 (see [LoRaWAN Specification Support](#LoRaWAN-1.0.2-Specification-Support)
 for more details)
-- Multiple concentrator support
-- Device and Gateway management done completely through Azure IoT Hub.
+- Device and Concentrator management done completely through Azure IoT Hub.
```
p-schuler, line 47: I feel like "completely" is a stretch. It was like this before, but you need to manage OS updates for example.

p-schuler (approved):
Generally, as an overview this looks good. I think we may want to have some documentation about deployment scenarios and what possibilities we enable there.

On the image: I think the picture with the 2 different concentrators is a bit misleading. The LoRaWAN gateway (especially the one that looks like a PI in a housing) could be mislead to be a LNS. That is a valid depoyment scenario, but on the picture that's not clear, as it also connects to the LNS separately.

docs/index.md
roel4ez: fair point!

roel4ez:
> Generally, as an overview this looks good. I think we may want to have some documentation about deployment scenarios and what possibilities we enable there.

Yep agreed, there is an issue open to track that. 

> On the image: I think the picture with the 2 different concentrators is a bit misleading. The LoRaWAN gateway (especially the one that looks like a PI in a housing) could be mislead to be a LNS. That is a valid depoyment scenario, but on the picture that's not clear, as it also connects to the LNS separately.
Good point, I've replaced it with a different image (Browan)

Issue #1221, opened 2022-01-11
Use native websocket client instead of external library for Simulator project


Mandur:
Please refer to https://github.com/Azure/iotedge-lorawan-starterkit/pull/1153#discussion_r781059740
<img width="728" alt="image" src="https://user-images.githubusercontent.com/8555833/148942511-2581d857-0fca-42ca-be60-54f9e8bd656d.png">

PR #1277, opened 2022-01-13
Use .NET WebSocket client in simulator

bastbu:
# PR for issue #1221 

## What is being addressed

With this PR we start using the .NET websocket client instead of the package we used so far. We also start raising received messages on a simulated basics station as events, instead of using the custom `SubscribeOnce` mechanism.


spygi (approved):
From inspecting the code (esp. the SimulatedBasicsStation) looks good to me (i.e. the sequence of steps taken, using the EventHandler, cancellation tokens and other best practices). Since this is not yet integrated with CI, I think it's ok to merge and when we have the pipeline support we can still make changes/review that PR.

PR #1289, opened 2022-01-17
Change `LoRaPayloadJoinAccept.RxDelay` to `RxDelay` type

atifaziz:
This PR adds to #1144, which missed an opportunity to use the `RxDelay` type.

## What is being addressed

`LoRaPayloadJoinAccept.RxDelay` was typed as `Memory<byte>`.

## How is this addressed

The type of the property is changed to the type `RxDelay`.


danigian (approved):
LGTM. Also validated by manually setting up an OTAA device with RXDelay==6, to verify that JoinAccept is properly handled in non-standard situation

PR #1233, opened 2022-01-12
added new NOTICE file Universal_Decoder_NOTICE 41545

roel4ez:

PR #1290, opened 2022-01-17
Multi-concentrator/multi-gateway compatible load tests

bastbu:
# PR for issue #1233 

PR co-authored by @Mandur 

## What is being addressed

With this PR we integrate the (simulated) load tests into the pipelines. They do not run per default, but they can easily be manually run by setting the input parameter `RunLoadTests` to `true`. This will simplify further reviews. They also run every night on schedule.

## How is this addressed

Even though we do not yet use a set of parameters that would guarantee a load test under maximum load, we enable the tests to make sure that they do not break again and to start asserting basic validity properties, such as:
- that all messages are delivered to IoT Hub (considering the deduplication strategies and a unique message fragment which is used to identify messages from a particular test case)
- that all confirmed messages receive a downlink response.

We are also:
- Introduce a mechanism to run the tests for an arbitrary number of concentrators, by automatically creating the devices in IoT Hub for these concentrators.
- Introduce a mechanism to distribute the concentrator-gateway connections between a collection of gateways, based on round-robin distribution
- Remove unnecessary values from the `appsettings.json` to clean up the configuration that we need for the load tests.
- Disable TCP logging for load tests, since this would flood the disks of the runner agent


.github/workflows/e2e-ci.yaml
```diff
@@ -376,7 +379,67 @@ jobs:
...
+      - certificates_job
+    runs-on: [ self-hosted, x64 ]
+    env:
+      INTEGRATIONTEST_IoTHubEventHubConnectionString: ${{ secrets.IOTHUB_EVENT_HUB_CONNECTION_STRING }}
+      INTEGRATIONTEST_IoTHubEventHubConsumerGroup: ${{ secrets.IOTHUB_CI_CONSUMER_GROUP }}
+      INTEGRATIONTEST_IoTHubConnectionString: ${{ secrets.IOTHUB_OWNER_CONNECTION_STRING }}
+      INTEGRATIONTEST_LeafDeviceGatewayID: itestarm1
+      INTEGRATIONTEST_DevicePrefix: '03'
+      INTEGRATIONTEST_RunningInCI: true
+      INTEGRATIONTEST_LoadTestLnsEndpoints: ${{ '[\"wss://itestarm1:5001\"]' }}
```
bastbu, line 404: Note: I will extract this to the `secrets` once I know if we need other configuration values that depend on the workflow runtime environment (before merging).

Tests/Common/SimulatedBasicsStation.cs
```diff
@@ -147,6 +144,15 @@ public async Task StopAndValidateAsync(CancellationToken cancellationToken = def
             }
         }
 
+        private static ClientWebSocket CreateClientWebSocket()
+        {
+            var result = new ClientWebSocket();
+#pragma warning disable CA5359 // Do Not Disable Certificate Validation
```
danigian, line 150: Let's add a comment saying that we are making use of self-signed certificates on the LNS that we are targeting for Load Tests and therefore we need to either disable the validation or trust the cert (we chose the first)

Tests/Common/SimulatedDevice.cs
```diff
@@ -226,50 +251,72 @@ private bool HandleJoinAccept(LoRaPayloadJoinAccept payload)
...
-                    var joinAccept = new LoRaPayloadJoinAccept(ConversionHelper.StringToByteArray(joinAcceptstring.Pdu), AppKey.Value);
-
-                    var result = HandleJoinAccept(joinAccept); // may need to return bool and only release if true.
-                    joinSuccessful = result;
+                    var joinAcceptResponse = JsonSerializer.Deserialize<JoinAcceptResponse>(response.Value);
+                    // PDU is null in case it is another message coming from the station.
+                    if (joinAcceptResponse is { } someJoinAcceptResponse && someJoinAcceptResponse.Pdu is { } somePdu && someJoinAcceptResponse.DevEui == DevEUI)
+                    {
+                        var joinAccept = new LoRaPayloadJoinAccept(ConversionHelper.StringToByteArray(somePdu), AppKey.Value);
+                        joinSuccessful = HandleJoinAccept(joinAccept); // may need to return bool and only release if true.
```
danigian, line 282: What about this comment?

Tests/Simulation/appsettings.json
```diff
@@ -3,14 +3,13 @@
...
     "NetworkServerModuleLogAssertLevel": "Error",
     "IoTHubAssertLevel": "Warning",
-    "TcpLog": "true",
+    "TcpLog": false,
     "CreateDevices": true,
     "LeafDeviceGatewayID": "#{INTEGRATIONTEST_LeafDeviceGatewayID}#",
-    "DevicePrefix": "#{INTEGRATIONTEST_DevicePrefix}#"
+    "DevicePrefix": "#{INTEGRATIONTEST_DevicePrefix}#",
+    "LoadTestLnsEndpointsString": "#{INTEGRATIONTEST_LoadTestLnsEndpoints}#",
+    "NumberOfLoadTestDevices": 20
```
danigian, line 13: What about setting NumberOfLoadTestDevices configurable from pipeline?

Tests/Common/SimulatedBasicsStation.cs
bastbu: 98c2a895d0ba721431d1f824aeb5d2c43f421b92

Tests/Common/SimulatedDevice.cs
bastbu: good catch, done in d71a6be4c05e105cc3c92f3efd5a2076325da6e6

.github/workflows/e2e-ci.yaml
bastbu: Done with 3219e66a3c6c9e5ec34cdefc358d099360e6c6d1

Tests/Simulation/appsettings.json
bastbu: Done with 3219e66a3c6c9e5ec34cdefc358d099360e6c6d1

bastbu:
Successful load test run: https://github.com/Azure/iotedge-lorawan-starterkit/runs/4867760354?check_suite_focus=true

danigian (approved):

Issue #973, opened 2021-12-03
Replace uses of `ConversionHelper` with `Hexadecimal`


atifaziz:

bastbu:
This story would benefit from a cleanup of other primitives first (DevEUI, DevAddr etc). With that cleanup, some occurrences of the helper will go away.

atifaziz:
> This story would benefit from a cleanup of other primitives first (DevEUI, DevAddr etc). With that cleanup, some occurrences of the helper will go away.

Indeed and with the merge of PR #1077 (`DevAddr` reuse), we are down to 22 usages from 67, a 67% reduction.

PR #1293, opened 2022-01-18
Replace `ConversionHelper` with `Hexadecimal`

atifaziz:
# PR for issue #973

## What is being addressed

`ConversionHelper` was redundant with functionality in `Hexadecimal`.

## How is this addressed

Uses of `ConversionHelper` have been replaced with `Hexadecimal`.

## Summary of related changes

- `LittleEndianReader` is added to consolidate duplication in reading of 24-bit unsigned integers in little-endian byte order.
- Extension methods have been added (see `HexadecimalExtensions`) to help with conversion of byte arrays to a hexadecimal string, which internally uses `Hexadecimal`. It uses the stack to format small byte array (<= 32 bytes or where result string will be <= 64 bytes in length where each character = 2 bytes) and the heap otherwise.


Tests/Common/SimulatedDevice.cs
```diff
@@ -295,5 +294,16 @@ public void SetupJoin(AppSessionKey appSKey, NetworkSessionKey nwkSKey, DevAddr
...
             LoRaDevice.DevAddr = devAddr;
         }
+
+        private static byte[] StringToByteArray(string hex)
+        {
+            if (hex is null) throw new ArgumentNullException(nameof(hex));
+
+            var numberChars = hex.Length;
+            var bytes = new byte[numberChars / 2];
+            for (var i = 0; i < numberChars; i += 2)
```
bastbu, line 304: [nit] why not reuse `Hexadecimal` for this private method?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/LittleEndianReader.cs
```diff
@@ -0,0 +1,21 @@
...
+    using System.Runtime.CompilerServices;
+
+    internal static class LittleEndianReader
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static uint ReadUInt24(ReadOnlySpan<byte> buffer) =>
+            buffer.Length >= 3 ? buffer[0] | ((uint)buffer[1] << 8) | ((uint)buffer[2] << 16) : ThrowInsufficientBufferLength();
+
+        [DoesNotReturn]
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
bastbu, line 17: Curious - why these `MethodImpl` attributes on these methods?

bastbu (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan/LittleEndianReader.cs
atifaziz: To override JIT's heuristics. The JIT doesn't inline methods with a `throw` so the throwing is done in a separate method that it's told not to inline.

Suppose you have the following:

```c#
public class C
{
    public uint M(byte[] buffer) =>
        LittleEndianReader.ReadUInt24(buffer);
}
```

Then you can see the inlining in action (see also on [sharplab.io]):

```asm
C.M(Byte[])
    L0000: sub rsp, 0x28
    L0004: test rdx, rdx
    L0007: je short L001c
    L0009: lea rax, [rdx+0x10]
    L000d: mov edx, [rdx+8]
    L0010: cmp edx, 3
    L0013: jge short L0022
    L0015: call 0x00007ffb39690038
    L001a: jmp short L0037
    L001c: xor eax, eax
    L001e: xor edx, edx
    L0020: jmp short L0010
    L0022: movzx edx, byte ptr [rax]
    L0025: movzx ecx, byte ptr [rax+1]
    L0029: shl ecx, 8
    L002c: or edx, ecx
    L002e: movzx eax, byte ptr [rax+2]
    L0032: shl eax, 0x10
    L0035: or eax, edx
    L0037: add rsp, 0x28
    L003b: ret
```

Notice the length 3 check at offset L0010 and left-shifts (`shl`) at offsets L0029 and L0032 appear _within_ the compiled method.

If you remove the `[MethodImpl(MethodImplOptions.AggressiveInlining)]` decoration on `LittleEndianReader.ReadUInt24`, you get a shorted body and `call` instead on L0025:

```asm
C.M(Byte[])
    L0000: sub rsp, 0x38
    L0004: xor eax, eax
    L0006: mov [rsp+0x28], rax
    L000b: test rdx, rdx
    L000e: je short L0030
    L0010: lea rcx, [rdx+0x10]
    L0014: mov eax, [rdx+8]
    L0017: mov [rsp+0x28], rcx
    L001c: mov [rsp+0x30], eax
    L0020: lea rcx, [rsp+0x28]
    L0025: call 0x00007ffb39760030
    L002a: nop
    L002b: add rsp, 0x38
    L002f: ret
    L0030: xor ecx, ecx
    L0032: xor eax, eax
    L0034: jmp short L0017
```

[sharplab.io]: https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgBEBLbAcwDsJcGDmFwsAwhAAmMAIK9sAGwCeuDrgDcNek1YAlAK68h+GOIj4ADhwUwoAZVsA3YTA00tAZiakGYmgG8aBmCmL30OIwYAWQAKYCUMGABtAF0GYH0AM0zbAEoGAF4APiCQsoAZDgwMGwBRXkkuXl0YbGkoFha2gFUASSNSFDisnKhczWoAX3daL2YkbwZK6rqGpq72gNLgpKiYDAALKV7LBRi9w+PTgHkLIQheURlublhcVUcYfoUIiO5clLbUI6BbhSIbPoDIYba68ZR2CzYXgAHniiSK6RGeUKGMMYAOMDAAGsYJIYkCyhlsrYWOUYLxuIcGEUCgwvAB+THUqBJOhpQgMGIxMEYXJU0ZJchpZHIhgADnyAqFIrFWJ5pGlsvISFyFLK+oNhpADAAKgcoBAAO79XAjYQcekYABCarpDMOMVy4xmZSSbAgrgAchAMC0MPooLxAdRfRcjpIThYznGrknbvdHixg99fgyAUCLFAOI5sIkQQwRabzVabXawA6jC7uW7GQdPTi9cFDhbLQxeDBezIoNx9CYjLUEGAYHcOA8YgAiWvZe2OrmjBg2d0HFjztAMefi2zz71TGhAA

Tests/Common/SimulatedDevice.cs
atifaziz: Good point. Done with 4c4ed9f8e97b508237bebbe04081655e0ccb29ed.

Issue #1251, opened 2022-01-12
CUPS Firmware Upgrade - Provide a step-by-step guide on how to execute a firmware upgrade


danigian:
- [x] Add user guide for the firmware upgrade
- [x] Link to Tools/Cups-Firmware-Upgrade
- [x] Update missing CLI parameters in the ADR

PR #1295, opened 2022-01-18
User guide for executing a firmware upgrade

MaggieSalak:
# PR for issue #1251 

## What is being addressed

User guide for executing a Basics Station firmware upgrade

## How is it addressed

- User guide added with steps required to run an upgrade. The guide links to [Firmware Upgrade Preparation](https://github.com/Azure/iotedge-lorawan-starterkit/tree/dev/Tools/Cups-Firmware-Upgrade) tool for details & script that generates they key, digest, checksum files.
- The Device Provisioning CLI tool documentation has been updated with the new command description (`upgrade-firmware`)
- The CUPS firmware upgrade ADR has been updated with previously missing parameter for the CLI (`StationEui`)


docs/user-guide/station-firmware-upgrade.md
```diff
@@ -0,0 +1,57 @@
...
+1. Retrieve the CRC32 Checksum of the signature key
+
+```shell
+cat sig-0.key | gzip -1 | tail -c 8 | od -t ${1:-u}4 -N 4 -An --endian=little | xargs echo > sig-0.crc
+```
+
+1. Use LoRa Device Provisioning CLI tool to trigger the upgrade
+
+```powershell
+dotnet run -- upgrade --deveui <station_eui> --version <version> --file <upgrade_file> --digest <file_digest> --checksum <checksum>
```
MaggieSalak, line 38: This part is no ready yet but I think we can already agree on that the parameters for the CLI tool will look like - let me know what you think @danigian

docs/user-guide/station-firmware-upgrade.md
```diff
@@ -0,0 +1,54 @@
...
+1. Retrieve the CRC32 Checksum of the signature key
+
+   ```shell
+   cat sig-0.key | gzip -1 | tail -c 8 | od -t ${1:-u}4 -N 4 -An --endian=little | xargs echo > sig-0.crc
+   ```
+
+1. Use LoRa Device Provisioning CLI tool to trigger the upgrade
+
+   ```powershell
+   dotnet run -- upgrade --deveui <station_eui> --version <version> --file <upgrade_file> --digest <file_digest> --checksum <checksum>
```
danigian, line 35: We are going to include a compiled version of the tool in our release, therefore user should not execute it with "dotnet run" but directly calling the binary

danigian (commented):
I will wait approving this PR until we have the updates to provisioning CLI in dev

MaggieSalak:
> I will wait approving this PR until we have the updates to provisioning CLI in dev

That would mean this PR would be blocked for quite some time. I think since we assigned this user story to this sprint, and not the CLI-related one, we should make a decision about what the CLI tool will accept as parameters, since in any case this will need to be determined very soon. @danigian

docs/user-guide/station-firmware-upgrade.md
MaggieSalak: Wouldn't that be out of line with the current user guide for the CLI tool, where we instruct the user to use `dotnet run`? https://azure.github.io/iotedge-lorawan-starterkit/dev/tools/device-provisioning/

docs/user-guide/station-firmware-upgrade.md
MaggieSalak: Updated this to be in line with the CLI implementation.

docs/user-guide/station-firmware-upgrade.md
MaggieSalak: Updated the doc to recommend calling the binary of the CLI directly. I also added a link to the documentation of the CLI tool (which is now also updated).

danigian (approved):

PR #1297, opened 2022-01-18
CUPS - Firmware upgrade logic implementation

danigian:
# PR for issue #1248 

## What is being addressed

This PR is implementing required logic for CUPS Firmware Upgrade management.

## How is this addressed

- [x] CupsResponse SerializeAsync is now handling the eventuality of a firmware upgrade
- [x] Related CupsTwinInfo and CupsUpdateInfoRequest have been extended, CupsEndpoint JsonHandler now handles additional package and keys fields
- [x] CupsProtocolMessageProcessor.cs has been adjusted to handle the new CupsResponse SerializeAsync return values
- [x] LoRaDeviceAPIService now implements a FetchStationFirmwareAsync method that is asynchronously streaming the Firmware from Facade Function
- [x] Unit tests have been adjusted properly (E2E test will follow in another PR to check that no regressions are introduced)

Validation of the work has been done by manually setting up the required fields in a concentrator twin and by manually uploading a firmware file to storage. BasicStation has been then executed with connection to CUPS and the update script was properly executed.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsResponse.cs
```diff
@@ -77,12 +72,48 @@ private int WriteUriConditionally(ConcentratorCredentialType endpointType, Span<
...
+
+            if (firmwareMismatch)
+            {
+                if (!this.cupsUpdateInfoRequest.KeyChecksums.Any(c => c == this.cupsTwinInfo.FwKeyChecksum))
+                {
+                    throw new InvalidOperationException("Remote firmware signature generated with a key whose checksum is not available in CUPS request.");
+                }
+
+                var signature = Convert.FromBase64String(this.cupsTwinInfo.FwSignature);
+                currentPosition += WriteToSpan(signature.Length + 4, response.Span[currentPosition..]);
```
atifaziz, line 93: What's the reason for adding 4 to the signature length?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsTwinInfo.cs
```diff
@@ -51,5 +57,17 @@ internal sealed record CupsTwinInfo
...
+        [JsonProperty("package")]
+        public string Package { get; }
+
+        [JsonPropertyName("fwKeyChecksum")]
+        [JsonProperty("fwKeyChecksum")]
+        public int FwKeyChecksum { get; }
+
+        [JsonPropertyName("fwSignature")]
+        [JsonProperty("fwSignature")]
+        public string FwSignature { get; }
```
atifaziz, line 71: What about calling this `FwSignatureInBase64` so it's obvious the string is Base64-encoded data?

Tests/Common/IntegrationTestFixtureBase.cs
```diff
@@ -322,7 +322,10 @@ public async Task UpdateExistingConcentratorCrcValues(StationEui stationEui, uin
                                                crc,
                                                crc,
                                                deserializedCupsTwinInfo.CupsCredentialUrl,
-                                               deserializedCupsTwinInfo.TcCredentialUrl);
+                                               deserializedCupsTwinInfo.TcCredentialUrl,
```
atifaziz, line 325: If you update `CupsTwinInfo` as follows so that its `CupsCredCrc` and `TcCredCrc` properties have initialisers:

```diff
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsTwinInfo.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsTwinInfo.cs
index cb7959e7..774c3649 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsTwinInfo.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsTwinInfo.cs
@@ -44,11 +44,11 @@ namespace LoRaWan.NetworkServer.BasicsStation
 
         [JsonPropertyName("cupsCredCrc")]
         [JsonProperty("cupsCredCrc")]
-        public uint CupsCredCrc { get; }
+        public uint CupsCredCrc { get; init; }
 
         [JsonPropertyName("tcCredCrc")]
         [JsonProperty("tcCredCrc")]
-        public uint TcCredCrc { get; }
+        public uint TcCredCrc { get; init; }
 
         [JsonPropertyName("cupsCredentialUrl")]
         [JsonProperty("cupsCredentialUrl")]
```

then you can use `with` here to have the compiler use the copy constructor (which is what you're doing in effect here):

```c#
            var newCupsInfo = JsonConvert.DeserializeObject<CupsTwinInfo>(cupsJson) with
            {
                CupsCredCrc = crc,
                TcCredCrc = crc
            };
```

Now this code doesn't have to be updated as more properties are added/removed in `CupsTwinInfo` or its constructor updated.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsTwinInfo.cs
```diff
@@ -51,5 +57,17 @@ internal sealed record CupsTwinInfo
...
         [JsonProperty("tcCredentialUrl")]
         public string TcCredentialUrl { get; }
+
+        [JsonPropertyName("package")]
+        [JsonProperty("package")]
+        public string Package { get; }
+
+        [JsonPropertyName("fwKeyChecksum")]
+        [JsonProperty("fwKeyChecksum")]
+        public int FwKeyChecksum { get; }
```
atifaziz, line 67: Is the checksum a _signed_ integer?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsResponse.cs
danigian: https://github.com/lorabasics/basicstation/issues/145

This is the reason @atifaziz

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsTwinInfo.cs
danigian: No, it's an unsigned one. Thanks for catching it

Tests/Common/IntegrationTestFixtureBase.cs
atifaziz: Done with 48377fa025c6d2ec6185c030b5bcd6a13f89d4f2.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIServiceBase.cs
```diff
@@ -52,6 +53,13 @@ public abstract class LoRaDeviceAPIServiceBase
         /// <param name="eui">EUI of the station.</param>
         public abstract Task<string> FetchStationCredentialsAsync(StationEui eui, ConcentratorCredentialType credentialtype, CancellationToken token);
 
+        /// <summary>
+        /// Fetch station firmware file.
+        /// </summary>
+        /// <param name="eui">EUI of the station.</param>
+        /// <returns>A tuple containing content-length and stream.</returns>
+        public abstract Task<(long?, Stream)> FetchStationFirmwareAsync(StationEui eui, CancellationToken token);
```
atifaziz, line 61: I believe the content length is always return so shouldn't be nullable.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/CupsResponse.cs
atifaziz: We should definitely reference that issue (for future readers/maintainers) as a comment in the code.

atifaziz: So if I understand correctly, the `sigLen` includes the 4 bytes of `keyCRC` _and_ `sig` fields of the CUPS response, which makes sense since if `sigLen` is 0 then `keyCRC` should be skipped. So `keyCRC` and `sig` should _both_ be present or absent.

atifaziz (commented):
Please see commit 5d0154270c8aef53da0aa9d6bfcae1c1b27d86c6 for possibly a simpler way to represent the CUPS response header that's more [aligned with the spec](https://lora-developers.semtech.com/build/software/lora-basics/lora-basics-for-gateways/?url=cupsproto.html#http-post-response) (with supporting span-writing code moved out as extensions that we can reuse elsewhere too). The tests need adjusting and happy to give it a final look when done. Thanks!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIServiceBase.cs
```diff
@@ -52,6 +53,13 @@ public abstract class LoRaDeviceAPIServiceBase
         /// <param name="eui">EUI of the station.</param>
         public abstract Task<string> FetchStationCredentialsAsync(StationEui eui, ConcentratorCredentialType credentialtype, CancellationToken token);
 
+        /// <summary>
+        /// Fetch station firmware file.
+        /// </summary>
+        /// <param name="eui">EUI of the station.</param>
+        /// <returns>A tuple containing content-length and stream.</returns>
```
atifaziz, line 60:
```suggestion
        /// <returns>An instance of <see cref="HttpContent"/> with the firmware content.</returns>
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Unit.NetworkServer
+{
+    using System;
+    using System.Buffers;
+    using System.Collections.Immutable;
```
atifaziz, line 8: Remove unused import:

```suggestion
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            Assert.True(serializedResponse.ToArray().All(b => b == 0));
+        }
+
+        [Fact]
+        public void Serialize_WithFirmwareUpdates()
+        {
+            // setting up the twin in such a way that there are only firmware updates
+            var signatureBytes = Convert.FromBase64String("ABCD");
+            var keyCRC = 6789U;
+            var updateDataLength = 100U;
```
atifaziz, line 41: These can be constants:

```suggestion
            const uint keyCRC = 6789;
            const uint updateDataLength = 100;
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            using var memoryPool = MemoryPool<byte>.Shared.Rent(2048);
+
+            // Act
+            var serializedResponse = updateResponseHeader.Serialize(memoryPool.Memory.Span).GetReader();
+
+            // Assert
+            Assert.Equal(0, serializedResponse.Read()); //no uri
+            Assert.Equal(0, serializedResponse.Read()); //no uri
+            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cred
+            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cred
```
atifaziz, line 58: Those `//no ...` comments are great, but would be good to align with distinct names used in the protocol documentation (here and in other tests in the same class):

```suggestion
            Assert.Equal(0, serializedResponse.Read()); // no cupsUri
            Assert.Equal(0, serializedResponse.Read()); // no tcUri
            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); // no cupsCred
            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); // no tcCred
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            var serializedResponse = updateResponseHeader.Serialize(memoryPool.Memory.Span).GetReader();
+
+            // Assert
+            Assert.Equal(updateUriString.Length, serializedResponse.Read());
+            Assert.Equal(updateUriString, serializedResponse.ReadUtf8String(updateUriString.Length));
+            Assert.Equal(0, serializedResponse.Read());
+            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cred
+            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cred
+            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian());
+            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian());
```
atifaziz, line 86: Suggest to add comments next to each 0-asserted length about help understand which field is being talked about:

```suggestion
            Assert.Equal(0, serializedResponse.Read()); //no tcUri
            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cupsCred
            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no tcCred
            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian()); // no sig + keyCRC
            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian()); // no updData
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            var serializedResponse = updateResponseHeader.Serialize(memoryPool.Memory.Span).GetReader();
+
+            // Assert
+            Assert.Equal(0, serializedResponse.Read());
+            Assert.Equal(updateUriString.Length, serializedResponse.Read());
+            Assert.Equal(updateUriString, serializedResponse.ReadUtf8String(updateUriString.Length));
+            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cred
+            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cred
+            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian());
+            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian());
```
atifaziz, line 110:
```suggestion
            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no cupsCred
            Assert.Equal(0, serializedResponse.ReadUInt16LittleEndian()); //no tcCred
            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian()); // no sig + keyCRC
            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian()); // no updData
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            };
+
+            using var memoryPool = MemoryPool<byte>.Shared.Rent(2048);
+
+            // Act
+            var serializedResponse = updateResponseHeader.Serialize(memoryPool.Memory.Span).GetReader();
+
+            // Assert
+            Assert.Equal(0, serializedResponse.Read());
+            Assert.Equal(0, serializedResponse.Read());
```
atifaziz, line 131:
```suggestion
            Assert.Equal(0, serializedResponse.Read()); //no cupsUri
            Assert.Equal(0, serializedResponse.Read()); //no tcUri
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            Assert.Equal(credentialBytes.Length, serializedResponse.ReadUInt16LittleEndian()); //cups cred length
+            Assert.Equal(credentialBytes, serializedResponse.Read(credentialBytes.Length)); //cups cred
+            Assert.Equal(credentialBytes.Length, serializedResponse.ReadUInt16LittleEndian()); //tc cred length
+            Assert.Equal(credentialBytes, serializedResponse.Read(credentialBytes.Length)); //tc cred
+            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian());
+            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian());
```
atifaziz, line 137:
```suggestion
            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian()); // no sig + keyCRC
            Assert.Equal(0U, serializedResponse.ReadUInt32LittleEndian()); // no updData
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                UpdateDataLength = 0,
+            };
+
+            using var memoryPool = MemoryPool<byte>.Shared.Rent(2048);
+
+            // Act
+            var serializedResponse = updateResponseHeader.Serialize(memoryPool.Memory.Span);
+
+            // Assert
+            Assert.True(serializedResponse.ToArray().All(b => b == 0));
```
atifaziz, line 32: I would assert here in all other tests whether the response was the expected size:

```suggestion
            Assert.Equal(14, serializedResponse.Length);
            Assert.True(serializedResponse.ToArray().All(b => b == 0));
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        [Fact]
+        public void Serialize_WithNoUpdates()
+        {
+            var updateResponseHeader = new CupsUpdateInfoResponseHeader
+            {
+                SignatureKeyCrc = 0,
+                UpdateDataLength = 0,
+            };
+
+            using var memoryPool = MemoryPool<byte>.Shared.Rent(2048);
```
atifaziz, line 26: Consider renaming `memoryPool` to `memoryRental` since it's not the pool but the rented memory from the pool.

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
atifaziz: I would initialise the rented memory with some non-zero/unexpected filler:

```c#
memoryPool.Memory.Span.Fill(0xbd);
```

Suppose also we have a method on `ByteSpanReader` that reads the rest into a returned array:

```c#
public byte[] ReadAll() => Read(this.span.Length);
```

Then you can assert that `Serialize` didn't write _too much_ and remaining bytes in the rented memory are still intact:

```c#
var reader = serializedResponse.GetReader();
Assert.True(reader.ReadAll().All(b => b == 0));
Assert.True(memoryPool.Memory.Span[serializedResponse.Length..].ToArray().All(b => b == 0xbd));
```

I would do this for all other tests too.

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+    using System;
+    using System.Buffers;
+    using System.Collections.Immutable;
+    using System.Linq;
+    using LoRaWan.NetworkServer.BasicsStation;
+    using Xunit;
+
+    public class CupsUpdateInfoResponseHeaderTests
+    {
+        private readonly string updateUriString = "https://localhost:1234";
```
atifaziz, line 15: This can be a `const`:

```c#
        private const string UpdateUriString = "https://localhost:1234";
```

Tests/Unit/NetworkServer/LoRaDeviceApiServiceTests.cs
```diff
@@ -124,6 +126,39 @@ public async Task SearchByEuiAsync_DevEui_Parses_Json(string json, string expect
...
+            using var httpHandlerMock = new HttpMessageHandlerMock();
+            httpHandlerMock.SetupHandler(r =>
+            {
+                var response = new HttpResponseMessage(System.Net.HttpStatusCode.OK)
+                {
+                    Content = new StreamContent(memorystream),
+
+                };
+                return response;
+            } );
```
atifaziz, line 147: You can simplify by returning a `ByteArrayContent` in the response from the mock:

```suggestion
            var contentBytes = Encoding.UTF8.GetBytes("Test");
            // arrange
            var facadeMock = new Mock<IServiceFacadeHttpClientProvider>();
            using var httpHandlerMock = new HttpMessageHandlerMock();
            httpHandlerMock.SetupHandler(_ => new HttpResponseMessage(System.Net.HttpStatusCode.OK)
            {
                Content = new ByteArrayContent(contentBytes),
            });
```

atifaziz, line 147: You can avoid the `MemoryStream` by returning a `ByteArrayContent` in the response from the mock:

```suggestion
            // arrange
            var facadeMock = new Mock<IServiceFacadeHttpClientProvider>();
            using var httpHandlerMock = new HttpMessageHandlerMock();
            httpHandlerMock.SetupHandler(_ => new HttpResponseMessage(System.Net.HttpStatusCode.OK)
            {
                Content = new ByteArrayContent(contentBytes),
            });
```

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
atifaziz: I would even skip the memory pool and just create an array (`new byte[2048]`). This is a test and we don't care about memory pooling.

Tests/Unit/NetworkServer/CupsProtocolMessageProcessorTests.cs
```diff
@@ -110,21 +118,167 @@ public async Task HandleUpdateInfoAsync_Fails_WithInvalidInput(string input, Typ
...
+                                                KeyChecksum,
+                                                "ABCD");
+            _ = this.basicsStationConfigurationService.Setup(m => m.GetCupsConfigAsync(It.IsAny<StationEui>(), It.IsAny<CancellationToken>()))
+                                                      .Returns(Task.FromResult(cupsTwinInfo));
+
+            var firmwareBytes = new byte[] { 1, 2, 3 };
+            using var fwStream = new MemoryStream(firmwareBytes);
+            using var streamContent = new StreamContent(fwStream);
+            _ = this.deviceAPIServiceBase.Setup(m => m.FetchStationFirmwareAsync(It.IsAny<StationEui>(), It.IsAny<CancellationToken>()))
+                                         .ReturnsAsync(streamContent);
```
atifaziz, line 147: You can use a `ByteArrayContent` and skip the memory stream entirely (ditto for other tests):

```suggestion
            using var httpContent = new ByteArrayContent(firmwareBytes);
            _ = this.deviceAPIServiceBase.Setup(m => m.FetchStationFirmwareAsync(It.IsAny<StationEui>(), It.IsAny<CancellationToken>()))
                                         .ReturnsAsync(httpContent);
```

Tests/Unit/NetworkServer/CupsProtocolMessageProcessorTests.cs
```diff
@@ -110,21 +118,167 @@ public async Task HandleUpdateInfoAsync_Fails_WithInvalidInput(string input, Typ
...
+            // act
+            await this.processor.HandleUpdateInfoAsync(httpContext.Object, default);
+
+            // assert
+            this.deviceAPIServiceBase.Verify(m => m.FetchStationFirmwareAsync(StationEui.Parse(StationEuiString), It.IsAny<CancellationToken>()), Times.Once);
+            var cupsBytes = new byte[memoryStream.Position];
+            memoryStream.Seek(0, SeekOrigin.Begin);
+            await memoryStream.ReadAsync(cupsBytes);
+            Assert.Equal(firmwareBytes, cupsBytes[^3..]);
```
atifaziz, line 156: You can just grab the array of the memory stream and boil this down to just one line:

```suggestion
            Assert.Equal(firmwareBytes, memoryStream.ToArray().TakeLast(firmwareBytes.Length));
```

However, this does not test the beginning of the response, that the serialized `CupsUpdateInfoResponseHeader`? My fear is that if someone introduces a regression in `CupsProtocolMessageProcessor` then this will only test the firmware bytes were delivered but not that the LBS will accept the entire response in case the header is corrupt.

Tests/Unit/NetworkServer/CupsProtocolMessageProcessorTests.cs
```diff
@@ -110,21 +118,167 @@ public async Task HandleUpdateInfoAsync_Fails_WithInvalidInput(string input, Typ
...
+            using var httpContent = new StringContent("firmware");
+            httpContent.Headers.ContentLength = contentLength;
+
+            _ = this.deviceAPIServiceBase.Setup(m => m.FetchStationFirmwareAsync(It.IsAny<StationEui>(), It.IsAny<CancellationToken>()))
+                                         .ReturnsAsync(httpContent);
+            _ = this.basicsStationConfigurationService.Setup(m => m.GetCupsConfigAsync(It.IsAny<StationEui>(), It.IsAny<CancellationToken>()))
+                                                      .Returns(Task.FromResult(cupsTwinInfo));
+            // act
+            var exception = await Assert.ThrowsAsync<InvalidOperationException>(async () => await this.processor.HandleUpdateInfoAsync(httpContext.Object, default));
+            Assert.Contains("firmware", exception.Message, StringComparison.OrdinalIgnoreCase);
```
atifaziz, line 215: Don't think this is good enough as many of the exception message contain the word firmware. When I checked, you're actually getting the message, “Remote firmware signature generated with a key whose checksum is not available in CUPS request.” I don't think that's what's expected based on the title of the test.

atifaziz (approved):
I've left a few comments about some of the tests that would be best addressed, but otherwise LGTM. Great stuff!

Tests/Unit/NetworkServer/CupsProtocolMessageProcessorTests.cs
danigian: Fixed and added additional assertions

Tests/Unit/NetworkServer/CupsUpdateInfoResponseHeaderTests.cs
danigian: Renamed, but did not get rid of the array creation

Issue #1039, opened 2021-12-10
LBS is retrying to reconnect too fast when an identity is not fetched from the Iot Hub


Mandur:
When we are failing to find a device for a concentrator, the LBS retry to connect at an amazingly fast speed, generating requests that we forward to the iot hub at a DDOS like speed. We should avoid that.

```
2021-12-10T00:16:22.2669664Z 00:16:22.266 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RO" received FIN.
2021-12-10T00:16:22.2673977Z 00:16:22.266 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RP" received FIN.
2021-12-10T00:16:22.2677559Z 00:16:22.266 [TCPLOG] [itestarm1] Failed to authenticate HTTPS connection.
2021-12-10T00:16:22.2681058Z 00:16:22.266 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RP" stopped.
2021-12-10T00:16:22.2687870Z 00:16:22.266 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RP" sending FIN because: "The Socket transport's send loop completed gracefully."
2021-12-10T00:16:22.2692580Z 00:16:22.266 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RQ" accepted.
2021-12-10T00:16:22.2696149Z 00:16:22.266 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RQ" started.
2021-12-10T00:16:22.2699667Z 00:16:22.266 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RQ" received FIN.
2021-12-10T00:16:22.2703710Z 00:16:22.266 [TCPLOG] [itestarm1] Failed to authenticate HTTPS connection.
2021-12-10T00:16:22.2707396Z 00:16:22.267 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RQ" stopped.
2021-12-10T00:16:22.2714831Z 00:16:22.267 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RQ" sending FIN because: "The Socket transport's send loop completed gracefully."
2021-12-10T00:16:22.2719382Z 00:16:22.267 [TCPLOG] [itestarm1] Connection id "0HMDR90QAB1RR" accepted.
[...]

PR #1301, opened 2022-01-18
Introduces rate-limiting so that we don't contact IoTHub too often

spygi:
# PR for issue #1039

## What is being addressed

When the network server is not able to find the concentrator twin, LBS keeps retrying immediately. For each of these retries we contact IoTHub.

## How is this addressed

When the twin resolution does not work we introduce an empty entry in the cache with a very short absolute expiration time (15 seconds). The next time LBS tries to connect it will be a cache hit and we will not contact IoTHub, until the expiration time comes. The empty cache entry will cause all subsequent code to fail which is correct since the twin could not be successfully retrieved.


danigian (commented):
LGTM, but I would also like to hear opinions from @atifaziz or @p-schuler if they have another way to do this

spygi:
@atifaziz do you think you will have time for this?

atifaziz:
> The empty cache entry will cause all subsequent code to fail which is correct since the twin could not be successfully retrieved.

Does this mean that we'll open the floodgates on logging of errors/failures?

spygi:
> Does this mean that we'll open the floodgates on logging of errors/failures?

Potentially we will log more errors, yes. I think this is the weakest point of this solution but since we can't control how often LBS tries to re-connect not sure we can do something about it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
```diff
@@ -60,20 +62,40 @@ private async Task<TwinCollection> GetTwinDesiredPropertiesAsync(StationEui stat
...
+                                                                LoRaProcessingErrorCode.InvalidDeviceConfiguration);
+                        }
+
+                        using var client = this.loRaDeviceFactory.CreateDeviceClient(stationEui.ToString(), key);
+                        var twin = await client.GetTwinAsync(cancellationToken);
+                        return twin.Properties.Desired;
                     }
+                    catch (LoRaProcessingException ex)
+                    {
+                        exception = ex; // keeps the exception to re-throw after creating the cache entry
```
atifaziz, line 85: The better way to capture exceptions for re-throw from _elsewhere_ is to use `ExceptionDispatchInfo` to capture the exception and have it re-throw, which has the effect of preserving the stack trace and more accurately representing the original throw-site.

I would recommend updating as follows:

```diff
diff --git a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
index d15901f6..63413458 100644
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationConfigurationService.cs
@@ -4,6 +4,7 @@
 namespace LoRaWan.NetworkServer.BasicsStation
 {
     using System;
+    using System.Runtime.ExceptionServices;
     using System.Text.Json;
     using System.Threading;
     using System.Threading.Tasks;
@@ -62,7 +63,7 @@ namespace LoRaWan.NetworkServer.BasicsStation
 
             try
             {
-                LoRaProcessingException exception = null;
+                ExceptionDispatchInfo edi = null;
                 var cacheValue = await this.cache.GetOrCreateAsync(cacheKey, async cacheEntry =>
                 {
                     try
@@ -82,18 +83,17 @@ namespace LoRaWan.NetworkServer.BasicsStation
                     }
                     catch (LoRaProcessingException ex)
                     {
-                        exception = ex; // keeps the exception to re-throw after creating the cache entry
+                        edi = ExceptionDispatchInfo.Capture(ex); // keeps the exception to re-throw after creating the cache entry
 
                         _ = cacheEntry.SetAbsoluteExpiration(CacheLbsRateLimitingTimeout);
-                        return new TwinCollection(); // dummy cache entries are used in order to "rate limit" LBS so that we don't contact IoTHub too often 
+                        return new TwinCollection(); // dummy cache entries are used in order to "rate limit" LBS so that we don't contact IoTHub too often
                     }
                 });
 
 #pragma warning disable CA1508 // Avoid dead conditional code
                 // false positive, variable is conditionally not null when an exception is thrown
-                if (exception != null)
+                edi?.Throw();
 #pragma warning restore CA1508 // Avoid dead conditional code
-                    throw exception;
 
                 return cacheValue;
             }
```

You can copy [the above diff](https://github.com/Azure/iotedge-lorawan-starterkit/files/7932150/diff.txt) to the clipboard and apply it from PowerShell using: `Get-Clipboard | git apply`

See the [**Remarks** section of `ExceptionDispatchInfo`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo?view=net-6.0#remarks) for more background:

> The [`ExceptionDispatchInfo`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo?view=net-6.0) object stores the stack trace information and Watson information that the exception contains at the point where it is captured. The exception can be thrown at another time and possibly on another thread by calling the [`ExceptionDispatchInfo.Throw`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.throw?view=net-6.0) method. The exception is thrown as if it had flowed from the point where it was captured to the point where the [`Throw`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.throw?view=net-6.0) method is called.

atifaziz:
> Potentially we will log more errors, yes.

So we'll either thrash the IoT Hub or logging. Aren't we just kicking the can down the road with this?

> since we can't control how often LBS tries to re-connect

Generally, I've seen the LBS employ a back-off strategy so I am surprised that it's bombarding the LNS. I am assuming we reproduced the problem and tested manually that it goes away with the proposed fix of a dummy cache entry?

p-schuler:
> > Does this mean that we'll open the floodgates on logging of errors/failures?
> 
> Potentially we will log more errors, yes. I think this is the weakest point of this solution but since we can't control how often LBS tries to re-connect not sure we can do something about it.

Seems like another issue? There could be several problems arising from this. Can't we simply indicate this was throttled and only raise a metrics rather than having exceptions / logs being raised? If we hit the logs at the rate we saw the requests coming in, we potentially

1. raise cost on the app insights (if app insights is connected)
2. miss other log entries - due to throttling on the logging system
3. make it hard to read the logs

atifaziz:
> Generally, I've seen the LBS employ a back-off strategy so I am surprised that it's bombarding the LNS.

So after a debugging session with @spygi and help from @danigian, we managed to reproduce the original problem and found that the back-off is applied in all cases (INFO and MUXS) _except_ CUPS. The retry seems to be every second:

    2022-01-25 11:32:23.092 [CUP:INFO] Starting a CUPS session in 0 seconds.
    2022-01-25 11:32:23.092 [CUP:INFO] Starting a CUPS session now.
    2022-01-25 11:32:23.092 [CUP:ERRO] No CUPS-bak URI configured
    2022-01-25 11:32:24.093 [CUP:INFO] Starting a CUPS session in 0 seconds.
    2022-01-25 11:32:24.093 [CUP:INFO] Starting a CUPS session now.
    2022-01-25 11:32:24.093 [CUP:ERRO] No CUPS-boot URI configured
    2022-01-25 11:32:25.094 [CUP:INFO] Starting a CUPS session in 0 seconds.
    2022-01-25 11:32:25.094 [CUP:INFO] Starting a CUPS session now.
    2022-01-25 11:32:25.094 [CUP:INFO] Connecting to CUPS ... https://192.168.1.1:5002 (try #11)
    2022-01-25 11:32:25.095 [any:INFO] ./cups.trust:
    cert. version     : 1
    serial number     : 01
    issuer name       : OU=StarterKit, O=StarterKit, C=CH, CN=root Root CA
    subject name      : OU=StarterKit, O=StarterKit, C=CH, CN=root Root CA
    issued  on        : 2022-01-25 10:51:08
    expires on        : 2023-01-25 10:51:08
    signed using      : ECDSA with SHA256
    EC key size       : 256 bits
    2022-01-25 11:32:25.095 [any:INFO] ./cups.crt:
    cert. version     : 1
    serial number     : 01
    issuer name       : OU=StarterKit, O=StarterKit, C=CH, CN=root Root CA
    subject name      : OU=StarterKit, O=StarterKit, C=CH, CN=AA555A0000000101
    issued  on        : 2022-01-25 11:18:33
    expires on        : 2023-01-25 11:18:33
    signed using      : ECDSA with SHA256
    EC key size       : 256 bits
    2022-01-25 11:32:25.095 [AIO:INFO]
    2022-01-25 11:32:25.101 [CUP:VERB] Retrieving update-info from CUPS https://192.168.1.1:5002...
    2022-01-25 11:32:25.323 [AIO:XDEB] [3] http_read state=4
    2022-01-25 11:32:25.685 [AIO:XDEB] [3] http_read state=4
    2022-01-25 11:32:25.685 [AIO:DEBU] [3] Connection closed unexpectedly
    2022-01-25 11:32:25.685 [AIO:ERRO] [3] Error reading HTTP Header
    2022-01-25 11:32:25.685 [AIO:DEBU] [3] HTTP connection shutdown...
    2022-01-25 11:32:25.685 [CUP:INFO] Interaction with CUPS failed - retrying in 1s

In fact [the source of the LBS even has an interesting comment in `cups.c` about retrying quickly](https://github.com/lorabasics/basicstation/blob/42d4b9c7dd132ca4126b24726f57efb7c382e9a1/src/cups.c#L110-L111):

```c
        // Someting went wrong - retry again more quickly
        msg = "Interaction with CUPS failed%s - retrying in %~T";
```

On the LNS side, this manifested as:

    Hosting starting
    Hosting started
    Loaded hosting startup assembly LoRaWan.NetworkServer
    Hosting environment: Production
    Content root path: P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWanNetworkSrvModule\bin\Debug\net6.0
    Now listening on: https://0.0.0.0:5001
    Now listening on: https://0.0.0.0:5002
    Connection id "0HMEVV2O82Q7P" accepted.
    Connection id "0HMEVV2O82Q7P" started.
    AA555A0000000101 An exception occurred while processing requests: LoRaWan.LoRaProcessingException: The configuration request of station 'AA555A0000000101' did not match any configuration in IoT Hub. If you expect this connection request to succeed, make sure to provision the Basics Station in the device registry.
    ErrorCode: InvalidDeviceConfiguration
    at LoRaWan.NetworkServer.BasicsStation.BasicsStationConfigurationService.GetTwinDesiredPropertiesAsync(StationEui stationEui, CancellationToken cancellationToken) in P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWan.NetworkServer\BasicsStation\BasicsStationConfigurationService.cs:line 96
    at LoRaWan.NetworkServer.BasicsStation.BasicsStationConfigurationService.GetAllowedClientThumbprintsAsync(StationEui stationEui, CancellationToken cancellationToken) in P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWan.NetworkServer\BasicsStation\BasicsStationConfigurationService.cs:line 123
    at LoRaWan.NetworkServer.BasicsStation.ClientCertificateValidatorService.ValidateAsync(X509Certificate2 certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors, CancellationToken token) in P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWan.NetworkServer\BasicsStation\ClientCertificateValidatorService.cs:line 60
    .
    Unhandled exception while processing 0HMEVV2O82Q7P.
    Connection id "0HMEVV2O82Q7P" stopped.
    Connection id "0HMEVV2O82Q7P" sending FIN because: "The Socket transport's send loop completed gracefully."

atifaziz:
@danigian, @spygi The following message in the LBS log was bothering me:

    Interaction with CUPS failed - retrying in 1s

So when I went looking for [the defaults in the LBS code, I saw that the CUPS retry is set to 1 minute](https://github.com/lorabasics/basicstation/blob/v2.0.5/src/s2conf.h#L173):

```c
CONF_PARAM(CUPS_RESYNC_INTV    , ustime, tspan_m ,             "\"1m\"", "check-in with CUPS for updates after a failure")
```

I dumped the configuration parameters on the LBS on my Raspberry Pi using:

    docker exec LoRaBasicsStation ./station.std --params

and you can see that `CUPS_RESYNC_INTV` in the output is configured to 1 second via a `station.conf` file instead:

    str RADIODEV             = "/dev/spidev?.0" builtin      default radio device
        u4 LOGFILE_SIZE         = "10MB"     builtin      default size of a logfile
        u4 LOGFILE_ROTATE       = 3          builtin      besides current log file keep *.1..N (none if 0)
        u4 TCP_KEEPALIVE_EN     = 1          builtin      TCP keepalive enabled
        u4 TCP_KEEPALIVE_IDLE   = 60         builtin      TCP keepalive TCP_KEEPIDLE [s]
        u4 TCP_KEEPALIVE_INTVL  = 15         builtin      TCP keepalive TCP_KEEPINTVL [s]
        u4 TCP_KEEPALIVE_CNT    = 4          builtin      TCP keepalive TCP_KEEPCNT
        u4 MAX_JOINEUI_RANGES   = 10         builtin      max ranges to suppress unwanted join requests
    ustime CUPS_CONN_TIMEOUT    = "60s"      builtin      connection timeout
    ustime CUPS_OKSYNC_INTV     = "24h"      builtin      regular check-in with CUPS for updates
    ustime CUPS_RESYNC_INTV     = "1s"       station.conf check-in with CUPS for updates after a failure
        u4 CUPS_BUFSZ           = "8KB"      builtin      read from CUPS in chunks of this size
    ustime GPS_REPORT_DELAY     = "120s"     builtin      delay GPS reports and consolidate
    ustime GPS_REOPEN_TTY_INTV  = "1s"       builtin      recheck TTY open if it failed
    ustime GPS_REOPEN_FIFO_INTV = "1s"       builtin      recheck if FIFO writer fake GPS
    ustime CMD_REOPEN_FIFO_INTV = "1s"       builtin      recheck if FIFO writer
    ustime RX_POLL_INTV         = "20ms"     builtin      interval to poll SX1301 RX FIFO
    ustime TC_TIMEOUT           = "60s"      builtin      reconnected to muxs
    ustime CLASS_C_BACKOFF_BY   = "100ms"    builtin      retry interval for class C TX attempts
        u4 CLASS_C_BACKOFF_MAX  = 10         builtin      max number of class C TX attempts
    ustime RADIO_INIT_WAIT      = "200ms"    builtin      max wait for radio init command to finish
    ustime PPS_VALID_INTV       = "10m"      builtin      max age of last PPS sync for GPS time conversions
    ustime TIMESYNC_RADIO_INTV  = "2100ms"   builtin      interval to resync MCU/SX1301
    ustime TIMESYNC_LNS_RETRY   = "71ms"     builtin      resend timesync message to server
    ustime TIMESYNC_LNS_PAUSE   = "5s"       builtin      pause after unsuccessful volley of timesync messages
        u4 TIMESYNC_LNS_BURST   = 10         builtin      volley of timesync messages before pausing
    ustime TIMESYNC_REPORTS     = "5m"       builtin      report interval for current timesync status
    ustime TX_MIN_GAP           = "10ms"     builtin      min distance between two frames being TXed
    ustime TX_AIM_GAP           = "20ms"     builtin      aim for this TX lead time, if delayed should not fall under min
    ustime TX_MAX_AHEAD         = "600s"     builtin      maximum time message can be scheduled into the future
    ustime TXCHECK_FUDGE        = "5ms"      builtin      check radio state this time into ongoing TX
    ustime BEACON_INTVL         = "128s"     builtin      beaconing interval
        u4 TLS_SNI              = false      env          Set and verify server name of TLS connections

So I think the problem comes from our `station.conf` that sets `CUPS_RESYNC_INTV` to 1 second:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/0ef97d38974616b58d3ced9a6b5f4059238067ae/LoRaEngine/modules/LoRaBasicsStationModule/station.conf#L22-L29

Since the LBS default is 1 minute, we shouldn't expect aggressive retries out of the box (unless someone configures otherwise). I think the right action here is to simply fix our `station.conf` rather than add a hack to the code. If we want to generally add rate-limiting rather than rely on the LBS back-offs to be properly configured, we should address that as a separate (and probably larger) story.

---
PS I don't know why we ever chose to set `CUPS_RESYNC_INTV` to 1 seconds.

danigian:
> 

thanks Atif for finding this out, I personally never realized that this 'CUPS_RESYNC_INTV' was there in our station.conf file 😮 

I agree with what you propose, I created #1389 and I also think we should just abandon this PR and create a new story in vNext for setting up a better rate-limit feature

p-schuler:
> > 
> 
> thanks Atif for finding this out, I personally never realized that this 'CUPS_RESYNC_INTV' was there in our station.conf file 😮
> 
> I agree with what you propose, I created #1389 and I also think we should just abandon this PR and create a new story in vNext for setting up a better rate-limit feature

I'm not sure we should not have a mechanism in place to defend against this still? That only means if the LBS is configured correctly it should not misbehave. 

I don't think we can make the assumption that an LBS will never be misconfigured - we do not own that and the LBS is definitely an external consumer to our service we are providing.

spygi:
Thanks everyone for the follow-ups. 

I agree that ideally we should also have some rate-limiting since LBS is an external component and therefore lies outside our system boundary. The question is how much effort we want to spend there. Let's maybe re-discuss with @p-schuler as well?

spygi:
The original issue turned out to be a configuration problem and was fixed in #1391 
We decided to abandon this PR, due to the extra logged exceptions that would be caused by it and create a backlog item for vNext #1416

Issue #1305, opened 2022-01-18
`TestDeviceInfo` doesn't use `TwinProperty`


atifaziz:
## Expected Behavior

Expected all code getting or setting twin properties to use the key string constants defined in `TwinProperty`.

## Current Behavior

`TestDeviceInfo` uses `nameof(...)` on its own properties as twin keys. Unfortunately, this could lead to device twin reading errors during tests because twin keys are case-sensitive. For example, using `nameof` on `TestDeviceInfo.AppEui` will be incorrect because `TwinProperty.AppEui` defines the key as `"AppEUI"` (note that EUI is all caps).

PR #1306, opened 2022-01-18
Fix `TestDeviceInfo` to use keys from `TwinProperty`

atifaziz:
# PR for issue #1305

## What is being addressed

`TestDeviceInfo` used its property names for twin keys. Twin keys are case-sensitive and therefore the twin can be misconfigured if `TestDeviceInfo` properties get renamed or differ in case (e.g. `AppEui` vs `AppEUI`).

## How is this addressed

`TestDeviceInfo` is updated to use keys from `TwinProperty` exclusively.


bastbu (approved):
Fixed this as well in #1290, makes sense to merge this separately though.

Issue #1308, opened 2022-01-18
Error MSB3030 building device provisioning CLI tool


atifaziz:
## Expected Behavior

Expected the build of the device provisioning CLI tool to succeed:

    dotnet build Tools/Cli-LoRa-Device-Provisioning

## Current Behavior

Build of device provisioning CLI tool fails with the following error:

    Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET
    Copyright (C) Microsoft Corporation. All rights reserved.

    Determining projects to restore...
    All projects are up-to-date for restore.
    C:\Program Files\dotnet\sdk\6.0.101\Microsoft.Common.CurrentVersion.targets(5100,5): error MSB3030: Could not copy the file "P:\dev\Tools\Cli-LoRa-Device-Provisioning\appsettings.local.json" because it was not found. [P:\dev\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.csproj]

    Build FAILED.

    C:\Program Files\dotnet\sdk\6.0.101\Microsoft.Common.CurrentVersion.targets(5100,5): error MSB3030: Could not copy the file "P:\dev\Tools\Cli-LoRa-Device-Provisioning\appsettings.local.json" because it was not found. [P:\dev\Tools\Cli-LoRa-Device-Provisioning\Cli-LoRa-Device-Provisioning.csproj]
        0 Warning(s)
        1 Error(s)

## Steps to Reproduce

1. Clone the repo.
2. Change current shell directory to clone working directory.
2. Checkout commit 4490b3287bdc166a1d6b728d504abfb0b1377a6f (where `dev` was at time of reporting)
3. Run `dotnet build Tools/Cli-LoRa-Device-Provisioning`

PR #1309, opened 2022-01-18
Fix MSB3030 error building device provisioning tool

atifaziz:
# PR for issue #1308

## What is being addressed

MSB3030 error building device provisioning tool due to missing when `appsettings.local.json` is missing locally.

## How is this addressed

Made the requirement of the file optional for building as it's also optional for configuration per line 33 of `ConfigurationHelper.cs`:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/4490b3287bdc166a1d6b728d504abfb0b1377a6f/Tools/Cli-LoRa-Device-Provisioning/Helpers/ConfigurationHelper.cs#L30-L34


danigian (commented):
As far as I understood, we were already including in the build something that might potentially include secrets from the Azure resources deployed.

What about just not including it?

atifaziz:
> As far as I understood, we were already including in the build something that might potentially include secrets from the Azure resources deployed.
> 
> What about just not including it?

Local settings files are ignored by Git (so not included):

https://github.com/Azure/iotedge-lorawan-starterkit/blob/9a08055e1ae0450c6791984ad0aecac86e77d026/.gitignore#L295-L303

The project so far has assumed that the file is always there on disk, but that won't be the case for a fresh clone. The solutions in the repo should build for something that is a run-time requirement. In fact, I would argue that `appsettings.json` should be excluded at run-time and `appsettings.local.json` should be required! Right now, if someone modifies `appsettings.json` and puts credentials there (because `appsettings.local.json` is not technically needed) then they can get accidentally committed into Git. This fix is about getting the solution to build again. We should open another issue if we want to review how and where the CLI tool picks up the settings to lower the risk of accidents. What do you think?

atifaziz:
BTW, just checked [our documentation](https://github.com/Azure/iotedge-lorawan-starterkit/blob/112d34826439841a204b8e6e09dc707cf070ba00/docs/tools/device-provisioning.md#setting-up) and it even recommends updating `appsettings.json`:

> [appsettings.json](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/Tools/Cli-LoRa-Device-Provisioning/appsettings.json) needs to be in the same directory as the cli-lora-device-provisioning binary (verifyloradevice.dll or verifyloradevice.exe).
>
> [appsettings.json](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/Tools/Cli-LoRa-Device-Provisioning/appsettings.json) needs to contain a connection string from the Azure IoT Hub you want to work with.

This is fine if someone didn't downloaded the sources, but I think there's a risk for accident if cloned.

atifaziz:
@danigian I've logged #1335 to deal with the issue you've raised.

Issue #1160, opened 2021-12-24
Simulation project restoration


Mandur:
- [x] Simulator should have a mechanism to onboard fake basics station (similarly as the normal devices used in tests)
- [x] Commented tests in the simulator projects should be restored.
- [x] Simulation project should be documented
- [x] #1224
- [x] #1221
- [x] #1223
- [x] #1253
- [x] #1254
- [x] #1256
- [x] #1323

PR #1316, opened 2022-01-19
Restore commented-out load tests

bastbu:
# PR for issue #1160

## What is being addressed

With this PR we restore (or remove) the commented-out load tests and ensure that they can run successfully in the CI.

## How is this addressed

We uncomment and refactor the existing load tests.
- The simulation test which tests the decoder we remove, since it does not add value over our E2E test and will only make the simulator setup harder when running locally
- The actual load test we uncomment and refactor - we homogenize the test with the helper methods of the other load tests that we have, and we remove assertions that we now cover through other assertions (e.g. we do not test the exact message any longer, but just that the IoT hub received N messages from a given device)
- We stabilize the assertions by allowing some receive windows to be missed based on a configurable parameter


Tests/Simulation/IntegrationTestFixtureSim.cs
```diff
@@ -95,69 +81,41 @@ public override void SetupTestDevices()
...
-            // Range of 10 OTAA devices from 3000 to 3009: Used for load testing
-            for (var deviceID = 3000; deviceID <= 3009; deviceID++)
-            {
-                this.deviceRange3000_10_OTAA.Add(
-                    new TestDeviceInfo
-                    {
-                        DeviceID = deviceID.ToString("0000000000000000", CultureInfo.InvariantCulture),
-                        AppEui = JoinEui.Parse(deviceID.ToString("0000000000000000", CultureInfo.InvariantCulture)),
-                        AppKey = GetAppKey(deviceID),
-                        GatewayID = gatewayID,
```
spygi, line 155: Not sure if it's an issue but with the new helper method `CreateOtaaDevice` no GatewayId is assigned.

Tests/Simulation/IntegrationTestFixtureSim.cs
```diff
@@ -95,69 +81,41 @@ public override void SetupTestDevices()
...
-                        IsIoTHubDevice = true,
-                        SensorDecoder = "DecoderValueSensor",
-                    });
-            }
-            */
+            for (var deviceId = 4000; deviceId < 4000 + Configuration.NumberOfLoadTestDevices; deviceId++)
+                this.deviceRange4000_OTAA_FullLoad.Add(CreateOtaaDevice(deviceId));
+
+            TestDeviceInfo CreateAbpDevice(int deviceId) =>
```
spygi, line 96: Did you consider reusing the methods `CreateABPDevice` and `CreateOTAADevice` from `TestDeviceInfo` here?

Tests/Simulation/SimulatedLoadTests.cs
```diff
@@ -119,10 +119,121 @@ async Task ActAsync(SimulatedDevice device, TimeSpan startOffset)
...
+                await Task.Delay(TimeSpan.FromMilliseconds(RandomNumberGenerator.GetInt32(10, 250)));
+                this.logger.LogInformation("[WARM-UP] {DeviceId}", device.LoRaDevice.DeviceID);
+                // warm-up messages should not contain the message fragment (we do not want to include them in the message count)
+                using var request = WaitableLoRaRequest.CreateWaitableRequest(device.CreateConfirmedDataUpMessage("foo"));
+                await device.SendDataMessageAsync(request);
+            }
+
+            // 2. ABP and OTAA devices send messages
+            const int initialMessageId = 1;
+            for (var messageId = initialMessageId; messageId < initialMessageId + scenarioMessagesPerDevice; ++messageId)
```
spygi, line 176: [nit] Is there a technical reason why you try to send messages for both ABP and OTAA devices in a single for-loop? I find that it complicates the parameters we need to validate (e.g. messagesBeforeConfirmed <= messagesBeforeJoin) and the logic (e.g. for OTAA if messageId < messagesBeforeJoin + initialMessageId we don't send anything)

Tests/Simulation/SimulatedLoadTests.cs
```diff
@@ -119,10 +119,121 @@ async Task ActAsync(SimulatedDevice device, TimeSpan startOffset)
...
+                        messagesBeforeJoin + initialMessageId => batch.Select(devices => JoinAsync(devices.Otaa)),
+                        > messagesBeforeJoin + initialMessageId => batch.Select(devices => SendUpstreamMessage(devices.Otaa, payload, messageId)),
+                        _ => Array.Empty<Task>()
+                    };
+
+                    await Task.WhenAll(abpTasks.Concat(otaaTasks));
+                    await Task.Delay(IntervalBetweenMessages);
+                }
+
+                static async Task SendUpstreamMessage(SimulatedDevice device, string payload, int messageId)
```
spygi, line 194: Not sure about this but method declarations should be cached, right? Asking because this is inside a for-loop.

spygi (commented):
LGTM - left some suggestions, waiting for approval as per our discussion.

Tests/Simulation/IntegrationTestFixtureSim.cs
bastbu: Eagle eyes :) It was intentional though, since for the multi-gateway scenarios we do not want to associate these devices with a gateway already.

Tests/Simulation/SimulatedLoadTests.cs
bastbu: In any case I would need to duplicate a loop (if I send ABP and OTAA separately it would need to be the loop over message IDs). Now, to make sure that I send OTAA and ABP for the same message ID at the same time, it needs to be in the same loop. We could argue whether it would be good enough to send OTAA and ABP after each other, but I decided to just draw a line somewhere and leave the old test as it was (with respect to its "flow").

No strong opinion here though, we can discuss whether the message ID is still a useful concept for that test at all.

Tests/Simulation/SimulatedLoadTests.cs
bastbu: Checked this in Sharplab, it is indeed converted to just a class method. I still moved the static local functions out of the loop, it is less distracting indeed.

Tests/Simulation/SimulatedLoadTests.cs
```diff
@@ -119,10 +119,121 @@ async Task ActAsync(SimulatedDevice device, TimeSpan startOffset)
...
 
+        // Scenario:
+        // - ABP devices
+        // - OTAA devices
+        // - Sending confirmed/unconfirmed messages
+        // - Goal: N devices in parallel based on configuration
+        [Fact]
+        public async Task Multiple_ABP_and_OTAA_Simulated_Devices_Confirmed()
+        {
+            // amount of messages to send per device (without warm-up phase)
```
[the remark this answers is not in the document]
bastbu: Valid point, done this in 6c71306eb99608263b5c4a034d2afde5698440a5

Tests/Simulation/IntegrationTestFixtureSim.cs
bastbu: I did, but to be honest I was not confident what I would break by reusing the mentioned methods. There are some specifics in there around net ids etc that I did not want to change as part of this PR, I did the least effort fix, which was extracting the duplicated code to a local function. Open to improve this if you have more insights about what the changes would imply.

Tests/Simulation/SimulatedLoadTests.cs
bastbu: As per our discussion, I get it now. Originally, I had it as you suggested, but that did not work as expected due to some details around materialization of the Enumerables

spygi: I see, in that case I think we can resolve, it was more of a nit comment in any case.

spygi (approved):

Issue #1253, opened 2022-01-12
Remove setters to make records (`JoinAcceptResponse`, `LnsVersionRequest`) immutable


Mandur:
see https://github.com/Azure/iotedge-lorawan-starterkit/pull/1153/files#r782399029 and https://github.com/Azure/iotedge-lorawan-starterkit/pull/1153/files#r780997634

PR #1317, opened 2022-01-19
Make simulator records immutable

bastbu:
# PR for issue #1253

## What is being addressed

With this PR we make the records that are used for conversion from and to payloads that are used by the simulator immutable.


Mandur (approved):

bastbu:
Waiting on #1290 before merging

Tests/Common/LnsVersionRequest.cs
```diff
@@ -8,25 +8,25 @@ namespace LoRaWan.Tests.Simulation.Models
     public sealed record LnsVersionRequest
```
atifaziz, line 8: This is serialized-only in one place so doesn't seem to warrant a whole separate record to maintain:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/4490b3287bdc166a1d6b728d504abfb0b1377a6f/Tests/Common/SimulatedBasicsStation.cs#L55

What's more, in two other cases we just serialize an anonymous version; here:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/4490b3287bdc166a1d6b728d504abfb0b1377a6f/Tests/Common/SimulatedBasicsStation.cs#L44

and here:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/4490b3287bdc166a1d6b728d504abfb0b1377a6f/Tests/Common/SimulatedBasicsStation.cs#L81-L102

So just thinking, why not remove this entirely and follow the same route for consistency?

Tests/Common/JoinAcceptResponse.cs
```diff
@@ -8,16 +8,19 @@ namespace LoRaWan.Tests.Common
     internal sealed record JoinAcceptResponse
```
atifaziz, line 8: I would consider moving this as a private record of `SimulatedDevice` where it's only used in one place. It avoids having a lot of stand-alone types floating at the root that are more private implementation details (for now) and which don't need the same design scrutiny applied. For example, you could define it simply as:

```c#
private sealed record JoinAcceptResponse(
    [property: JsonPropertyName("pdu")] string Pdu,
    [property: JsonPropertyName("DevEui")] string DevEuiString);
```

bastbu:
Successful load test: https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1718638427

Issue #1315, opened 2022-01-19
Reuse `Rx2ReceiveWindow` as `ReceiveWindow`


atifaziz:
`Rx2ReceiveWindow` is already defined as:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/ad545dcd93e456f5fbeda9ccb0fc5d170a1d9ee8/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/RX2ReceiveWindow.cs#L8

but the definition can be generalized so it can be reused to represent RX1 or RX2 more broadly, for example here in `DownlinkMessage` (and elsewhere):

https://github.com/Azure/iotedge-lorawan-starterkit/blob/ad545dcd93e456f5fbeda9ccb0fc5d170a1d9ee8/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/DownlinkMessage.cs#L30-L31

PR #1318, opened 2022-01-19
Reuse `RX2ReceiveWindow` as `ReceiveWindow`

atifaziz:
# PR for issue #1315

## What is being addressed

Coupling of data rate and frequency into a tuple when `RX2ReceiveWindow` already serves the same type so `RX2ReceiveWindow` can be generalised for broader reuse as `ReceiveWindow`.

## How is this addressed

`RX2ReceiveWindow` is renamed to `ReceiveWindow` (note that order element has been changed to data rate then frequency to borrow form the [LoRaWAN regional parameters document](https://lora-alliance.org/resource_hub/rp2-1-0-3-lorawan-regional-parameters/)) since it's not specific to RX1 or RX2.

Wherever data rate plus frequency was needed as a pair (as tuples in method signatures or separate properties in classes like `DownlinkMessage`), `ReceiveWindow` is used instead.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -262,17 +263,22 @@ internal async Task ProcessJoinRequestAsync(LoRaRequest request)
...
+                var rx1 = windowToUse is not ReceiveWindow2
+                        ? new ReceiveWindow(loraRegion.GetDownstreamDataRate(request.RadioMetadata.DataRate, loRaDevice.ReportedRX1DROffset), freq)
+                        : (ReceiveWindow?)null;
+
+                var rx2 = new ReceiveWindow(loraRegion.GetDownstreamRX2DataRate(this.configuration.Rx2DataRate, null, this.logger),
+                                            loraRegion.GetDownstreamRX2Freq(this.configuration.Rx2Frequency, this.logger));
+
+                var downlinkMessage = new DownlinkMessage(joinAcceptBytes,
+                                                          request.RadioMetadata.UpInfo.Xtime,
+                                                          rx1, rx2,
```
spygi, line 276:
```suggestion
                                                          rx1,
                                                          rx2,
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/ReceiveWindow.cs
```diff
@@ -5,5 +5,5 @@ namespace LoRaTools.Utils
 {
     using LoRaWan;
 
-    public record struct RX2ReceiveWindow(Hertz Frequency, DataRateIndex DataRate);
+    public record struct ReceiveWindow(DataRateIndex DataRate, Hertz Frequency);
```
spygi, line 8: [nit] The combination of `DataRateIndex` and `Frequency` is also used in `RadioMetadata` as properties. I think using the `ReceiveWindow` there wouldn't make sense from a semantics-perspective but just bringing it up in case you missed it and you think it makes sense.

spygi (approved):
LGTM, having a type rather than a tuple provides more context.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Utils/ReceiveWindow.cs
atifaziz: That's an interesting point, but as you said, while the information seems identical, the fact that we have _receive_ in `ReceiveWindow` makes it odd to use/see on the transmission (TX) side of things (thus semantically different?). Moreover, the [radio metadata comes from the LNS protocol](https://lora-developers.semtech.com/build/software/lora-basics/lora-basics-for-gateways/?url=tcproto.html#radio-metadata), but `ReceiveWindow` maps to specific mentions of _receive windows_ in the LoRaWAN specification. Perhaps this needs more thought, but I'd rather not over-generalise for now.

Issue #1319, opened 2022-01-19
`DeviceJoinInfo` is not always passed to methods on region


bastbu:
## Expected Behavior

To ensure that all methods on the `Region` work for all regions, we must make sure to always pass all necessary parameters. 

## Current Behavior

For example, the `DeviceJoinInfo` is required for `TryGetDownstreamChannelFrequency` in CN RP2. Making the parameter required shows that it is not passed from all the places, which results in some paths (e.g. https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs#L283) being broken for CN RP2. We should make such parameters mandatory to ensure that all regions are always supported.

This issue is related to #1313

Issue #1313, opened 2022-01-18
Device Join error - "failed to handle join request. Value cannot be null. (Parameter 'upstreamDataRate')" using 2.0.0 beta bits


gerfen:
## Expected Behavior

I'm in the process of testing the new 2.0.0 beta bits.  I have set up a network server on a Linux VM and I have configured a Milesite LG67 gateway as a basic station.  The basic station successfully connects to the network server [1].  I have added a LoRaWAN device to Azure Iot Hub.  When this device tries to join the network I see the error in the logs[2].

It feels like I'm missing some configuration for the basic station device.

## Current Behavior

The device should successfully join the network.

### Device (Host) Operating System

Network Server: Ubuntu 20.04
Basic Station:  Open WRT

### Architecture

AMD64

### Container Operating System

Linux containers

## Logs

[1] Basic station connecting
```
No PFX is set for the server side authentication. No need to trust any certificate.
No certifiate copied, therefore no update-ca-certificates needed.
Getting properties from module twin...
Hosting starting
Hosting started
Loaded hosting startup assembly LoRaWan.NetworkServer
Hosting environment: Production
Content root path: /app
Now listening on: http://0.0.0.0:5000
Connection id "0HMEQLS4F5VFD" accepted.
Connection id "0HMEQLS4F5VFD" started.
Request starting HTTP/1.1 GET http://20.112.16.171:5000/router-info - -
Wildcard detected, all requests with hosts will be allowed.
1 candidate(s) found for the request path '/router-info'
Request matched endpoint '/router-info HTTP: GET'
Executing endpoint '/router-info HTTP: GET'
WebSocket connection from 103.84.136.171 established
Received discovery request from: 24E124FFFEF0DE07
Executed endpoint '/router-info HTTP: GET'
Connection id "0HMEQLS4F5VFD" received FIN.
Connection id "0HMEQLS4F5VFD" sending FIN because: "The client closed the connection."
Connection id "0HMEQLS4F5VFD" disconnecting.
Request finished HTTP/1.1 GET http://20.112.16.171:5000/router-info - - - 101 - - 637.2482ms
Connection id "0HMEQLS4F5VFD" stopped.
Connection id "0HMEQLS4F5VFE" accepted.
Connection id "0HMEQLS4F5VFE" started.
Request starting HTTP/1.1 GET http://20.112.16.171:5000/router-data/24E124FFFEF0DE07 - -
[...]

PR #1320, opened 2022-01-19
Upstream data rate is not passed to region

bastbu:
# PR for issue #1313

## What is being addressed

With this PR we address the exceptions that were thrown in US915 when joining a device. The exceptions were caused by an optional parameter `upstreamDataRate` which is not optional.

## How is this addressed

We reproduce the bug with an integration test. By making the `upstreamDataRate` required, we had to fix a couple of occurrences where the parameter was not passed to the methods on the region. A follow-up bug (caused by the same problem) #1319 will be fixed as a separate issue.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DownlinkMessageBuilder.cs
```diff
@@ -79,21 +79,21 @@ internal static class DownlinkMessageBuilder
 
             if (receiveWindow is ReceiveWindow2)
             {
-                freq = loRaRegion.GetDownstreamRX2Freq(configuration.Rx2Frequency, logger, deviceJoinInfo);
```
MaggieSalak, line 82: [nit] This is technically related to bug #1319 but no need to revert since this will change anyway

bastbu: Agree, needed to fix it with the minimum changes possible since it broke my integration test.

MaggieSalak (approved):

PR #1321, opened 2022-01-19
Make DeviceJoinInfo a required parameter in region methods

MaggieSalak:
# PR for issue #1319 

## What is being addressed

`DeviceJoinInfo` was not always passed to Region methods which would cause errors for region where it's required (CN470RP2). 

## How is this addressed

- `DeviceJoinInfo` parameter is required in all applicable Region methods


bastbu:
Should we add integration tests that validate that the paths that were broken due to this bug do work?

Tests/Integration/ClassCIntegrationTests.cs
```diff
@@ -50,7 +66,7 @@ public async Task When_ABP_Sends_Upstream_Followed_By_DirectMethod_Should_Send_U
 
             var twin = simDevice.CreateABPTwin(reportedProperties: new Dictionary<string, object>
                 {
-                    { TwinProperty.Region, LoRaRegionType.EU868.ToString() }
+                    { TwinProperty.Region, regionType.ToString() }
```
bastbu, line 69: On L98, where we use `CreateWaitableRequest`, we always create the request for region EU868. I think we should use the injected lora region type consistently

bastbu (approved):

Issue #1263, opened 2022-01-13
Assertion failures in Moq callbacks do not fail the test


bastbu:
## Expected Behavior

All tests should fail if an assertion fails.

## Current Behavior

Tests that have assertions within `Callback` from Moq do not fail, even if the assertions within the callback fail. One example of such a test is `DeduplicationStrategyIntegrationTests.Validate_Dup_Message_Processing`. This can easily be reproduced by changing an `Assert.True` to an `Assert.False` inside a `Callback` and observe that the test still passes.

PR #1355, opened 2022-01-21
Fix test cases that never fail due to Moq callback usage

bastbu:
# PR for issue #1263

## What is being addressed

With this PR we get rid of all the assertions within Moq callbacks. We also fix the `DeduplicationStrategyIntegrationTests`, which was the only test which was broken due to the non-failing test assertions.


Tests/Integration/ClassCIntegrationTests.cs
```diff
@@ -173,10 +173,6 @@ public async Task When_OTAA_Join_Then_Sends_Upstream_DirectMethod_Should_Send_Do
                     savedAppSKey = AppSessionKey.Parse(t[TwinProperty.AppSKey].Value);
                     savedNwkSKey = NetworkSessionKey.Parse(t[TwinProperty.NwkSKey].Value);
                     savedDevAddr = t[TwinProperty.DevAddr];
-
-                    Assert.NotNull(savedAppSKey);
-                    Assert.NotNull(savedNwkSKey);
-                    Assert.NotEmpty(savedDevAddr);
```
spygi, line 179: A general comment about the issue just to double check that I understand it correctly: changing these assertions for me causes the test to fail as it should. I think the problem with these assertions is that they make it hard to debug failures because they are reported on the call site (i.e. Line 210 instead of here), right?

Tests/Integration/DeduplicationStrategyIntegrationTests.cs
```diff
@@ -26,111 +32,143 @@ public DeduplicationStrategyIntegrationTests(ITestOutputHelper testOutputHelper)
...
+                           var isDup = messageProcessed;
+                           messageProcessed = true;
+                           return new FunctionBundlerResult()
+                           {
+                               DeduplicationResult = new DeduplicationResult { IsDuplicate = isDup }
+                           };
+                       }
+                   });
+
+                api.Setup(x => x.NextFCntDownAsync(It.IsAny<DevEui>(), It.IsAny<uint>(), It.IsAny<uint>(), It.IsAny<string>()))
```
spygi, line 53: I don't think we need this setup in any of the modes since we are sending unconfirmed messages.

spygi (approved):

Tests/Integration/ClassCIntegrationTests.cs
bastbu: I was able to reproduce on my machine and the CI (run https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1744091670 on commit 310959f9ca653c7170e9ed3fe1a8f680f080feb2) that changing an assertion does not fail the test. Do we want to sync offline why it fails on your machine?

Tests/Integration/DeduplicationStrategyIntegrationTests.cs
bastbu: Good catch, refactored the test further in 734c7ae8f3edc7a3ee3effa347fb797ac1a00bb7

PR #1356, opened 2022-01-21
Make unused raw message a private detail of payload

atifaziz:
This PR removes the `RawMessage` array on `LoRaPayload` since there were no public accesses except one in `ConcentratorDeduplicationDataMessagesIntegrationTests` that has been fixed to use `Serialize` instead. The property now is a private field of the respective subclasses that need it. This takes care of [CA1819 (Properties should not return arrays)](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1819) and prevents the array from being modified without the knowledge of the payload class.

## Summary of changes

- `LoRaPayload.RawMessage` is removed and becomes a private field of `LoRaPayloadData` and `LoRaPayloadJoinAccept` where it was used.
- `LoRaPayload` constructors are removed, all of which are no longer required.
- `LoRaPayloadJoinAccept.GetByteMessage` that provided access to the raw message is inlined into its single place of use.
- `ConcentratorDeduplicationDataMessagesIntegrationTests` is updated to use the public `Serialze` method of `LoRaPayloadData` instead of accessing (sole case) the raw message.
- `ConcentratorDeduplicationDataMessagesIntegrationTests` test arrangement were updated to create distinct payload instances for the two requests used in tests because `Serialze` (unlike `RawMessage`) update the payload in place and would break tests since requests would share the same underlying payload object. Instead now, a factory is use to create each payload so modifications to one doesn't affect the other.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
```diff
@@ -134,8 +136,7 @@ public byte[] Serialize(AppKey appKey)
         {
             Mic = LoRaWan.Mic.ComputeForJoinAccept(appKey, MHdr, AppNonce, NetId, DevAddr, DlSettings, RxDelay, CfList);
             _ = PerformEncryption(appKey);
-
-            return GetByteMessage();
+            return this.rawMessage.Prepend((byte)MHdr).ToArray();
```
bastbu, line 139: Aren't we prepending the `MHdr` twice now?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
```diff
@@ -35,27 +27,6 @@ public abstract class LoRaPayload
...
-        /// Initializes a new instance of the <see cref="LoRaPayload"/> class.
-        /// Wrapper of a LoRa message, consisting of the MIC and MHDR, common to all LoRa messages
-        /// This is used for uplink / decoding.
-        /// </summary>
-        protected LoRaPayload(byte[] inputMessage)
-        {
-            RawMessage = inputMessage ?? throw new ArgumentNullException(nameof(inputMessage));
-            MHdr = new MacHeader(RawMessage[0]);
-            // MIC 4 last bytes
-            Mic = LoRaWan.Mic.Read(RawMessage.AsSpan(inputMessage.Length - 4, 4));
```
bastbu, line 48: I don't see where we do this in the new constructor. Why could we delete this?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
bastbu, line 139: Aren't we prepending the `MHdr` twice now?

atifaziz: I don't have a good answer for this right now as I haven't looked at the `rawMessage` preparation in detail. All that was done here was inlining of `GetByteMessage` where this was the sole remaining call, and `GetByteMessage` was always prepending the header byte. However, if we consider than it's being set as the first byte of `rawMessage` on line 54 (like before) then it does seem like a duplication.

atifaziz: So I took a closer look as I was wondering how all the tests were still passing and turns out that this worked because `PerformEncryption` rewrites `rawMessage` without the header and this adds it back. So it turns out, what read like a duplicate addition of the header byte, isn't actually one. This also means that the `rawMessage` setup in the constructor was never used since it was always overwritten by `PerformEncryption`. I've completely removed the need for `rawMessage` with a97c56b53ca8cb4c4a51098997d3defa93216eed. It's a more dramatic change than I was hoping to make here, but I think it makes sense to go with it as it reads cleaner (the return value from `PerformEncryption` also doesn't get discarded anymore. I am also tempted to inline the single use of `PerformEncryption` into `Serialize`. Thoughts?

atifaziz: By the way, just realized that the `rawMessage` setup in the constructor was never needed or used anymore so completely eliminated it with f5a2217e2d4e86c1ff87f3c48552b3983ea24085.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayload.cs
atifaziz: The MIC for a `LoraPayloadData` is given to its constructor so this was a redundant read for that subclass anyway. `LoRaPayloadJoinAccept` calculates the MIC in `Serialize`. The `Mic` property for a `LoRaPayloadJoinAccept` is never used anywhere in the code.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadJoinAccept.cs
atifaziz: > I am also tempted to inline the single use of `PerformEncryption` into `Serialize`.

Done with 9d9a0d211e8e2881474ed78443e4f43671b015c3 but can be reverted if you're not in favour of this.

bastbu (approved):
Would be good to run a fullci before merging, LGTM

Issue #1329, opened 2022-01-19
`ConcentratorCredentialsFunction`, `ConcentratorFirmwareFunction` shouldn't return 422


atifaziz:
## Expected Behavior

Expected `ConcentratorCredentialsFunction` and `ConcentratorFirmwareFunction` to return 500 when unable to read CUPS information from the twin or if the information is incorrect.

## Current Behavior

The functions returns a [422 (Unprocessable Entity)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422). MDN clearly warns that this is not designed for server problems because it requires the client to modify the request before re-submitting:

> The HyperText Transfer Protocol (HTTP) **422 Unprocessable Entity** response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.
>
> **Warning:** The client should not repeat this request without modification.

422 originated in WebDAV, [which says](https://datatracker.ietf.org/doc/html/rfc4918#section-11.2):

> The 422 (Unprocessable Entity) status code means the server understands the content type of the request entity (hence a    415(Unsupported Media Type) status code is inappropriate), and the syntax of the request entity is correct (thus a 400 (Bad Request) status code is inappropriate) but was unable to process the contained instructions.  For example, this error condition may occur if an XML request body contains well-formed (i.e., syntactically correct), but semantically erroneous, XML instructions.

So even in WebDAV, it's referring to the instructions in the _request entity_, not the _response entity_.

bastbu:
Bug also applies `ConcentratorCredentialFunction`

PR #1362, opened 2022-01-24
Concentrator functions shouldn't return 422

MaggieSalak:
# PR for issue #1329 

## What is being addressed

Concentrator functions shouldn't return `UnprocessableEntityResult` in case device twin property is missing or misconfigured.

## How is this addressed

- Both affected functions (`ConcentratorCredentialsFunction`, `ConcentratorFirmwareFunction`) return `500` in the described cases.
- Unit tests for both functions are updated.

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
@@ -5,6 +5,7 @@ namespace LoraKeysManagerFacade
 {
     using System;
     using System.IO;
+    using System.Net;
     using System.Security.Cryptography;
     using System.Threading;
     using System.Threading.Tasks;
@@ -102,7 +103,10 @@ namespace LoraKeysManagerFacade
                                               or InvalidOperationException)
                 {
                     this.logger.LogError(ex, "'{PropertyName}' desired property was not found or misconfigured.", CupsPropertyName);
-                    return new UnprocessableEntityResult();
+                    return new ObjectResult($"Failed to fetch '{CupsPropertyName}' property from device twin")
+                    {
+                        StatusCode = (int)HttpStatusCode.InternalServerError,
+                    };
                 }
             }
             else
--- a/Tests/Unit/LoraKeysManagerFacade/ConcentratorCredentialTests.cs
+++ b/Tests/Unit/LoraKeysManagerFacade/ConcentratorCredentialTests.cs
@@ -146,6 +146,35 @@ namespace LoRaWan.Tests.Unit.LoraKeysManagerFacade
             Assert.IsType<BadRequestObjectResult>(result);
         }
 
+        [Fact]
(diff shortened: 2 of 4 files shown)


Tests/Unit/LoraKeysManagerFacade/ConcentratorFirmwareFunctionTests.cs
```diff
@@ -137,7 +137,7 @@ public async Task RunFetchConcentratorFirmware_Returns_BadRequest_ForMissingQuer
         }
 
         [Fact]
-        public async Task RunFetchConcentratorFirmware_Returns_UnprocessableEntityResult_ForTwinMissingCups()
+        public async Task RunFetchConcentratorFirmware_Returns_InternalServerErros_ForTwinMissingCups()
```
bastbu, line 140:
```suggestion
        public async Task RunFetchConcentratorFirmware_Returns_InternalServerError_ForTwinMissingCups()
```

Tests/Unit/LoraKeysManagerFacade/ConcentratorCredentialTests.cs
```diff
@@ -146,6 +146,35 @@ public async Task RunFetchConcentratorCredentials_Returns_BadRequest_ForMissingQ
...
 
+        [Fact]
+        public async Task RunFetchConcentratorCredentials_Returns_InternalServerError_ForMissingCupsProperty()
+        {
+            var blobBytes = Encoding.UTF8.GetBytes(RawStringContent);
+            using var blobStream = new MemoryStream(blobBytes);
+            SetupBlobMock(blobStream);
+
+            // http request
+            var httpRequest = new Mock<HttpRequest>();
```
bastbu, line 157: Consider reusing the existing logic in the test for setting up the request and the twin instead of copying the existing `arrange` part of other tests.

LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
```diff
@@ -102,7 +103,10 @@ or InvalidCastException
                                               or InvalidOperationException)
                 {
                     this.logger.LogError(ex, "'{PropertyName}' desired property was not found or misconfigured.", CupsPropertyName);
-                    return new UnprocessableEntityResult();
+                    return new ObjectResult($"Failed to fetch '{CupsPropertyName}' property from device twin")
```
bastbu, line 106: By changing the type to `ObjectResult` we lose observability compared to before on the LNS side. For example, in `LoRaDeviceAPIService.FetchStationCredentialsAsync` we only log the `ReasonPhrase` (= Internal Server Error) and the Status Code (= 500), which does not tell us much about what went wrong. I would argue that we should also change the log statements on the client side.

bastbu (approved):

LoRaEngine/LoraKeysManagerFacade/ConcentratorCredentialsFunction.cs
MaggieSalak: Updated the log messages for Credentials and Firmware function usage.

PR #1365, opened 2022-01-24
[docs] Add images for observability scenarios

roel4ez:
This PR adds two images to clarify the observability scenarios.


docs/user-guide/observability.md
```diff
@@ -19,6 +21,8 @@ Generate a deployment manifest from [`deployment_observability.layered.template.
 
 ## Integrating with the Elastic stack
 
+![Observability with ELK](../images/observability-elk.png){ width="450" }
```
bastbu, line 24: [nit] for the file beat you can also consider whether you want to indicate that logs can be scraped from the edgeAgent and edgeHub as well

bastbu (approved):

docs/user-guide/observability.md
roel4ez: fair point, but it will clutter the image. However to anyone familiar with Beats it should be clear that it can scrape _any_ logs

Issue #1281, opened 2022-01-14
(Some) MAC commands are sent downstream with reversed bytes


atifaziz:
## Expected Behavior

The `ABP_Unconfirmed_Sends_Valid_Mac_Commands_In_Fopts_And_Reply_In_Fopts` test in `ProcessingTests` sends a `LinkCheckReq` MAC command via `LinkCheckRequest` and expects a `LinkCheckAns` via `LinkCheckAnswer`. The downstream message has 21 bytes laid out as follows:

    [0] = 96
    [1] = 1 
    [2] = 0 
    [3] = 0 
    [4] = 2 
    [5] = 3 
    [6] = 1 
    [7] = 0 
    [8] = 1 
    [9] = 15
    [10] = 2 
    [11] = 1 
    [12] = 42
    [13] = 154
    [14] = 243
    [15] = 239
    [16] = 125
    [17] = 85
    [18] = 216
    [19] = 5 
    [20] = 188

At offset 5, the `FOptsLen` correctly indicates 3 bytes of `FOpts`, which is supposed to be the MAC command in the offsets range 8..10 (inclusive). At offset 8, the command identifier should be 2 for `LinkCheckAns`.

## Current Behavior

The command identifier is incorrect and instead can be found at offset 10. The MAC commands bytes seemed to be written in the reverse order.

The test passes nonetheless because the `LinkCheckAnswer` constructor used in the test for assertion interprets the bytes in the reverse order:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/f19bd7b5d8c060d77d9a30f938276de561f2cbca/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkCheckAnswer.cs#L38-L43

## Steps to Reproduce

1. Put a breakpoint on line 306 of `ProcessingTests`:
   https://github.com/Azure/iotedge-lorawan-starterkit/blob/f19bd7b5d8c060d77d9a30f938276de561f2cbca/Tests/Integration/ProcessingTests.cs#L306
1. Run the `ABP_Unconfirmed_Sends_Valid_Mac_Commands_In_Fopts_And_Reply_In_Fopts` test in `ProcessingTests`.
1. Examine the contents of `request.ResponseDownlink.Data` memory when the breakpoint hits. Note that bytes from offset 8 to 10 are inverted.

danigian:
[...]

PR #1371, opened 2022-01-24
Fix reversed bytes of (some) MAC commands

atifaziz:
# PR for issue #1281

## What is being addressed

Bytes of some MAC commands were reversed.

## How is this addressed

All MAC commands now have their bytes written in the natural order (first to last).

Other changes made to align some properties of MAC commands with actual types (avoids casts later):

- Property `Frequency` of `RXParamSetupRequest` was changed from `byte[]` to `int` (actually 24-bit and validated).
- Property `Freq` of `NewChannelRequest` was changed from `byte[]` to `int` (actually 24-bit and validated).
- Properties `Margin` and `Gwcnt` of `LinkCheckAnswer` were changed from `uint` to `byte`.

### DIFF
--- a/Tests/Integration/CloudToDeviceMessageSizeLimitShouldAcceptTests.cs
+++ b/Tests/Integration/CloudToDeviceMessageSizeLimitShouldAcceptTests.cs
@@ -22,7 +22,7 @@ namespace LoRaWan.Tests.Integration
     {
         public CloudToDeviceMessageSizeLimitShouldAcceptTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper) { }
 
-        [Theory(Skip = "Fails on CI - works locally. To enable with #562")]
+        [Theory(/*Skip = "Fails on CI - works locally. To enable with #562"*/)]
         [CombinatorialData]
         public async Task Should_Accept(
             bool isConfirmed,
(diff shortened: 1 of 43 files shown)


Tests/Integration/CloudToDeviceMessageSizeLimitShouldAcceptTests.cs
```diff
@@ -141,7 +141,7 @@ public class CloudToDeviceMessageSizeLimitShouldAcceptTests : CloudToDeviceMessa
             if (expectedMacCommandsCount > 0)
             {
                 // Possible problem: Manually casting payloadDataDown.Fopts to array and reversing it
```
danigian, line 143: Remove the comment

danigian (commented):
Looks great. Before approving, I would like to do some manual tests with real devices (even though I see the Mac E2E tests pass)

Tests/Integration/CloudToDeviceMessageSizeLimitShouldAcceptTests.cs
atifaziz: Done with 4b057e1b4542dcc991592714f6fefe59a6577295.

danigian (approved):
E2E passed, plus I enjoyed seeing how flawless the test of LinkCheckReq, DutyCycleReq, DevStatusReq, RXTimingSetupReq mac commands was.

PR #1382, opened 2022-01-25
Ensure that `IotHubDeviceInfo` is serialized correctly

bastbu:
# PR for issue #1149

## What is being addressed

With this PR we ensure that the `IotHubDeviceInfo` is (de-)serialized correctly.

### DIFF
--- a/Tests/Unit/NetworkServer/IotHubDeviceInfoConversionTests.cs
+++ b/Tests/Unit/NetworkServer/IotHubDeviceInfoConversionTests.cs
@@ -12,7 +12,7 @@ namespace LoRaWan.Tests.Unit.NetworkServer
     public sealed class IotHubDeviceInfoConversionTests
     {
         [Fact]
-        public void Serialization_And_Deserialization_Preserves_Information_Using_JsonNet()
+        public void Can_Be_Json_Deserialized_As_IoTHubDeviceInfo_From_NetworkServer()
         {
             // arrange
             var original = new IoTHubDeviceInfo { DevAddr = new DevAddr(123), DevEUI = new DevEui(234), PrimaryKey = "someprimarykey" };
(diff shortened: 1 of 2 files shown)


Tests/Unit/NetworkServer/IotHubDeviceInfoConversionTests.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace LoRaWan.Tests.Unit.NetworkServer
+{
+    using global::LoraKeysManagerFacade;
+    using Newtonsoft.Json;
+    using Xunit;
+
+    public sealed class IotHubDeviceInfoConversionTests
+    {
+        [Fact]
+        public void Serialization_And_Deserialization_Preserves_Information_Using_JsonNet()
```
atifaziz, line 15: Since it gets deserialized into a _different_ type, I would recommend to reflect that in the test fact:

```suggestion
        public void Can_Be_Json_Deserialized_As_IoTHubDeviceInfo_From_NetworkServer()
```

atifaziz (approved):

Tests/Unit/NetworkServer/IotHubDeviceInfoConversionTests.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+namespace LoRaWan.Tests.Unit.NetworkServer
+{
+    using global::LoraKeysManagerFacade;
+    using Newtonsoft.Json;
```
atifaziz, line 9: Seems an import was missing that was needed for my [earlier suggestion](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1382/files#r791650734) to compile:

```suggestion
    using Newtonsoft.Json;
    using Newtonsoft.Json.Linq;
```

Sorry.

Issue #1377, opened 2022-01-24
Device Provisioning tool - document `rotate-certificate` and `revoke` commands 


MaggieSalak:
## Description

Device Provisioning CLI tool [page ](https://azure.github.io/iotedge-lorawan-starterkit/dev/tools/device-provisioning/#setting-up) is currently missing documentation of `rotate-certificate` and `revoke` commands.

PR #1383, opened 2022-01-25
CLI tool - documentation for `rotate-certificate` and `revoke` commands

MaggieSalak:
# PR for issue #1377 

## What is being addressed

There was no documentation in the CLI Device Provisioning tool user guide for `rotate-certificate` and `revoke` commands which are both related to client certificate management for the Basics Station.

## How is this addressed

- The two commands and their parameters are documented.

### DIFF
--- a/docs/tools/device-provisioning.md
+++ b/docs/tools/device-provisioning.md
@@ -62,13 +62,15 @@ The following verbs/commands are supported:
 
 |verb|description|
 |-|-|
-|list|Lits devices in IoT Hub.|
+|list|List devices in IoT Hub.|
 |query|Query a device twin.|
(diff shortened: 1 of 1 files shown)


docs/tools/device-provisioning.md
```diff
@@ -257,6 +257,48 @@ The query verb supports the following parameters:
 |--help|no|Display this help screen.|
 |--version|no|Display version information.|
 
+### rotate-certificate
```
roel4ez, line 260: also add these two command to the table on line 63.
(there is also a small spelling mistake on line 65 (`Lits` instead of `List`), but I cannot comment on those lines from here.

roel4ez (approved):

Issue #1389, opened 2022-01-25
Remove CUPS_RESYNC_INTV from default station.conf


danigian:
Currently a `CUPS_RESYNC_INTV` of 1s is set in `station.conf`:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/0ef97d38974616b58d3ced9a6b5f4059238067ae/LoRaEngine/modules/LoRaBasicsStationModule/station.conf#L22-L29

The value should not be there so [the default of 1 minute](https://github.com/lorabasics/basicstation/blob/v2.0.5/src/s2conf.h#L173) gets used.

PR #1391, opened 2022-01-26
Removes custom CUPS retry config so that the default of 1 minute will be used

spygi:
# PR for issue #1389

## What is being addressed

In our station.conf we previously set CUPS retries to happen on every second.

## How is this addressed

By removing the custom setting - the LBS will use the default value of 60 seconds.

This was manually verified with Atif: by removing the setting in station.conf and re-starting the basics module, the default value of 60 seconds was picked up.


atifaziz:
@spygi Doesn't this address #1039 instead of #1392?

atifaziz (approved):
LGTM

---
PS Doesn't this address #1039 instead of #1392? Also we might want to link to the issue too?

Issue #1392, opened 2022-01-26
`IsCurrentDownstreamDRIndexWithinAcceptableValue` shouldn't accept nullables


MaggieSalak:
While looking at bug #1390 we saw the message `0018B20000020E90 the provided RX2DataRate is not valid` which will be logged when the data rate passed to `IsCurrentDownstreamDRIndexWithinAcceptableValue` is null. The method compares the DR index with the allowed range and as such should not accept null as parameter.

Issue #1390, opened 2022-01-26
Inconsistent Join behavior


gerfen:
I've been testing out the new basics station functionality and I'm having inconsistent results with join requests.  I have deployed the network Server module to a Linux VM hosted on Azure and I have deployed the LBS module to an AAEON AIOT-IP68 gateway to act as a basics station.  I am trying to connect an Adeuis network tester to the network.

## Expected Behavior

A device should be able to rejoin the network after getting disconnected.

## Current Behavior

After an initial join, devices struggle to get reconnected.  The join requests show the JoinAccept messages at the network server but the devices fail to join repeatedly.

## Steps to Reproduce

1. Set a up a device in Azure IoT hub
2. See the initial join request and see the data get decoded and forwarded to Azure IoT hub
3. Cycle the power on the device
4. Watch the device struggle to rejoin successfully via the network server logs.

### Device (Host) Operating System

Network Server - Ubuntu 20.04
Basics Station - Ubuntu 20.04

### Architecture

AMD64 for both NS and LBS

### Container Operating System

Linux containers.

#### Docker

Network Server  - 20.10.11+azure-3
LBS -  20.10.7+azure

## Logs

Network Server logs
```
0018B20000020E90 join request received
0018B20000020E90 querying the registry for OTAA device
0018B20000020E90 updating twin
0018B20000020E90 twin updated
0018B20000020E90 the provided RX2DataRate is not valid
0018B20000020E90 Device registered in cache.
0018B20000020E90 previous device devAddr (02723CE7) removed from cache.
0018B20000020E90 using standard region RX2 datarate DR8
0018B20000020E90 using standard region RX2 frequency 923300000
0018B20000020E90 sending message to station with EUI '024299FFFED2F3E4' with diid -2045090691. Payload '2085CC36887C7F531095A1BF1BE0CDE04E'.
0018B20000020E90 processing time: 00:00:00.1638424
[...]

PR #1393, opened 2022-01-26
`IsCurrentDownstreamDRIndexWithinAcceptableValue` should not accept null

MaggieSalak:
# PR for issue #1392 

## What is being addressed

`IsCurrentDownstreamDRIndexWithinAcceptableValue()` shouldn't accept null as parameter.

## How it is being addressed 

The error message "0018B20000020E90 the provided RX2DataRate is not valid" which was spotted in #1390 is logged when `DesiredRX2DataRate` is null. In this case calling `IsCurrentDownstreamDRIndexWithinAcceptableValue()` doesn't make sense, as the method simply compares the data rate index with the allowed range. This PR changes the method to not accept nullable `DataRateIndex`.

Note that this issue does not appear to be the cause of errors reported in the mentioned bug and because of this a separate bug has been created for this.


spygi (approved):
Approving - I have some doubts whether this is an improvement in the code as we need to move the null checks outside the method in the callers but I also don't have all the context of the discussions.

Issue #1384, opened 2022-01-25
[docs] Deploy button in quickstart is not rendering correctly


roel4ez:
and it should point to the version tag and not the latest master

PR #1396, opened 2022-01-26
Getting rid of mixed content issue in docs and pointing to v2.0.0 tag

danigian:
# PR for issue #1384 

# What changed
- [x] Getting rid of mixed content issue by pointing to different https "button" image
- [x] Updating to v2.0.0 tag the link for deployment (it still does not work, will work when new release is released)


roel4ez (approved):
🐿️

Issue #1326, opened 2022-01-19
Don't start a testrun before the modules are restarted


roel4ez:
Related to #1262 

We should ensure the modules are restarted (triggered by whatever reason) before any testrun is scheduled since this could lead to intermittent test failures

PR #1397, opened 2022-01-26
Wait with E2E test execution until newest version LNS has started

bastbu:
# PR for issue #1326

## What is being addressed

With this PR we introduce a mechanism that sets the current LNS version as a reported property in IoT Hub. When deploying the LNS to the gateways in CI, we wait until the property has been reported as an indicator that the correct version of the LNS is running (takes ca. 10 minutes, depending on the gateway).


bastbu:
Successful run at https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/1751023693

danigian (approved):

PR #1401, opened 2022-01-26
Track device twin load requests

MaggieSalak:
# PR for issue #1186 

## What is being addressed

Device twin load requests should be tracked in a dedicated metric

## How is this addressed

- New `TwinLoadRequests` metric has been added
- Twin load requests are tracked in `LoRaDeviceClient`


Tests/Unit/NetworkServer/LoRaDeviceFactoryTest.cs
```diff
@@ -13,8 +14,9 @@ namespace LoRaWan.Tests.Unit.NetworkServer
     using Moq.Protected;
     using Xunit;
 
-    public class LoRaDeviceFactoryTest
+    public sealed class LoRaDeviceFactoryTest : IDisposable
     {
+        private readonly Meter meter = new Meter(MetricRegistry.Namespace, MetricRegistry.Version);
```
bastbu, line 19: [nit] Depending on the metric, we might get interference from other tests if we use the metric with the same registry as in other tests. Typically, therefore we use `TestMeter.Instance` for tests since we do not listen on these metrics anywhere.

Issue #1307, opened 2022-01-18
Workbooks fleet overview is broken


bastbu:
## Expected Behavior

The IoT Edge fleet overview curated workbook seems to have changed between what's available in IoT Hub under "workbooks", and what's mentioned [in the documentation](https://docs.microsoft.com/en-us/azure/iot-edge/how-to-explore-curated-visualizations?view=iotedge-2020-11&tabs=devices%2Cmessaging#iot-edge-fleet-view-workbook). We should make sure that our workbook is up to date and has all the necessary device details. 

## Current Behavior

When clicking a device in the workbook fleet overview, it does not select the IoT hub resource and hence it does not display any metrics. This is because it does not correctly pass the owner ID of the workbook to the template.

## Steps to Reproduce

Open the E2E environment workbook.

Instead of trying to fix the current workbook, we could also consider using the updated version of the IoT Edge fleet overview.

PR #1402, opened 2022-01-26
Replace nested workbook with template reference to community workbook

bastbu:
# PR for issue #1307

## What is being addressed

With this PR we fix the broken workbook that is deployed with the starter kit. While the workbook previously worked, there seem to have been some changes to the community workbooks which broke our workbook. With this PR we make the dependency on the community workbooks more stable by referencing it in a template group. The updated community workbooks section will look this (in a fresh environment):

image


danigian (approved):
Much cleaner, thanks!

Issue #1334, opened 2022-01-20
Resubmit threshold does not consider number of concentrators connected


spygi:
## Expected Behavior

In the following scenario:
- Deduplication Mark or None
- A device is sending a message that is picked up from 4 concentrators connected to the same network server.

For Mark and None we expect all 4 messages to make it upstream.

## Current Behavior

The first 3 messages make it upstream, after that the LNS starts dropping messages due to `resubmit from confirmed message exceeds threshold of 3`.

## Severity 

Depending on the setup could affect the current deployment.

spygi:
Possible solutions to be discussed:
- We store additional state in `LoRaDevice`: instead of having only a `confirmationResubmitCount` we store the resubmit count per stationEui and frame counter up. This way we can check for each station Eui that the threshold was not reached.
- We set the threshold dynamically depending on how many LBSs are connected. For this we would need to make use of the `WebSocketWriterRegistry`.

PR #1404, opened 2022-01-26
Allows messages to go upstream when more than 4 concentrators are used

spygi:
# PR for issue #1334

## What is being addressed

When more than 4 concentrators were connected to the same LNS, messages were dropped as they were considered a potential replay attack.

## How is this addressed

- By ensuring that only resubmissions (from the same station) count against the resubmit threshold.
- If we get a late resubmission (e.g. because the ConcentratorDeduplication in-memory cache entry expired or the network server crashed) we also count this message against the threshold.
- Integration test that will break if we change the threshold used in the code.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -849,7 +849,7 @@ protected virtual async Task<LoRaADRResult> PerformADR(LoRaRequest request, LoRa
             if (!isFrameCounterFromNewlyStartedDevice && payloadFcnt <= loRaDevice.FCntUp)
             {
                 // most probably we did not ack in time before or device lost the ack packet so we should continue but not send the msg to iothub
-                if (payload.RequiresConfirmation && payloadFcnt == loRaDevice.FCntUp)
+                if (payload.RequiresConfirmation && payloadFcnt == loRaDevice.FCntUp && (concentratorDeduplicationResult is ConcentratorDeduplicationResult.NotDuplicate or ConcentratorDeduplicationResult.DuplicateDueToResubmission))
```
spygi, line 852: @p-schuler we added the `ConcentratorDeduplicationResult.NotDuplicate` case in the condition to cover for the case of a late resubmission (when the cache entry is expired or the LNS crashed)

p-schuler: Makes sense!

Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
```diff
@@ -268,7 +269,35 @@ LoRaRequest CreateOTAARequest(LoRaPayloadData payload, string station)
...
             // act/assert
-            await ActAndAssert(request1, request2, this.loraABPDevice, null, expectedNumberOfBundlerCalls, null, expectedMessagesUp, expectedMessagesDown, null);
+            await ActAndAssert(requests, this.loraABPDevice, null, expectedNumberOfBundlerCalls, null, expectedMessagesUp, expectedMessagesDown, null);
+        }
+
+        [Theory]
+        [InlineData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", "33-33-33-33-33-33-33-33", "44-44-44-44-44-44-44-44", "55-55-55-55-55-55-55-55", DeduplicationMode.Mark, 5, 1)]
+        [InlineData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", "33-33-33-33-33-33-33-33", "44-44-44-44-44-44-44-44", "55-55-55-55-55-55-55-55", DeduplicationMode.None, 5, 1)]
+        public async Task When_More_Than_4_Concentrators_Messages_Should_Go_Upstream(
+            string station1,
```
p-schuler, line 279: if you change this to string[] stations, you can validate the count against the MaxConfirmationResubmitCount without generating the warning.

p-schuler (approved):
LGTM

Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
spygi: Makes sense, done in 9b9f7c7b62e30fe04a54451477f5078f3f46bf7e

Issue #1283, opened 2022-01-14
LNS incorrectly rejects data frames (`updf`) where `FRMPayload` is an empty string


atifaziz:
## Expected Behavior

When a data frame is has no frame payload, the LBS sends an empty string for the `FRMPayload` property (and an `FOpts` of -1; see #1282) of the data frame JSON object. The LNS should allow and ignore the empty string.

## Current Behavior

The LNS rejects the frame as being invalid due to the following argument validation in `LoRaPayloadDataLns`:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/f19bd7b5d8c060d77d9a30f938276de561f2cbca/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LoRaMessage/LoRaPayloadDataLns.cs#L32

Issue #1282, opened 2022-01-14
LNS incorrectly rejects data frames (`updf`) where `FPort` = -1


atifaziz:
## Expected Behavior

Expected LNS to handle data frames where `FPort` is -1 (meaning missing), as per the spec:

    {
        "msgtype"   : "updf",
        "MHdr"      : UINT,
        "DevAddr"   : INT32,
        "FCtrl"     : UINT,
        "FCnt",     : UINT,
        "FOpts"     : "HEX",
        "FPort"     : INT(-1..255),
        "FRMPayload": "HEX",
        "MIC"       : INT32,
        ..
    }

## Current Behavior

The LNS rejects the frame as being invalid with the message: _Received unexpected 'jreq' message: …."_

The problem seems to stem from the fact that the `Port` is read as an unsigned byte (see line 98):

https://github.com/Azure/iotedge-lorawan-starterkit/blob/f19bd7b5d8c060d77d9a30f938276de561f2cbca/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/JsonHandlers/LnsData.cs#L91-L106

danigian:
As additional detail, this scenario might only happen if FRMPayload is empty. FPort MUST be specified if FRMPayload is filled instead
![image](https://user-images.githubusercontent.com/1955514/149526516-4ce397ee-f3bf-4913-9900-003fdf0e1874.png)

atifaziz:
@danigian Right, so either:

- `FPort` is -1 and `FRMPayload` an empty string
- `FPort` is a byte (0…255) and `FRMPayload` a non-empty string whose length is always even (hexadecimal encoding so always two characters per byte)

PR #1409, opened 2022-01-27
Allow missing payload case: FPort = -1, FRMPayload = ""

atifaziz:
# PR for issue #1282 and #1283

## What is being addressed

When the LBS sends an uplink data frame that has no frame payload, the members `FPort` and `FRMPayload` of the JSON message object are set to -1 and an empty string, respectively. The LNS would reject such requests as invalid message when the specification allows it. This can happen, for example, when the data frame has only MAC commands sent in `FOpts`.

## How is this addressed

`FPort` is allowed to be -1 in addition to a byte from 0 to 255 and `FRMPayload` is allowed to be an empty string. The combination of the two marks denotes “no frame payload”.


danigian (changes requested):
Changes look good to me.

We only need to do one additional thing.
As soon as we are allowing a null FPort, we should change LoRaDeviceTelemetry to handle this possibility.

In its constructor, as of now, we have a upstreamPayload.Fport.Value that will throw when Fport is null.

Please change  LoRaDeviceTelemetry accordingly 👍

danigian:
FYI i added #1415 to track the introduction of a E2E Test to cover this scenario

danigian (approved):

Issue #1407, opened 2022-01-27
Make sure we have a valid connection for cache refreshes


p-schuler:
We added an automatic refresh of the cache entries to ensure we have them up to date. 

## Expected Behavior
When we refresh, we keep the connection busy so it can not be released when a `KeepAliveTimeout` is set.

## Current Behavior
The connection could be potentially closed while we refresh the twins.

PR #1410, opened 2022-01-27
Bug/1407 ensure connection on cache refresh

p-schuler:
# PR for issue #1407

## What is being addressed

When refreshing the device cache, we did not ensure the connection during the refresh. When we start using KeepAliveTimeout, it could interfere with the refresh.

## How is this addressed

Reserve the connection for the duration of the update.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
@@ -108,8 +108,8 @@ namespace LoRaWan.NetworkServer
         protected virtual async Task RefreshDeviceAsync(LoRaDevice device, CancellationToken cancellationToken)
         {
             _ = device ?? throw new ArgumentNullException(nameof(device));
-            using var activity = device.BeginDeviceClientConnectionActivity();
-            _ = await device.InitializeAsync(this.configuration, cancellationToken);
+            using (device.BeginDeviceClientConnectionActivity())
+                _ = await device.InitializeAsync(this.configuration, cancellationToken);
         }
 
         public virtual bool Remove(LoRaDevice device, bool dispose = true)
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
@@ -105,10 +105,11 @@ namespace LoRaWan.NetworkServer
             }
         }
 
-        protected virtual Task RefreshDeviceAsync(LoRaDevice device, CancellationToken cancellationToken)
+        protected virtual async Task RefreshDeviceAsync(LoRaDevice device, CancellationToken cancellationToken)
         {
             _ = device ?? throw new ArgumentNullException(nameof(device));
-            return device.InitializeAsync(this.configuration, cancellationToken);
+            using var activity = device.BeginDeviceClientConnectionActivity();
+            _ = await device.InitializeAsync(this.configuration, cancellationToken);
         }
 
         public virtual bool Remove(LoRaDevice device, bool dispose = true)
--- a/Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
+++ b/Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
@@ -23,7 +23,7 @@ namespace LoRaWan.Tests.Unit.NetworkServer
         {
             var moqCallback = new Mock<Action<LoRaDevice>>();
             using var cache = new TestDeviceCache(moqCallback.Object, this.quickRefreshOptions, true);
-            var deviceMock = new Mock<LoRaDevice>(new DevAddr(0xffffffff), new DevEui(0), Mock.Of<ILoRaDeviceClientConnectionManager>());
+            var deviceMock = CreateMockDevice();
             var device = deviceMock.Object;
             cache.Register(device);
             await cache.WaitForRefreshAsync(CancellationToken.None);
@@ -62,7 +62,7 @@ namespace LoRaWan.Tests.Unit.NetworkServer
         public async Task When_Disposed_While_Refreshing_We_Shutdown_Gracefully()
         {
             using var cache = new TestDeviceCache(this.quickRefreshOptions, true);
-            var deviceMock = new Mock<LoRaDevice>(new DevAddr(0xabc), new DevEui(0x123), null);
+            var deviceMock = CreateMockDevice();
             deviceMock.Setup(x => x.InitializeAsync(It.IsAny<NetworkServerConfiguration>(), It.IsAny<CancellationToken>()))
                 .ReturnsAsync((NetworkServerConfiguration config, CancellationToken token) =>
                 {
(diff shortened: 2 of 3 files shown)


Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
```diff
@@ -22,11 +22,13 @@ public class LoRaDeviceCacheTest
...
             var moqCallback = new Mock<Action<LoRaDevice>>();
-            using var cache = new TestDeviceCache(moqCallback.Object, this.quickRefreshOptions);
-            using var device = CreateTestDevice();
+            using var cache = new TestDeviceCache(moqCallback.Object, this.quickRefreshOptions, true);
+            var deviceMock = CreateMockDevice();
+            var device = deviceMock.Object;
             cache.Register(device);
             await cache.WaitForRefreshAsync(CancellationToken.None);
             moqCallback.Verify(x => x.Invoke(device));
+            deviceMock.Verify(x => x.BeginDeviceClientConnectionActivity(), Times.Once);
```
atifaziz, line 31: The one thing this doesn't test is that `RefreshDeviceAsync` does indeed dispose the object returned by `BeginDeviceClientConnectionActivity`. Otherwise regression can go unnoticed if some edits the `using` statement in the method.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceCache.cs
```diff
@@ -105,10 +105,11 @@ private async Task RefreshCacheAsync(CancellationToken cancellationToken)
             }
         }
 
-        protected virtual Task RefreshDeviceAsync(LoRaDevice device, CancellationToken cancellationToken)
+        protected virtual async Task RefreshDeviceAsync(LoRaDevice device, CancellationToken cancellationToken)
         {
             _ = device ?? throw new ArgumentNullException(nameof(device));
-            return device.InitializeAsync(this.configuration, cancellationToken);
+            using var activity = device.BeginDeviceClientConnectionActivity();
+            _ = await device.InitializeAsync(this.configuration, cancellationToken);
```
atifaziz, line 112: nit: since `activity` is unused and this `using` form doesn't allow discards, how about just going back to the good old `using` block with no variable assignment?

```suggestion
            using (device.BeginDeviceClientConnectionActivity())
                _ = await device.InitializeAsync(this.configuration, cancellationToken);
```

atifaziz (approved):
Apart from [my comment on an additional assertion we could use on `When_Device_Expires_It_Is_Refreshed`](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1410/files#r793680822), LGTM.

Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
p-schuler: Please feel free to add the test. @atifaziz

atifaziz: Done with afa2bb8da6c1a57cb2e6e98e55bad0ecccdd4e54.

Please note that I had to change `TestDeviceCache.RefreshDeviceAsync` to be more stable otherwise the test would fails sometimes. I believe this is because the (semaphore) signaling was done too early, before calling the base implementation. As a consequence, `WaitForRefreshAsync` could return early at times and the verifications of calls on mocks could fail.

p-schuler: Perfect! Thanks! Don't you think the analyzers should actually be taking that responsibility of ensuring IDisposable are called correctly? We do have rules for exactly that issue. I would assume you can't check in code that does not properly dispose. Otherwise we should fix the analyzers.

atifaziz: So if you're talking about rule [CA2000 (Dispose objects before losing scope)][ca2000] then what I've realized is that it only flags when `new`-ing disposable objects, but not when a method returns one.

I also dug into the [rule's source code](https://github.com/dotnet/roslyn-analyzers/blob/36bb3a560ae6e7ae11dfdbf9c2925c0112e6f2de/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs) to see if I could find out more. It seems that it does support inter-procedure analysis but that it gets skipped based on some conditions. There's even a [local function named `SkipInterproceduralAnalysis` in there](https://github.com/dotnet/roslyn-analyzers/blob/36bb3a560ae6e7ae11dfdbf9c2925c0112e6f2de/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs#L158-L186) with some interesting commentary:

```c#
// Skip interprocedural analysis if we are invoking a method and not passing any disposable object as an argument
// and not receiving a disposable object as a return value.
// We also check that we are not passing any object type argument which might hold disposable object
// and also check that we are not passing delegate type argument which can
// be a lambda or local function that has access to disposable object in current method's scope.
```

Based on my reading, it seems that it should be able to detect the case of a method returning a disposable object. I then looked at whether there's a configuration option to turn this on and seems that there's something called [`DisposeAnalysisKind`](https://github.com/dotnet/roslyn-analyzers/blob/086d1cc888ca0d8c96b94444935f488fcaf3fe98/src/Utilities/FlowAnalysis/Options/DisposeAnalysisKind.cs) (abridged version below):

```c#
// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.

enum DisposeAnalysisKind
{
    // NOTE: Below fields names are used in the .editorconfig specification
    //       for DisposeAnalysisKind option. Hence the names should *not* be modified,
    //       as that would be a breaking change for .editorconfig specification.

    AllPaths,
    AllPathsOnlyNotDisposed,
    NonExceptionPaths,
    NonExceptionPathsOnlyNotDisposed,
}
```

Clearly there's a mention of `.editorconfig` but there's nothing in the [official documentation][ca2000] that mentions those. After some more digging, I found [a document in the analyzers repo that mentions it](https://github.com/dotnet/roslyn-analyzers/blob/c4a3caa09006b26efa282bc94afcf09bef84c89a/docs/Analyzer%20Configuration.md#dispose-analysis-kind-for-idisposable-rules):

    dotnet_code_quality.dispose_analysis_kind = AllPaths

The default is `NonExceptionPaths` that's described as:

> Track and report missing dispose violations only on non-exception program paths. Additionally, also flag use of non-recommended dispose patterns that may cause potential dispose leaks.

Anyway, I added the following `.editorconfig` in `LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/.editorconfig`:

    [*.cs]
    dotnet_code_quality.interprocedural_analysis_kind = None
    dotnet_code_quality.dispose_analysis_kind = AllPaths

and proceeded to remove all `using` statements around `BeginDeviceClientConnectionActivity` invocations and it doesn't produce the CA2000 error. It does however produce CA2000 for 4 other cases so one can't say it's not picking up the configuration:

```
P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWan.NetworkServer\ServiceFacadeHttpClientHandler.cs(70,28,74,22): error CA2000: Object created by 'new HttpResponseMessage(HttpStatusCode.BadRequest)' is not disposed along all exception paths. Call System.IDisposable.Dispose on the object before all references to it are out of scope.
P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWan.NetworkServer\LoRaDeviceAPIServiceBase.cs(89,31,89,59): error CA2000: Object created by 'new ByteArrayContent(buffer)' is not disposed along all exception paths. Call System.IDisposable.Dispose on the object before all references to it are out of scope.
P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWan.NetworkServer\LoRaDeviceClientConnectionManager.cs(146,82,146,101): error CA2000: Object created by 'out var removedItem' is not disposed along all exception paths. Call System.IDisposable.Dispose on the object before all references to it are out of scope.
P:\dev\LoRaEngine\modules\LoRaWanNetworkSrvModule\LoRaWan.NetworkServer\LoRaDeviceClient.cs(268,22,268,108): error CA2000: Object created by 'DeviceClient.CreateFromConnectionString(this.connectionString, this.transportSettings)' is not disposed along all exception paths. Call System.IDisposable.Dispose on the object before all references to it are out of scope.
````

Again, I don't know why this is not in the [official documentation of the rule][ca2000], but I can't get it to work reliably and so I think it's best to have tests in place for assurance. It also protects us from potential bugs in the analyzer.

[ca2000]: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000

Issue #1411, opened 2022-01-27
Inconsistent DeviceClient creation


p-schuler:
When creating the DeviceClient, we do not employ a common factory method, but rather rely on 2 places where we initialize it. In case of connection closing, we do create it without retry policies attached.

## Expected Behavior
`DeviceClient` instances are always created and initialized with the same settings.

## Current Behavior
`DeviceClient` instances that are re-created are missing retry policy settings.

## Steps to Reproduce
1. Setup a KeepAliveTimeout on a device
2. Send a message with that device
3. Wait for the KeepAliveTimeout to expire
4. Send another message with that device

You should now have a `DeviceClient` without retry policies

PR #1412, opened 2022-01-27
Add factory method to create identical DeviceClient

p-schuler:
# PR for issue #1411

## What is being addressed

Device clients are created with different retry policies.

## How is this addressed

Added a factory method to ensure the same code is executed for re-creating the DeviceClient.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
@@ -37,11 +37,7 @@ namespace LoRaWan.NetworkServer
             this.connectionString = connectionString;
             this.primaryKey = primaryKey;
             this.logger = logger;
-            var deviceClient = this.deviceClient = DeviceClient.CreateFromConnectionString(this.connectionString, this.transportSettings);
-            deviceClient.SetRetryPolicy(new ExponentialBackoff(int.MaxValue,
-                                                               minBackoff: TimeSpan.FromMilliseconds(100),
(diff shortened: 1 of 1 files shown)


atifaziz (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -261,6 +257,16 @@ public bool EnsureConnected()
             return true;
         }
 
+        private DeviceClient CreateDeviceClient()
+        {
+            var deviceClient = DeviceClient.CreateFromConnectionString(this.connectionString, this.transportSettings);
+            deviceClient.SetRetryPolicy(new ExponentialBackoff(int.MaxValue,
```
danigian, line 263: We realized that this retry policy makes no sense for two reasons:
1. it's equal to the default one
2. potentially specifies to retry for a huge amount of days, given the retry count set to int.maxvalue

Anyways, according to the documentation [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/main/iothub/device/devdoc/retrypolicy.md) the "retry mechanism will stop after DefaultOperationTimeoutInMilliseconds which is currently set at 4 minutes"

danigian (changes requested):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
p-schuler: Absolutely - let's tackle the validation of the policy separately. This is only to have a common place for initializing it, so we need to deal with it only in one place.

danigian (approved):

Issue #1006, opened 2021-12-07
[Story] Investigate usages of TXPowertoMaxEIRP


MaggieSalak:
## Description

`Region` class defines `TXPowertoMaxEIRP` dictionary which is populated with TXPower and EIRP values for specific regions. We discovered that the starterkit implementation only ever uses the keys of that dictionary, but the values (EIRP configuration) are never actually used. We should verify if the existing usages are correct and change if necessary.

- [ ] Adaptive Datarate calculation must be understood and documented.

MaggieSalak:
cc @Mandur

PR #1433, opened 2022-01-28
Improve documentation of ADR calculation

MaggieSalak:
# PR for issue #1006 

## What is being addressed

Documentation of the ADR calculation wasn't very detailed; the way the `TxPowerToMaxEIRP` table was used in the solution (selecting the highest TX power index in ADR calculation) wasn't documented.

## How is this addressed

- The documentation of ADR calculation is improved with more details of how the ADR algorithm works.


p-schuler (approved):
LGTM - would be good to create a separate section for the ADR part.

Issue #1432, opened 2022-01-28
Flaky Integration Test `Join_And_Send_Unconfirmed_And_Confirmed_Messages`


p-schuler:
I saw `Join_And_Send_Unconfirmed_And_Confirmed_Messages` fail in https://github.com/Azure/iotedge-lorawan-starterkit/runs/4976729164?check_suite_focus=true

## Expected Behavior
Test produces reproducible results.

## Current Behavior
It may fail.

The test failed with a DeduplicationDrop on the join message. The only way this can happen is if the concentrator does consider it a duplicate join. However, this should never happen here, as it is the first request. 

The only reason I see why this could happen is because the MemoryCache and the ConcentratorDeduplication are both static in the `MessageDispatcher`. To ensure a clean start for each test, these should be reset or initialized differently. 

btw: why is the cached named `testMemoryCache`?

https://github.com/Azure/iotedge-lorawan-starterkit/blob/9141b56ffc579b77ec0163b9764e45bc7f29d78a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs#L26-L27

PR #1438, opened 2022-01-28
Avoids sharing ConcentratorDeduplicator among tests

spygi:
# PR for issue #1432

## What is being addressed

Previously, `ConcentratorDeduplicator` was static and as tests run on the same AppDomain, they were sharing this one instance. That could cause intermittent failures (unexpected cache hits).

## How is this addressed

By not having the `ConcentratorDeduplicator` a static field and instead creating a new one each time a test needs a `MessageDispatcher`. We are using a static `Create` method so that we can first populate the `ConcentratorDeduplicator` and then re-use the constructor that the DI also uses.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
@@ -23,9 +23,6 @@ namespace LoRaWan.NetworkServer
         private readonly ILogger<MessageDispatcher> logger;
         private readonly Histogram<double> d2cMessageDeliveryLatencyHistogram;
 
-        private static readonly IMemoryCache testMemoryCache = new MemoryCache(new MemoryCacheOptions());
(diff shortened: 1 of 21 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MessageDispatcher.cs
```diff
@@ -50,17 +47,21 @@ public sealed class MessageDispatcher : IDisposable, IMessageDispatcher
         }
 
         /// <summary>
-        /// Use this constructor only for tests.
+        /// Use this only for tests.
```
atifaziz, line 50: If this is designed to be used from tests only then doesn't it belong in `Tests/Common/LoRaWan.Tests.Common.csproj` instead?

spygi: Moved it to its own type in Tests.Common with 9ea05dc05bad78a3a9df60fd16d4dc0990559098

atifaziz (approved):

Issue #672, opened 2021-10-31
HttpClient should be inject using DefaultHttpClientFactory


Mandur:
The implementation of the function caller instantiates Http client itself. Now that we have DI, we should use [DefaultHttpClientFactory](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests)

Additionally, the HttpClient should start using retry handling. Currently retries are not performed when talking to the function backend. For unconfirmed messages, it would be critical to try to deliver the message.

PR #1453, opened 2022-02-02
Use DI `IHttpClientFactory` and DI `HttpClient`

bastbu:
# PR for issue #672 

## What is being addressed

With this PR we use the `IHttpClientFactory` and DI `HttpClient` instead of instantiating them manually.

We add a retry policy to the `HttpClient` used in the `LoRaDeviceAPIService`, which retries aggressively first (due to the receive window limits), and then retries with more time in between.


kbeaugrand:
Hi, 

I'm coming after the battle, but I used to implement this feature in another way : 
I add the HttpClient factory by default to the DI and another httpClient factory specific for an implementation type that requires a client.

This gives me also the possibility to manage at the startup the policyHandler: 
```
            var transientHttpErrorPolicy = HttpPolicyExtensions
                                    .HandleTransientHttpError()
                                    .OrResult(c => c.StatusCode == HttpStatusCode.NotFound)
                                    .WaitAndRetryAsync(3, attempt => TimeSpan.FromMilliseconds(100));

            services.AddHttpClient("RestClient")
                .AddPolicyHandler(transientHttpErrorPolicy);

            services.AddHttpClient<ILoraDeviceMethodManager, LoraDeviceMethodManager>(client =>
            {
                client.BaseAddress = new Uri(configuration.LoRaKeyManagementUrl);
                client.DefaultRequestHeaders.Add("x-functions-key", configuration.LoRaKeyManagementCode);
            })
                .AddPolicyHandler(transientHttpErrorPolicy);
```

I guessed that this was a better way. What are your opinion ?

bastbu:
@kbeaugrand Thanks for your suggestion! I did not get to adding retry policies yet for the `LoRaDeviceAPIService` (which is where your suggestion would apply if I understood correctly). I added these in e7af10f5942aa0afa6571218e521e2bd111f85d3, such that we have this change in the PR now as well.

As for your suggestion - the main benefit I see from using typed clients is the increased type safety by not needing to rely on a `public constant string` as an identifier for the client. If I look at the implications of registering the `LoRaDeviceAPIService` as a typed client, it will mean that the `LoRaDeviceAPIService` becomes transient instead of the singleton it now is.

If you see the simplification in not needing to invoke the `CreateClient` helper method, I think we can also solve this by storing the `HttpClient` on the `LoRaDeviceAPIService` instance instead of the `IHttpClientFactory`. I did not make this change since I do not grasp all the implications this has with respect to how the underlying connections are managed. Are there potential problems from keeping the same `HttpClient` instance for the lifetime of the `LoRaDeviceAPIService`? I would assume not, but I do not know. If you have more insights here, I would be curious to learn.

If the suggestion applies more to setting the base URI on the client instead of building the full URI on each invocation, I fully agree and I will try to add tests for this and change this, to make sure that nothing breaks.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaApiHttpClient.cs
```diff
@@ -0,0 +1,57 @@
...
+                return handler;
+            });
+
+        internal static IServiceCollection AddApiClient(this IServiceCollection services, Func<HttpMessageHandler> createHttpMessageHandler)
+        {
+            _ = services.AddHttpClient(LoRaApiHttpClient.Name)
+                        .ConfigurePrimaryHttpMessageHandler(createHttpMessageHandler)
+                        // Retry aggressively first to not miss the receive window if possible.
+                        .AddTransientHttpErrorPolicy(policyBuilder =>
+                            policyBuilder.WaitAndRetryAsync(NumberOfRetries, i => (i <= NumberOfAggressiveRetries ? AggressiveRetryInterval : RetryInterval) * Math.Pow(2, i)));
```
bastbu, line 52: For the reviewer: Since all messages of the same device are processed in series, and the retries as defined here might take up to a minute, it's possible that the request queue is blocked for too long. Should we make this shorter?

kbeaugrand:
> @kbeaugrand Thanks for your suggestion! I did not get to adding retry policies yet for the `LoRaDeviceAPIService` (which is where your suggestion would apply if I understood correctly). I added these in [e7af10f](https://github.com/Azure/iotedge-lorawan-starterkit/commit/e7af10f5942aa0afa6571218e521e2bd111f85d3), such that we have this change in the PR now as well.
> 
> As for your suggestion - the main benefit I see from using typed clients is the increased type safety by not needing to rely on a `public const string` as an identifier for which `HttpClient` to use. If I look at the implications of registering the `LoRaDeviceAPIService` as a typed client, it will mean that the `LoRaDeviceAPIService` becomes transient instead of the singleton it now is.
> 
> If you see the simplification in not needing to invoke the `CreateClient` helper method, I think we can also solve this by storing the `HttpClient` on the `LoRaDeviceAPIService` instance instead of the `IHttpClientFactory`. I did not make this change since I do not grasp all the implications this has with respect to how the underlying connections are managed. Are there potential problems from keeping the same `HttpClient` instance for the lifetime of the `LoRaDeviceAPIService`? I would assume not, but I do not know. If you have more insights here, I would be curious to learn.
> 
> If your suggestion applies to setting the `BaseAddress` on the client instead of building the full URI on each invocation, I fully agree and I will try to add tests for this and change this, to make sure that nothing breaks.

In my mind, this is a bad option to keep an instance of HttpClient on your instances (maybe wrong). I remember reading some articles that told the HttpClientFactory was designed to provide a better way to manage the http client lifetime and mainly for the HttpMessageHandler that is taking the TCP connections... By using the HttpFactory (or with the DI) it could reduce the risk to loose a dispose ...

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaApiHttpClient.cs
spygi: Not a very educated opinion here but 8 retries in total sound a lot, maybe 6 (3 of them aggressive) and shorter intervals would be good enough?

spygi:
> Are there potential problems from keeping the same HttpClient instance for the lifetime of the LoRaDeviceAPIService?

I think the issue there is handling DNS updates:
> Another [issue](https://github.com/dotnet/runtime/issues/18348) that developers run into is when using a shared instance of HttpClient in long-running processes. In a situation where the HttpClient is instantiated as a singleton or a static object, it fails to handle the DNS changes as described in this issue of the dotnet/runtime GitHub repository.

[source](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#:~:text=Another%20issue%20that,runtime%20GitHub%20repository.)

Tests/Integration/LoRaApiHttpClientTests.cs
```diff
@@ -0,0 +1,40 @@
...
+            handler.SetupHandler(_ =>
+            {
+                ++count;
+                return new HttpResponseMessage(HttpStatusCode.InternalServerError);
+            });
+            var client = new ServiceCollection().AddApiClient(() => handler)
+                                                .BuildServiceProvider()
+                                                .GetRequiredService<IHttpClientFactory>()
+                                                .CreateClient(LoRaApiHttpClient.Name);
+            using var request = new HttpRequestMessage(HttpMethod.Get, new Uri("https://aka.ms"));
```
spygi, line 32: very-nit: should we maybe have a more clearly wrong URL here?

Tests/Common/DisposableValue.cs
```diff
@@ -9,13 +9,17 @@ namespace LoRaWan.Tests.Common
 
     public sealed class DisposableValue<T> : IDisposable
```
spygi, line 10: [nit] Can we add some docs explaining when this util class should be used (disposing related objects to the type T but not dispose explicitly the type T, if I understand correctly).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaApiHttpClient.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        private const int NumberOfAggressiveRetries = 4;
+        private const int NumberOfRetries = 8;
+        private static readonly TimeSpan AggressiveRetryInterval = TimeSpan.FromMilliseconds(50);
+        private static readonly TimeSpan RetryInterval = TimeSpan.FromMilliseconds(120);
+
+        public static IServiceCollection AddApiClient(this IServiceCollection services,
+                                                      NetworkServerConfiguration configuration,
+                                                      ApiVersion expectedFunctionVersion) =>
+            AddApiClient(services, () =>
```
spygi, line 29: Makes sense to have 2 separate methods here for easier testing 👍

spygi (approved):
LGTM and well tested!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaApiHttpClient.cs
p-schuler: I think that's ok. We will miss the downstream window potentially, but for any unconfirmed messages up, we get higher deliver guarantee.

Issue #955, opened 2021-12-02
[Story] Ensure that every message has its own Activity ID


bastbu:
## Description

Since we connect stations to the LNS by upgrading an HTTP request to a WS request, there is a single activity for all the messages that pass through the LNS. We should make sure that every message starts its own activity.

PR #1461, opened 2022-02-03
Add AppInsights tracing for data messages

bastbu:
# PR for issue #955

## What is being addressed

With this PR we ensure that separate data messages (i.e. upstream data frames) receive their own Request ID in Application Insights. This makes it easier to correlate logs with a certain data message (namely by the `operation_parentId` of the `traces` item in AppInsights). A typical span will then look like the following:

image

Under the hood, the Application Insights creates and starts the `System.Diagnostics.Activity` instances for us. Hence, we rely directly on the Application Insights SDK to implement the functionality.


p-schuler (approved):
LGTM - makes a great demo! we should keep this on the list of things to demo in the final CP.

Issue #1475, opened 2022-02-09
Function based LNS affinity - ADR


p-schuler:
## Acceptance Criteria

- ADR published to docs
- Implementation stories on backlog

Issue #1473, opened 2022-02-09
Single device connection per edge gateway


roel4ez:
In a multi-concentrator and multi-gateway scenario, a device can be in reach of multiple concentrators which are connected to a different edge gateway. When this happens, there will be a ping-pong for the open connection from the edge to iothub. There can only be one open connection, and iothub will actively close a connection as soon as a new one is initiated. However, the edgehub is reopening the connection as soon as it is closed (by design).
This means we need to implement a way for ensuring only one connection is open.

- [x] #1475
- [x] #1477
- [x] #1478
- [x] #1496
- [x] #1497
- [x] #1498
- [x] #1499
- [x] #1505
- [x] #1568

Issue #1468, opened 2022-02-08
Function does not handle the case when the winning LNS crashes after winning the race


spygi:
## Scenario - current behavior
- Device sends message that needs confirmation
- LNS1 wins the race and then crashes
- Device resubmits because it didn't get the confirmation
- LNS2 picks it up and contacts the function
- Function thinks this is a duplicate because a different gateway id is in Redis for the same frame counter

## Expected Behavior

The Function realises (how?) that LNS1 crashed and allows LNS2 to process the message.

PR #1483, opened 2022-02-11
Sticky LNS affinity

spygi:
# PR for issue #1475

### DIFF
--- a/docs/adr/009_lns_affinity.md
+++ b/docs/adr/009_lns_affinity.md
@@ -177,7 +177,7 @@ or via fetching it using DeviceGetter.GetDevice). Changes are marked in **bold**
    connection to Iot Hub is opened and only LNS1 has the connection to Iot Hub. **LNS2 updates its
    in memory state that it does not own the connection for this device**.
 1. When message B gets send (with a higher frame counter*), assuming that this time LNS2 gets it
-   first it **checks again its local state that indicates it's not owning the connection for the device and
+   first, it **checks again its local state that indicates it's not owning the connection for the device and
    therefore delays itself X ms before contacting the Function**.
    - Here we do *not* want to simply drop the message as LNS1 might not be available anymore (due to
      a crash, device not in range etc).
@@ -185,9 +185,12 @@ or via fetching it using DeviceGetter.GetDevice). Changes are marked in **bold**
     back to the previous case of message A. The active connection stays with LNS1.
 1. If this delay is not sufficient for LNS1 to win the race, LNS2 will contact the Function which
    now awards LNS2 as the "winning" LNS. LNS2 can now process the message upstream. It also **removes the "losing flag" from its in-memory store**.
-1. **The Function also proactively informs LNS1** that it's not anymore the winning LNS for this device.
-   - The reason why we do this is to ensure that LNS1 knows connection ownership was transferred to another network server and it can drop the connection. This is for the case, where the upstream message does not reach LNS1.
-  - For that we can use a C2D message to LNS. Alternative: direct method could be used here but
+1. **The Function also proactively informs LNS1** that it's not anymore the winning LNS for this
+   device.
+    - The reason why we do this is to ensure that LNS1 knows connection ownership was transferred to
+    another network server and it can drop the connection. This is for the case, where the upstream
+    message does not reach LNS1.
+    - For that we can use a C2D message to LNS. Alternative: direct method could be used here but
     delivery will not be guaranteed then.
 1. If LNS1 in the meantime gets message B and contacts the Function, it will let it know
    that it lost the race for this frame counter and must therefore drop the message, **mark
--- a/docs/adr/009_lns_affinity.md
+++ b/docs/adr/009_lns_affinity.md
@@ -72,19 +72,28 @@ Version, LNS discovery and CUPS update endpoints are not affected.
 The main idea is to give the current connection holder as indicated from the Function, the edge to
 continue processing messages for this device. That gateway, will have 0 impact on performance.
 
-### Handling of background tasks
+### Handling of cache refresh
 
-When we create a (singleton) instance of the NetworkServer.LoRaDeviceCache we start a background periodical
-check to ensure the device twins for the DevEuis we encountered are kept fresh.
+When we create a singleton instance of the NetworkServer.LoRaDeviceCache, we start a background periodical
+check to ensure the device twin for all the devices that connected to the LNS are kept fresh.
 
-- If we were previously the losing LNS for the DevEui that needs refresh, we could adjust the
-  LastSeen property but not actually refresh the entry
-  - this way we won't try to read the twin that could result in a connection switch
-  - when the next data message comes and we check the cache
-    - if entry is in cache, we have a stale twin in the cache - how should it be solved ❔
-    - if entry not in cache, see handling in [the section on LoRaDevice not in cache](#data-flow-loradevice-not-in-loradevicecache)
-  
-Alternatives
+This periodical background task is checking if the LNS received a device for such DevEUI in the last 10 minutes,
+if not the device is removed from cache, otherwise a "get twin" operation is executed for such device.
+
+In case multiple LNS are handling the same device "XYZ", this is triggering a connection ping-pong.
+
+Ideally we would need to refresh the cache only if, after reaching out to the function,
+the LNS is marked as the "winning" one for device "XYZ".
+
+It is important though to not remove this device from the cache, in case the LNS is still the "losing" gateway.
+If we remove it, this would trigger a "get twin" operation as soon as the next data message is coming in for "XYZ".
+
+To prevent this, it could be possible to adjust the LastSeen property but not actually refresh the entry in the cache.
+
+When the next data message for "XYZ" comes in and an entry is in cache for such device, in the eventuality that
+the Azure Function is marking our LNS as the new "winning" gateway, we have a stale twin in the cache, to be updated before processing.
+
+Alternatives considered, but not viable
 
 - Remove entry completely from the cache which is the same as [the section on LoRaDevice not in
   cache](#data-flow-loradevice-not-in-loradevicecache)
@@ -105,7 +114,7 @@ handle "backoff" in the same way as in the Data flow. The advantages of doing th
 
 - subsequent first Data (or future Join) requests do not switch the connection
 - same handling as in the Data flow (therefore easier to understand)
-  
+
 Alternative is to do nothing and accept that a one-off connection stealing on the first Data message can happen.
 
 ### Data flow: LoRaDevice not in LoRaDeviceCache
--- a/docs/adr/009_lns_affinity.md
+++ b/docs/adr/009_lns_affinity.md
@@ -73,7 +73,7 @@ Version, LNS discovery and CUPS update endpoints are not affected.
 
 ## Possible solution
 
-The main is to store locally on the LNS a dictionary mapping DevEui and whether we were the losing
+The main idea is to store locally on the LNS a dictionary mapping DevEui and whether we were the losing
 gateway for that DevEui, as decided from the Function. Based on that we would delay the processing
 of future messages for all losing gateways. This should give enough time to the preferred LNS to
 process the message and keep the active connection to Iot Hub.
@@ -98,7 +98,7 @@ Alternatives
   - Not really a background task anymore
 - Do nothing and accept there will be a potential connection switching periodically (currently every
   2 days)
-- Remove the background refresh
+- Remove the background refresh feature all-together.
 
 ### Handling of Join requests
 
@@ -142,15 +142,13 @@ We should instead:
 - Otherwise we process message normally.
 - And at the end force update the twin.
 
-Alternative considered but discarded:
-
-Use the Function to do both operations
+Alternative considered but discarded: use the Function to do both operations
 
 - update the device twin frame counter to 0 (it needs a DeviceClient that would
   cause a temporary connection switch)
 - Clear or update the cache entry with frame counter down and up to 0.
 - Returns the result to the LNS: whether it was the winning or losing one
-- LNS reacts as described before
+- LNS reacts as described in the [main data flow section](#main-data-message-flow)
 
 ### Main data message flow
 
--- a/docs/adr/009_lns_affinity.md
+++ b/docs/adr/009_lns_affinity.md
@@ -55,38 +55,80 @@ stations.
   - C2D messages
 - Roaming edge devices (that potentially lose connectivity to an LNS) are in scope.
 
-### Problematic connections opened on behalf of edge devices
+### Problematic IoT Hub operations on behalf of edge devices
 
-- Data message flows
-  - Join: 
-  - Data: MessageDispatcher.DispatchLoraDataMessage gets the twin (if LoRaDevice not in cache)
-    - Direct method for Class C should return an error if it could not send
-      message downlink so that the user can retry.
+- Background tasks
+  - Periodically we refresh the LoRaDeviceCache, which results in device twin reads that could
+    switch the connection -> see [handling of background tasks section](#handling-of-background-tasks)
+- Message flows
+  - Join -> see [handling of join requests](#Handling-of-Join-requests)
+  - Data:
+    - if the device is not in LoRaDeviceCache, we fetch the device twin -> see [LoRaDevice not in cache section](#data-flow-loradevice-not-in-loradevicecache)
+    - if a frame counter reset happened, we update the twin immediately -> see [handling of resets section](#handling-of-device-resets)
+    - in the main data flow we send upstream, downstream and write the new twin -> see [main data flow section](#main-data-message-flow)
+    - Class C direct downstream messages sent from the portal could result in a connection switch if
+    the handling LNS changed in the meantime but we consider this as out of scope.
 
 Version, LNS discovery and CUPS update endpoints are not affected.
-
-### Background tasks
-
-- LoRaDeviceCache refresh in the background: every 10 minutes we are checking for devices that need
-  refreshing (those that were not refreshed in the last 2 days). Refreshing a device gets the twin
-  of the device ❔
-  - if we are the losing LNS we adjust LastSean property but not refresh the entry
-  - in order to not have stale entries we could refresh the twin once we become the winning LNS from
-    the data (also required for join?) flow  
-
-Alternative:
-SearchByDevAddr: the function returns info about the winning gateway
-Requires doing the Mic computation on the Function
 
 ## Possible solutions
 
 ### Delayed processing of messages from losing LNSs
 
-The main idea here is to delay the processing of future messages for all gateways *besides* the
-winning one. This should give enough time to the chosen LNS to process the message and keep the
-active connection to Iot Hub. The winning gateway is decided by the Function.
+The main idea here is to store locally on the LNS a dictionary mapping DevEui and whether we were
+the losing gateway for that DevEui, as decided from the Function. Based on that we would delay the
+processing of future messages for all losing gateways. This should give enough time to the preferred
+LNS to process the message and keep the active connection to Iot Hub.
 
-#### Example scenario in main data message flow
+#### Handling of background tasks ❔
+
+When we create a (singleton) instance of the NetworkServer.LoRaDeviceCache we start a background periodical
+check to ensure the device twins for the DevEuis we encountered are kept fresh.
+
+- If we were previously the losing LNS for the DevEui that needs refresh, we could adjust the
+  LastSeen property but not actually refresh the entry
+  - this way we won't try to read the twin that could result in a connection switch
+  - when the next data message comes and we check the cache
+    - if entry is in cache, we have a stale twin in the cache - how should it be solved ❔
+    - if entry not in cache, see handling in [the section on LoRaDevice not in cache](#data-flow-loradevice-not-in-loradevicecache)
+  
+Alternatives
+
+- Remove entry completely from the cache which is the same as [the section on LoRaDevice not in
+  cache](#data-flow-loradevice-not-in-loradevicecache)
+- Delay and then call the Function to check whether we are still the losing LNS
+  - Not really a background task anymore
+- Do nothing and accept there will be a potential connection switching periodically (currently every
+  2 days)
+- Remove the background refresh
+
+#### Handling of Join requests
+
+Join requests in isolation currently do not have the connection stealing issue, as they already rely on the
+Function for the DevNonce check. If the current LNS is not the preferred gateway, it drops the
+message immediately.
+
+❔ We could also consider storing the information that we were the losing LNS locally on the LNS and
+handle "backoff" in the same way as in the Data flow. The advantages of doing that are:
+
+- subsequent first Data (or future Join) requests do not switch the connection
+- same handling as in the Data flow (therefore easier to understand)
+  
+Alternative is to do nothing and accept that a one-off connection stealing on the first Data message can happen.
+
+#### Data flow: LoRaDevice not in LoRaDeviceCache
+
+Currently, if LoRaDevice is not in the cache we search on the Function for devices with the specific
+DevAddr then we load their twins.
+
+We could ❔
+
+- Amend DeviceGetter.GetDevice so that it returns info about whether we are the winning gateway
+  - if we are (we already have the connection) continue to load the twin
+  - if we are not the winning LNS, drop the message and mark our selves as the losing gateway
+  - This would require moving the Mic computation on the Function (?)
+
+#### Main data message flow
 
 Assuming the topology:
 
@@ -105,7 +147,7 @@ Here is a rundown of what should happen marked in **bold**:
 - Device sends first data message A.
 - We assume that LNS1 gets the message first. **LNS1 checks against
   an in-memory dictionary DevEui -> flag** and since it has not seen this DevEui before (flag is
-  false) contacts the FunctionBundler.
+  false) contacts the GetDevice.
 - The Function hasn't seen this DevEui either and therefore does not have an assigned LNS for it
   yet. LNS1 wins the race and gets immediately a response and processes the message upstream.
 - LNS2 eventually receives message A, **checks its local dictionary** and also contacts the FunctionBundler since it does not have prior info about this DevEui.
@@ -116,16 +158,18 @@ Here is a rundown of what should happen marked in **bold**:
 - When message B gets send (with a higher frame counter), assuming that this time LNS2 gets it
   first it **checks again its local dictionary it's not the preferred LNS for this device and
   therefore delays itself X ms before contacting the FunctionBundler**.
+  - Here we can *not* simply drop as LNS1 might not be available anymore (due to a crash, device not
+    in range etc).
 - This delay gives LNS1 a time advantage to reach the FunctionBundler first and win the race again, failing
   back to the previous case of message A. The active connection stays with LNS1.
   - If this delay is not sufficient for LNS1 to win the race, LNS2 will contact the FunctionBundler which
     now awards LNS2 as the "winning" LNS. LNS2 will process message upstream (therefore the active
     connection will switch to it) and **removes the "losing flag" from the in-memory store**.
-  - **The Function also proactively informs LNS1** that is not anymore the winning LNS for this
+  - **The Function also proactively informs LNS1** that it's not anymore the winning LNS for this
     device. The reason why we do this proactively is that otherwise its Edge Hub will try to
     reconnect to IoT Hub even if there is no more messages picked up from LNS1 (out of range roaming
     client). Additionally, **LNS1 would not need to refresh its LoRaDeviceCache anymore for this
-    device**.
+    device** (see [previous section about caching](#data-flow-loradevice-not-in-loradevicecache)).
     - Direct method and should be retried (e.g. in memory retries or durable Function): direct
       method could fail due to the module being or Edge Hub being down, other connectivity issue
       etc. Is in memory retries sufficient here? ❔
@@ -133,27 +177,52 @@ Here is a rundown of what should happen marked in **bold**:
     that it lost the race for this frame counter and must therefore drop the message, **mark
     itself as the losing LNS and close the connection**.
 
-#### Handling of Join requests
+NB: the FunctionBundler is not called in certain topologies e.g. when multiple LBSs are connected to
+the same LNS but these topologies are not relevant for the issue here).
 
-Join requests in isolation currently do not have the connection stealing issue, as they rely on the Function for
-the DevNonce check. If the current LNS is not the preferred gateway, it drops the message
-immediately.
+#### Handling of device resets
 
-We could also consider storing the information that we were the losing LNS locally on the LNS. The
-advantages of that are:
+What happens in the above scenario, if we have a device reset between message A and B? Currently we
+are saving the twin immediately and then clear the Function cache.
 
-- subsequent Data (or Join) requests do not potentially switch the connection
-- same handling as in the Data flow (therefore easier to follow)
-  
-Not sure the following section is relevant as both data and join flow need DevEui.
-The problem is that the Data message flow relies on the DevAddr while here we only have the DevEui.
-Ways to solve this:
+This is both dangerous as the request is not yet validated but is changing state of the twin as well
+as not required as at the end of processing the message we do sync changes on the twin.
 
-- Split the creation of the LoRaDevice from fetching the twin, create a dummy LoRaDevice and
-  mark our LNS as the losing one.
-- Store outside of LoRaDevice a data-structure with DevEui: flag (winning or losing) and inject
-  also on the data message handler
-- Do nothing and accept that a one-off connection stealing on the first data message can happen.
+We should instead:
+
+- Clear the cache in the Function as we do currently.
+- Function returns if we are the losing LNS. If we are the losing one, we drop the message here,
+  mark ourselves as the losing gateway etc.
+- Otherwise we process message normally.
+- And at the end force update the twin.
+
+Alternative considered but discarded:
+
+Use the Function to do both operations
+
+- update the device twin frame counter to 0 (it needs a DeviceClient that would
+  cause a temporary connection switch)
+- Clear or update the cache entry with frame counter down and up to 0.
+- Returns the result to the LNS: whether it was the winning or losing one
+- LNS reacts as described before
+
+##### Handling Class C devices downstream direct messages
+
+When we send a downstream message using Direct method via the Function,
+the Function is responsible for choosing the prererred gateway based on its existing knowledge of
+the preferred gateway. Due to that we can be sure that we have the active connection and can
+therefore send downstream messages.
+
+#### Implementation
+
+The LNS in-memory dictionary is mapping DevEuis to a flag indicating whether we are the winning or losing LNS.
+
+- Where should we store locally this structure ❔
+
+  - LNS LoRaDeviceClient.ConnectionManager since all of the operations pass through it.
+  - LoRaDevice: in the case of Join requests we don't yet have a LoRaDevice and would need to create
+    one without fetching the twin
+  - LoRaDeviceClient
 
 #### Delay on the LNS itself or on the Function ❔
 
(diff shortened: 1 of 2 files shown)


docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+per device but we find it acceptable for the Mark and None strategies to not be as scalable as the
+Drop strategy and will only document this limitation for potential users to be aware of.
+
+- LNS performs operations on behalf of a device/sensor and a concentrator/station. However since a
+concentrator can be connected to at most one LNS, there is no ping-pong happening with operations on
+stations.
+
+## Limitations
+
+- We should avoid introducing additional calls to the Function as this would also hurt scalability.
```
p-schuler, line 46: if we would need to do so, we could. I would not see this as a limitation that needs to be called out here.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+  
+## In-scope
+
+- The problem can be manifested whenever we do operations against Iot Hub on behalf of edge devices.
+  These can be:
+  - Twin reads
+  - Twin writes (updates/deletes)
+  - D2C messages
+  - C2D messages
+- Roaming edge devices (that potentially lose connectivity to an LNS) are in scope.
```
p-schuler, line 56: I think you want to say "leaf" devices? edge is not roaming. If that's what you meant, then "connectivity" is also not accurate, as the leaf devices are not "connected" they are in range.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Problematic connections opened on behalf of edge devices
+
+- Data message flows
+  - Join:
```
p-schuler, line 61: seems missing some information?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+  - Twin writes (updates/deletes)
+  - D2C messages
+  - C2D messages
+- Roaming edge devices (that potentially lose connectivity to an LNS) are in scope.
+
+### Problematic connections opened on behalf of edge devices
+
+- Data message flows
+  - Join: 
+  - Data: MessageDispatcher.DispatchLoraDataMessage gets the twin (if LoRaDevice not in cache)
```
p-schuler, line 62: both join and data get the twins. That's the case anytime we create and initialize a device that is not in the cache.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+
+Version, LNS discovery and CUPS update endpoints are not affected.
+
+### Background tasks
+
+- LoRaDeviceCache refresh in the background: every 10 minutes we are checking for devices that need
+  refreshing (those that were not refreshed in the last 2 days). Refreshing a device gets the twin
+  of the device ❔
+  - if we are the losing LNS we adjust LastSean property but not refresh the entry
+  - in order to not have stale entries we could refresh the twin once we become the winning LNS from
```
p-schuler, line 74: I think that would be the most optimal solution. Might be too complex. If so, we can consider moving the refresh to the message handling flow rather than doing it in the background. This is somewhat overlapping with the work Atif is doing on the connection management on the LNS side.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+    the data (also required for join?) flow  
+
+Alternative:
```
p-schuler, line 77: I don't understand this section. Alternative to what? the background task?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+SearchByDevAddr: the function returns info about the winning gateway
+Requires doing the Mic computation on the Function
+
+## Possible solutions
+
+### Delayed processing of messages from losing LNSs
+
+The main idea here is to delay the processing of future messages for all gateways *besides* the
+winning one. This should give enough time to the chosen LNS to process the message and keep the
```
p-schuler, line 86: I would formulate this a bit differently. We are giving the current connection owner an edge, to win the processing race. That gateway, will have 0 impact on performance.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+    Device-->LBS2-->LNS2-->Function;
+    LNS1-->IoTHub;
+    LNS2-->IoTHub;
+```
+
+where Device sends data message A and then B.
+
+Here is a rundown of what should happen marked in **bold**:
+
+- Device sends first data message A.
```
p-schuler, line 105: nit: I would suggest to use numbers here.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+- We assume that LNS1 gets the message first. **LNS1 checks against
+  an in-memory dictionary DevEui -> flag** and since it has not seen this DevEui before (flag is
+  false) contacts the FunctionBundler.
```
p-schuler, line 108: nit: FunctionBundler seems too much info here.

p-schuler: and I think, if it is not in the cache, it will first talk to the GetDevice endpoint?

p-schuler: I think I got it now. You would keep the state in that dictionary? But you would always call the function anyway no? There should be no additional action before the call?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+    now awards LNS2 as the "winning" LNS. LNS2 will process message upstream (therefore the active
+    connection will switch to it) and **removes the "losing flag" from the in-memory store**.
+  - **The Function also proactively informs LNS1** that is not anymore the winning LNS for this
+    device. The reason why we do this proactively is that otherwise its Edge Hub will try to
+    reconnect to IoT Hub even if there is no more messages picked up from LNS1 (out of range roaming
+    client). Additionally, **LNS1 would not need to refresh its LoRaDeviceCache anymore for this
+    device**.
+    - Direct method and should be retried (e.g. in memory retries or durable Function): direct
+      method could fail due to the module being or Edge Hub being down, other connectivity issue
+      etc. Is in memory retries sufficient here? ❔
```
p-schuler, line 131: I would argue yes. Worst case we will keep the connection for this device. If the device is not down, it's likely we can send the notification. If it is down, it will re-initialize when coming back up - as the cache is gone.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+  - If LNS1 in the meantime gets message B and contacts the FunctionBundler, it will let it know
+    that it lost the race for this frame counter and must therefore drop the message, **mark
+    itself as the losing LNS and close the connection**.
+
+#### Handling of Join requests
+
+Join requests in isolation currently do not have the connection stealing issue, as they rely on the Function for
+the DevNonce check. If the current LNS is not the preferred gateway, it drops the message
+immediately.
```
p-schuler, line 140: what about re-joins?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+
+We could also consider storing the information that we were the losing LNS locally on the LNS. The
```
p-schuler: we need to talk about this section. We will need the information for sure. In one case, we would also need it to delay but most of the cases we need the information to decide, if we should create the connection if asked to.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+advantages of that are:
+
+- subsequent Data (or Join) requests do not potentially switch the connection
+- same handling as in the Data flow (therefore easier to follow)
+  
+Not sure the following section is relevant as both data and join flow need DevEui.
+The problem is that the Data message flow relies on the DevAddr while here we only have the DevEui.
```
p-schuler: not sure I get that. data message flow relies on the devaddr while here we only have the deveui? Let's talk about this section. I would like to understand what you had in mind here.

docs/adr/009_lns_affinity.md
p-schuler: I would focus on the function side processing, rather than the LNS side delay.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+  
+#### Device reset ❔
+
+What happens if we have a device reset between message A and B? We save the twin
+immediately (irrelevant but how does the Function get it for the deduplication decision?)
+
+Unclear what is the relationship between the twin and the cache on the function: in the cache we
+keep the framecounter down from the twin but not the fcnt up ❔
+
+- If both LNSs receive message B, LNS1 SHOULD reset the frame counter. LNS2 then
```
p-schuler: I think this was a problem before. Both would perform the reset. Not sure this is in scope here?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+  - Clear cache entry or update the entry with both frame counter down and up to 0.
+  - Returns the result to the LNS: whether it was the winning or losing one
+  - LNS reacts as described before
+
+Alternative would be to switch the order of the operations: first clear the cache and then (if we
+are the winning LNS) update the twin.
+
+#### Feature flag
+
+Unclear what is the exact use-case here ❔ We could still set the delay to 0.
```
p-schuler: Anything will work that allows to conditionally disable it.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+- This flag is only checked if we are on the Drop deduplication strategy (as Mark and None do not
+  support stickiness anyway)
+
+#### Consequences / implementation
+
+- In memory storage is DevEui -> boolean.
+- Where to store locally on LNS❔ Can be on the
+
+  - LNS LoRaDeviceClient.ConnectionManager since all of the operations pass through it.
+  - LoRaDevice
```
p-schuler: I think it will need to go to the LoRaDevice. But that's too far into the implementation detail for the ADR I think.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+On deployments where multiple gateway servers are used, they need to have a parent 
+
+Child LNS A ____ | Parent Gateway using AMQP maintains a single connection to IoTHub | Child LNS B
+                ____|
+
+Not possible due to single parent limitation (not supporting roaming LNSs)
+
+## Questions
+
+❔ Should we investigate if creating a DeviceClient already steals the connection or we know this
```
p-schuler, line 272: I don't think it matters. If we want to be sure, we dispose the client actively. We can't be sure about what happens in the SDK even if we verify that behavior now.

docs/adr/009_lns_affinity.md
spygi: Section was still WIP.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+
+- Split the creation of the LoRaDevice from fetching the twin, create a dummy LoRaDevice and
+  mark our LNS as the losing one.
+- Store outside of LoRaDevice a data-structure with DevEui: flag (winning or losing) and inject
+  also on the data message handler
+- Do nothing and accept that a one-off connection stealing on the first data message can happen.
+
+#### Delay on the LNS itself or on the Function ❔
+
+What are the scenarios that it's better that the function implements the delay?
```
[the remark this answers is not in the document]
spygi: Not sure we talk about the same thing: the question here is whether it's better to do a Task.Delay on the Function or on the losing LNS itself. 

I do agree (it's written down on the main data flow) that for roaming devices, the Function should proactively inform LNSs that they are now the losing gateways.

docs/adr/009_lns_affinity.md
spygi: True, the SDK behavior might change. And looking [at the current code where the SDK creates the DeviceClient](https://github.com/Azure/azure-iot-sdk-csharp/blob/2cf35f3ee541f3311cfcda3ce3b11290a9f88ae7/iothub/device/src/ClientFactory.cs#L485), they defer creating the right type of instance to later. Will remove this.

docs/adr/009_lns_affinity.md
spygi: Correct, adjusted the wording.

docs/adr/009_lns_affinity.md
spygi: Removed.

docs/adr/009_lns_affinity.md
spygi: Sorry, section was still WIP.

docs/adr/009_lns_affinity.md
spygi: Ok, removed this sub-section then.

docs/adr/009_lns_affinity.md
spygi: Removed this section - was a misunderstanding (and also too deep in implementation detail for the ADR)

docs/adr/009_lns_affinity.md
spygi: Reworded.

docs/adr/009_lns_affinity.md
spygi: The section is re-written but to your comment I think for joins, we first call `loRaDeviceAPIService.SearchAndLockForJoinAsync` before getting the twin, which means if the DevNonce is already used we don't get the twin.

p-schuler: absolutely, however, there are edge cases, where the load of the twin for a stale dev address matching your join deveui could conflict with the load you are doing for the join (resoling to the same devaddr). It's an edge case, but simply by protecting using the devnonce, you are not protected against all races when loading the twins. Something Atif is looking into.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,253 @@
...
+  - Periodically we refresh the LoRaDeviceCache, which results in device twin reads that could
+    switch the connection -> see [handling of background tasks section](#handling-of-background-tasks)
+- Message flows
+  - Join -> see [handling of Join requests section](#Handling-of-Join-requests)
+  - Data:
+    - if the device is not in LoRaDeviceCache, we fetch the device twin -> see [LoRaDevice not in cache section](#data-flow-loradevice-not-in-loradevicecache)
+    - if a frame counter reset happened, we update the twin immediately -> see [handling of resets section](#handling-of-device-resets)
+    - in the main data flow we send upstream, downstream and write the new twin -> see [main data flow section](#main-data-message-flow)
+    - Class C direct downstream messages sent from the portal could result in a connection switch if
+    the handling LNS changed in the meantime but we consider this as out of scope.
```
Mandur, line 66:
```suggestion
    it targets a LNS other than the last connected one but we consider this as out of scope.
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Handling of background tasks
+
+When we create a (singleton) instance of the NetworkServer.LoRaDeviceCache we start a background periodical
+check to ensure the device twins for the DevEuis we encountered are kept fresh.
+
+- If we were previously the losing LNS for the DevEui that needs refresh, we could adjust the
+  LastSeen property but not actually refresh the entry
+  - this way we won't try to read the twin that could result in a connection switch
+  - when the next data message comes and we check the cache
```
Mandur, line 83: I would be more specific here to indicate you are indicating the case of a primary gateway switch

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,253 @@
...
+    - if entry is in cache, we have a stale twin in the cache - how should it be solved ❔
```
Mandur, line 84: If I understood the discussion from earlier in this case we would detect that the current gateway switch from passive to active mode and we would load the twin fresh to refresh the cache... Is this correct?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,253 @@
...
+Function for the DevNonce check. If the current LNS is not the preferred gateway, it drops the
+message immediately.
+
+❔ We could also consider storing the information that we were the losing LNS locally on the LNS and
+handle "backoff" in the same way as in the Data flow. The advantages of doing that are:
+
+- subsequent first Data (or future Join) requests do not switch the connection
+- same handling as in the Data flow (therefore easier to understand)
+  
+Alternative is to do nothing and accept that a one-off connection stealing on the first Data message can happen.
```
Mandur, line 109: IMHO, I think this is fine that there are eventual connection steal on join

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,253 @@
...
+
+Currently, if LoRaDevice is not in the cache we search on the Function for devices with the specific
+DevAddr then we load their twins.
+
+We could ❔
+
+- Amend DeviceGetter.GetDevice so that it returns info about whether we are the winning gateway
+  - if we are (we already have the connection) continue to load the twin
+  - if we are not the winning LNS, drop the message and mark our selves as the losing gateway
+  - This would require moving the Mic computation on the Function (?)
```
Mandur, line 121: I Do not understand why this would change MIC computation, could we not simply do that when after the getdevice

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Handling of device resets
```
Mandur, line 123:
```suggestion
### Handling of ABP relax frame counter reset
```

docs/adr/009_lns_affinity.md
spygi: Agree as well. Removed from the questions then but we can still discuss with the team.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,262 @@
...
+   first it **checks again its local dictionary it's not the preferred LNS for this device and
+   therefore delays itself X ms before contacting the FunctionBundler**.
+   - Here we can *not* simply drop as LNS1 might not be available anymore (due to a crash, device not
+     in range etc).
+1. This delay gives LNS1 a time advantage to reach the FunctionBundler first and win the race again, failing
+    back to the previous case of message A. The active connection stays with LNS1.
+1. If this delay is not sufficient for LNS1 to win the race, LNS2 will contact the FunctionBundler which
+   now awards LNS2 as the "winning" LNS. LNS2 will process message upstream (therefore the active
+   connection will switch to it) and **removes the "losing flag" from the in-memory store**.
+1. **The Function also proactively informs LNS1** that it's not anymore the winning LNS for this
```
Mandur, line 193: Is the mechanism for that a C2D message?

Mandur: Oh I see below you mention a direct method.. Why not a C2D message?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,262 @@
...
+   memory that it was the losing gateway for this DevEui**.
+1. When message B gets send (with a higher frame counter), assuming that this time LNS2 gets it
+   first it **checks again its local dictionary it's not the preferred LNS for this device and
+   therefore delays itself X ms before contacting the FunctionBundler**.
+   - Here we can *not* simply drop as LNS1 might not be available anymore (due to a crash, device not
+     in range etc).
+1. This delay gives LNS1 a time advantage to reach the FunctionBundler first and win the race again, failing
+    back to the previous case of message A. The active connection stays with LNS1.
+1. If this delay is not sufficient for LNS1 to win the race, LNS2 will contact the FunctionBundler which
+   now awards LNS2 as the "winning" LNS. LNS2 will process message upstream (therefore the active
```
Mandur, line 191: I think here a step is missing:

LNS2 contacts the IoT Hub to refresh its device twins. 

As the cache entry might be stale?

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,262 @@
...
+
+- Where should we store locally this structure ❔
+
+  - LNS LoRaDeviceClient.ConnectionManager since all of the operations pass through it.
+  - LoRaDevice: in the case of Join requests we don't yet have a LoRaDevice and would need to create
+    one without fetching the twin
+  - LoRaDeviceClient
+
+#### Delay on the LNS itself or on the Function ❔
+
```
Mandur, line 226: I think an advantage of doing it locally, might be the possibility to compute the artifical delay programmatically. ie. If you lost a lot of time in the first phase of the message you could reduce the delay to avoid that the new gateway miss the receive window.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,262 @@
...
+   an in-memory dictionary DevEui -> flag** and since it has not seen this DevEui before (flag is
+   false) contacts the GetDevice.
+1. The Function hasn't seen this DevEui either and therefore does not have an assigned LNS for it
+  yet. LNS1 wins the race and gets immediately a response and processes the message upstream.
+1. LNS2 eventually receives message A, **checks its local dictionary** and also contacts the FunctionBundler since it does not have prior info about this DevEui.
+1. The Function responds to LNS2 that it lost the race to process this message.
+1. Since deduplication strategy is Drop, LNS2 drops the message immediately, therefore no
+   connection to Iot Hub is opened and only LNS1 has the connection to Iot Hub. **LNS2 also notes in
+   memory that it was the losing gateway for this DevEui**.
+1. When message B gets send (with a higher frame counter), assuming that this time LNS2 gets it
```
Mandur, line 183: **I really love this scenario breakdown**

I think here we only consider the case with higher frame counter. However it could be cases where the frame counter could be lower (ABP Relax mode, or simply FCNT 32 bits rollover). I would suggest to also state in this flow when you actually check the FrameCounter. We should probably have a smart way of not dropping for wrong fcnt reason when you are the passive gateway.

Mandur:
Amazing work!

docs/adr/009_lns_affinity.md
spygi: I am not sure as well - I discussed this with @danigian and left the note there but don't get it either.

docs/adr/009_lns_affinity.md
spygi: Yes, this flow assumes higher frame counter for later messages, the case of ABP relax mode is discussed in the previous section.

Do you think we should merge the 2 in one? My thinking of separating the 2 was to keep the main (or most common?) flow clean but we can change that. Or the feedback is something else?

docs/adr/009_lns_affinity.md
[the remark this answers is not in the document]
spygi: Thinking more about it I think the only way to handle correctly resubmits and roaming clients is to have the delay on the function too:

- A confirmed message comes in and picked up by LNS1 and 2, LNS1 wins the race this time.
- C2D window missed, device resubmits and this time it's picked only by LNS2 (out of range of LNS1). Without an expiration of the winning gateway, Function thinks LNS1 is still handling this message and never switches the connection to LNS2.

docs/adr/009_lns_affinity.md
spygi: Edit: I think we are missing the resubmit case (frame counter equal to before) - I commented on another thread https://github.com/Azure/iotedge-lorawan-starterkit/pull/1483/files#r807395835

docs/adr/009_lns_affinity.md
p-schuler: We need to think through the cases. That's an option. We can also make it depending on the last time we did actually refresh. Not sure we have many issues besides the fcnt drift. That would happen only, if we didn't refresh for a very long time.

docs/adr/009_lns_affinity.md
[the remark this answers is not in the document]
p-schuler: let's talk about it. I think this is an existing issue. Not sure we can reliably handle this anyway. The delay does not really have to be on the server though to be able to fix that? We could indicate the delay as part of the request and know on the server, that the request was delayed and then assume the original winning gateway won't process it. But we would need to know it's a resubmit in the function. I think we talked about this. Something out of scope for this, but good to consider the case.

docs/adr/009_lns_affinity.md
spygi: Resolving this as the section was re-written and the comment isn't applicable.

docs/adr/009_lns_affinity.md
spygi: Removed this section (but this comment might as well apply to the full ADR, though given the different code paths I think we ought to be specific and more technical in the ADR)

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+
+
+### Parent-child gateways
+
+On deployments where multiple gateway servers are used, they need to have a parent 
+
+Child LNS A ____ | Parent Gateway using AMQP maintains a single connection to IoTHub | Child LNS B
+                ____|
+
+Not possible due to single parent limitation (not supporting roaming LNSs)
```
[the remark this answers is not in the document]
spygi: Fixed.

docs/adr/009_lns_affinity.md
spygi: Was added explicitly that we are now the winning LNS.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,248 @@
...
+    - if the device is not in LoRaDeviceCache, we fetch the device twin -> see [LoRaDevice not in cache section](#data-flow-loradevice-not-in-loradevicecache)
+    - if a frame counter reset happened, we update the twin immediately -> see [handling of resets section](#handling-of-device-resets)
+    - in the main data flow we send upstream, downstream and write the new twin -> see [main data flow section](#main-data-message-flow)
+
+Version, LNS discovery and CUPS update endpoints are not affected by this issue.
+
+## Possible solution
+
+The main idea is to give the current connection holder (as indicated from the Function), the edge to
+continue processing messages for this device. That gateway, will have 0 impact on performance.
```
MaggieSalak, line 74: Would it make sense to rephrase:
```suggestion
The performance of that gateway should not be impacted.
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,248 @@
...
+
+A local (per LNS) in-memory dictionary will be used for that purpose to map DevEuis to a flag
+indicating whether the LNS is the preferred or not gateway.
```
MaggieSalak, line 77: [nit]
```suggestion
indicating whether the LNS is the preferred gateway or not.
```

docs/adr/009_lns_affinity.md
spygi: > I would focus on the function side processing, rather than the LNS side delay.

As discussed we think it's sufficient to use local delay on the LNS. Resolving this.

docs/adr/009_lns_affinity.md
spygi: The join section was re-written, please resolve if OK for you @p-schuler

docs/adr/009_lns_affinity.md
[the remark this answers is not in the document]
spygi: Bug is https://github.com/Azure/iotedge-lorawan-starterkit/issues/1468

docs/adr/009_lns_affinity.md
spygi: Section was re-written and 2 issues were created to track changes: #1499  and #1505

docs/adr/009_lns_affinity.md
spygi: We decided to close the connection after a join proactively (the option to allow a connection steal is still listed as a candidate)

docs/adr/009_lns_affinity.md
spygi: After the discussion, we need the Mic computation to see if we can decrypt the message (find out the real DevEui from the DevAddr). This is something we can not do on the Function because we need to send the payload to the cloud.

docs/adr/009_lns_affinity.md
spygi: Changed to a C2D message, thanks for the suggestion!

docs/adr/009_lns_affinity.md
spygi: Bug https://github.com/Azure/iotedge-lorawan-starterkit/issues/1468 is tracking the issue with resubmit

spygi: Please let me know if there are other cases here missing.

docs/adr/009_lns_affinity.md
spygi: Listed it as a potential advantage.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,244 @@
...
+  also on the data message handler
+- Do nothing and accept that a one-off connection stealing on the first data message can happen.
+
+#### Delay on the LNS itself or on the Function ❔
+
+What are the scenarios that it's better that the function implements the delay?
+
+Disadvantages of using a Task.Delay on the Function:
+
+- Observability: potentially we are messing up the measurements of the Function duration.
```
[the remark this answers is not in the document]
spygi: Added that feedback.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+
+Consider the topology:
+
+```mermaid
+flowchart LR;
+  Device-->LBS1-->LNS1--1-->IoTHub;
+  Device-->LBS2-->LNS2--2-->IoTHub;
+```
+
+where LNS 1 and 2 make use of their IoT Edge Hub modules to connect to IoT Hub.
```
p-schuler, line 20: nit: they do use their module client, indeed, but I don't see why that is relevant when we want to talk about device connections? I would remove this, as it only adds confusion and distracts from the topic we want to describe.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+
+- LNS performs operations on behalf of a device/sensor and a concentrator/station. However since a
+concentrator can be connected to at most one LNS, there is no ping-pong happening with operations on
+stations.
+
+- Class C downstream messages using Direct Method:
+  - When we send a downstream message via the Function, the Function is responsible for choosing the
+    prererred gateway based on its existing knowledge of the preferred gateway. Due to that we can
+    be sure that we have the active connection and can therefore send downstream messages.
+  - Messages sent from the portal could result in a connection switch if they target a LNS other
```
p-schuler, line 48: correction: we can actually limit the impact. We know if we are the connection owner for that device and can act accordingly. This is a very simple case and we should implement it (close the connection, if we don't own it).

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+
+Version, LNS discovery and CUPS update endpoints are not affected by this issue.
+
+## Possible solution
+
+The main idea is to give the current connection holder (as indicated from the Function), the edge to
+continue processing messages for this device. The performance of that gateway will not be impacted.
+
+A local (per LNS) in-memory dictionary will be used for that purpose to map DevEuis to a flag
+indicating whether the LNS is the preferred gateway or not.
```
p-schuler, line 82: I would re-write this to broader strokes and not implementation details (use of a dictionary):
```suggestion
The main idea is to give the current connection owner (as indicated from the Function), the edge to
continue processing messages for this device. The performance of that gateway will not be impacted.

If the current connection owner will keep the connection open. Any other network servers receiving messages, will not maintain an active connection to IoT Hub.

If the owning network server stops responding or gets out of reach, the ownership is transferred to the next winning network server.

```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+For now we don't make changes here and accept there will be a potential connection switching
+periodically (currently every 2 days if the device was not refreshed in the meantime). This will be
+revisited with either [this issue about the frame counter
+  drift](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1499)
+ (one of the solutions there is to fetch twin and close connection immediately) or [this issue
+ specific to cache refresh](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1505)
+
+#### Alternatives
+
+- Contact the Function to see whether we are the owning gateway. If we are, refresh the twin (and
```
p-schuler, line 101: no need to contact the function: we know if we are the owner with the state on the LoRaDevice in memory.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+
+Problem
+
+- Join received from LNS1 and LNS2: LNS1 wins
+- First data message received only on LNS2, Function should inform LNS1 that it's the losing one
+  (information about the owning LNS needs to be "shared" between joins and data messages).
+
+#### Decision
+
+We decide to go with the simpler solution of closing the connection immediately after a Join and let
```
p-schuler, line 128: The decision to close the connection is not only for the join request, but for any initialization operation. When we load the device into the cache for the first time, we need to load the twins. No matter if it is a join or not, we need to close the connection, because there is no "leader election pattern" on the load operations. We do not know if we own the connection after that operation, potentially leaving the connection open. Only data message flow, does determine the ownership of the connection and maintains it.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+the data flow re-establish it if needs to be. Disadvantage is that the first data message is more likely
+  to miss the window due to having re-establish the connection.
+  LNS also stores locally that it was the losing LNS and delays on a future re-join or future data messages.
+
+#### Alternative
+
+- Do nothing and accept that a one-off connection stealing on the first Data message can happen.
```
p-schuler, line 135: This would only work, if you could guarantee that the first data message reaches the gateway the device joined on. Since we can't, we can't rely on the data flow to fix this problem.

p-schuler: The only alternative I see here is that during the nonce evaluation, we mark the owning gateway in the function and track it that way. Due to added complexity we discarded that though.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+where Device sends data message A and then B.
+
+Here is a rundown of what should happen assuming both LNSs have a fresh twin (via background refresh
+or via fetching it using DeviceGetter.GetDevice). Changes are marked in **bold**:
+
+1. Device sends first data message A.
+1. We assume that LNS1 gets the message first. **LNS1 checks against
+   its in-memory dictionary DevEui -> flag** and since it has not seen this DevEui before contacts the Function.
```
p-schuler, line 178: Too much detail.
```suggestion
1. We assume that LNS1 gets the message first. **LNS1 checks against
   its in-memory state** and since an owner for the connection was not elected yet, directly contacts the function without delay.
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+1. The Function hasn't seen this DevEui either and therefore does not have an assigned LNS for it
+  yet. LNS1 wins the race and gets immediately a response and processes the message upstream.
+1. LNS2 eventually receives message A, **checks its local dictionary** and also contacts the
```
p-schuler, line 181:
```suggestion
1. LNS2 eventually receives message A, **checks its local state** and also contacts the
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+   FunctionBundler immediately since it does not have prior info about this DevEui.
```
p-schuler, line 182:
```suggestion
   FunctionBundler immediately since it does not have prior info about this device.
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+1. The Function responds to LNS2 that it lost the race to process this message.
+1. Since deduplication strategy is Drop, LNS2 drops the message immediately, therefore no
+   connection to Iot Hub is opened and only LNS1 has the connection to Iot Hub. **LNS2 notes in
+   memory that it was the losing gateway for this DevEui**.
```
p-schuler, line 186:
```suggestion
   connection to Iot Hub is opened and only LNS1 has the connection to Iot Hub. **LNS2 updates its
   in memory state that it does not own the connection for this device**.
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+1. When message B gets send (with a higher frame counter*), assuming that this time LNS2 gets it
+   first it **checks again its local dictionary it's not the preferred LNS for this device and
```
p-schuler, line 188:
```suggestion
   first it **checks again its local state that indicates it's not owning the connection for the device and
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+   therefore delays itself X ms before contacting the FunctionBundler**.
+   - Here we do *not* want to simply drop the message as LNS1 might not be available anymore (due to
+     a crash, device not in range etc).
+   - Since it was the losing LNS previously it should **make sure that it operates on a fresh twin**
```
p-schuler, line 192: this is relevant only for the fcnt drift we discussed. We will keep the background refresh of the twin also on the not owning device to ensure it's refreshed periodically. The connection will be opened only temporarily and closed after the refresh.

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+     (especially regarding the frame counter) see also [background refresh section](#handling-of-cache-refresh)
+1. This delay gives LNS1 a time advantage to reach the FunctionBundler first and win the race again, failing
+    back to the previous case of message A. The active connection stays with LNS1.
+1. If this delay is not sufficient for LNS1 to win the race, LNS2 will contact the FunctionBundler which
+   now awards LNS2 as the "winning" LNS. LNS2 can now process the message upstream. It also **removes the "losing flag" from its in-memory store**.
+1. **The Function also proactively informs LNS1** that it's not anymore the winning LNS for this device.
+   - The reason why we do this proactively is that otherwise its Edge Hub will try to reconnect to IoT Hub even if there is no more messages picked up from LNS1 (out of range roaming client).
```
p-schuler, line 199:
```suggestion
   - The reason why we do this is to ensure, LNS1 knows connection ownership was transferred to another network server and it can drop the connection. This is for the case, where the upstream message does not reach LNS1.
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+  - For that we can use a C2D message to LNS. Alternative: direct method could be used here but
```
p-schuler, line 200:
```suggestion
   - For that we can use a C2D message to LNS. Alternative: direct method could be used here but
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,286 @@
...
+    delivery will not be guaranteed then.
+1. If LNS1 in the meantime gets message B and contacts the FunctionBundler, it will let it know
+   that it lost the race for this frame counter and must therefore drop the message, **mark
+   itself as the losing LNS and close the connection if it hasn't done so yet**.
+
+Notes:
+
+- The FunctionBundler is not called in certain topologies e.g. when multiple LBSs are connected to the same LNS but these topologies are not relevant for the issue here).
```
p-schuler, line 208:
```suggestion
- The FunctionBundler is not called in certain topologies e.g. when multiple LBSs are connected to the same LNS but these topologies are not relevant for the issue here as they employ a single connection per device by design).
```

docs/adr/009_lns_affinity.md
spygi: I had it here because the problem is more pronounced when using the Edge module but will remove since it doesn't go away completely even in direct mode.

docs/adr/009_lns_affinity.md
spygi: ok, I will add it back to in-scope then and added a task for it in #1473

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,284 @@
...
+Notes:
+
+- The FunctionBundler is not called in certain topologies e.g. when multiple LBSs are connected to the same LNS but these topologies are not relevant for the issue here as they employ a single connection per device by design).
+- For the main flow above we consider only frame counter B > frame counter A. Resets are covered in
+  [the reset section](#handling-of-abp-relax-frame-counter-reset). Resubmits (when frame counter B ==
+  frame counter A) are also a current issue and should be [addressed in this issue](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1468).
+
+### Handling of ABP relax frame counter reset
+
+An special case on the main data flow is when we detect a device reset after a message. Currently we save the twin immediately and then clear
```
p-schuler, line 207:
```suggestion
A special case on the main data flow is when we detect a device reset after a message. Currently we save the twin immediately and then clear
```

docs/adr/009_lns_affinity.md
```diff
@@ -0,0 +1,284 @@
...
+
+The delay amount should be configurable to allow users to customize behavior for their scenarios.
+During load testing we tested with 400ms but smaller values should be tried as well. If 0ms are
+specified the stickiness feature is disabled which means potential connection switching.
+
+## Other candidates considered
+
+### Single point of connection handling on LoRaDevice
+
+The changes presented from Atif: would also ensure by design that new code does not open more
```
p-schuler, line 266: That's weird to put under "Other candidates considered" - I would put this under a related section to indicate that this is additional.

p-schuler (changes requested):
made a few suggestions. Let me know if you would like to go over them.

docs/adr/009_lns_affinity.md
spygi: Agree. Not sure if you request a change here?

If you would like the wording to be more generic: this section is specifically for the joins but I made sure to clarify that we close the connection also on the background refresh and cache initialization cases.

docs/adr/009_lns_affinity.md
spygi: Agree, moved to a related section

docs/adr/009_lns_affinity.md
spygi: correct, removed this now (104740f )

docs/adr/009_lns_affinity.md
spygi: Fixed (104740f)

docs/adr/009_lns_affinity.md
spygi: Resolving now.

docs/adr/009_lns_affinity.md
p-schuler: All good - was referring to the close when initializing for a data message, which is covered later.

p-schuler (approved):
LGTM - good investigation!

Issue #1479, opened 2022-02-09
Refactoring of device queue - ADR


p-schuler:
If we do not create a single queue / device we want to change the way we do cache refreshes: #990

## Acceptance Criteria

- ADR published to docs
- Estimation on impact, depending on this Implementation stories on backlog

PR #1494, opened 2022-02-15
Add ADR on single device queue design

atifaziz:
# PR for issue #1479

## What is being addressed

The potential cost of refactoring for a single device per queue to avoid connection ping-pong between a device, multiple LNS via Edge Hub and IoT Hub.

## How is this addressed

Published an ADR based on approaches explored in a spike and the decision taken.

### DIFF
--- a/docs/adr/010_device_request_queue.md
+++ b/docs/adr/010_device_request_queue.md
@@ -84,32 +84,38 @@ be summed up as follows:
 
 ## Decision
 
-Use the approach to simplify the entire flow and processing of a message into
-a simple request-response model. A quick spike demonstrated that the changes
-to the main code base would be fairly contained and the largest impact is
-expected to be in adapting the tests (which could also be done with a stop-gap
-measure where the tests are adapted after the initial refactoring of the code
-base).
+All operations for a particular device requiring the IoT Hub connection are to
+be executed in a serial/exclusive fashion on the `LoRaDevice` itself. This
+will ensure that there is only one operation acting on the connection at any
+given time. Making each execution exclusive (think queue) gives a
+deterministic way of adding a close operation without affecting any other
+operations that may be in flight.
 
-All operations for a particular device requiring the IoT hub connection, are
-requested to be executed on the `LoRaDevice` itself. The `LoRaDevice` becomes
-a singleton. This will ensure, we only have one operation acting on the
-connection and allows us to have a deterministic way of adding a close
-operation without affecting any other operations currently being processed.
-The approach has a problem with the creation of the devices. A `LoRaDevice`
-requires the twins to be valid. That operation can't be synchronized on the
-`LoRaDevice` as that instance is technically not available. Since it is not
-available, there should be no operation coming in, or those should also
-trigger the load. A simple solution is to synchronize both load operations in
-the `LoRaDeviceRegistry` and make sure there is only ever 1 operation loading
-the twins for a particular DevEui. We also concluded that closing the
-connection after initialization is the easiest approach to delay the
-connection ownership decision to after the first message has arrived.
+Unfortunately, the initialization of a `LoRaDevice` cannot be addressed with
+the same approach since it requires a two-step initialization per its current
+design. That is, a `LoRaDevice` is _minimally initialized_ when created so its
+`InitializeAsync` method can be called to fetch the twin information. Until
+that completes, a `LoRaDevice` is not considered technically ready for being
+put in the cache. Since a `LoRaDevice` cannot be discovered during its
+initialization, it cannot be synchronized with the other operations. To work
+around this,m load operations will be synchronized in the `LoRaDeviceRegistry`
+to ensure there is only ever one operation loading the twins for a particular
+DevEUI. It was also decided that closing the connection after initialization
+is the easiest approach to delay the connection ownership decision to after
+the arrival of the first message.
 
-Based on a spike of the refactoring this would require, it seems plausible to
-achieve the refactoring of the actual code base within two weekly sprints. It
-is expected that the bulk of the time will be spent in refactoring the test
-code.
+The approach to simplify the entire flow and processing of a message into a
+simple request-response model will be taken a stretch goal so that the
+connection management can be implemented and tested earlier. A quick spike
+demonstrated that the changes to the main code base would be fairly contained
+and the largest impact is expected to be in adapting the tests (which could
+also be done with a stop-gap measure where the tests are adapted after the
+initial refactoring of the code base).
+
+Based on a spike of the changes the simple request-response model would
+require, it seems plausible to achieve the refactoring within (at most) two
+weekly sprints. It is expected that the bulk of the time will be spent in
+refactoring the test code.
 
 ## Consequences
 
--- a/docs/adr/010_device_request_queue.md
+++ b/docs/adr/010_device_request_queue.md
@@ -5,13 +5,13 @@
 
 ## Context
 
-LoRa devices, which can reach multiple LNS, pose an interesting challenge for
+LoRa devices, which can reach multiple LNSs, pose an interesting challenge for
 Edge Hub since Edge Hub wasn't designed for devices competing for gateways.
 When a device connection is made to IoT Hub, Edge Hub continues to maintain
 that connection in the open state (for network resiliency) until explicitly
 closed. Moreover, IoT Hub allows maintining only a single connection per
-device. When another is opened, the existing one closed. If a device can reach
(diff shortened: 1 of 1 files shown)


docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,164 @@
...
+tries to open a connection for the same device, IoT Hub closes the connection
+for the other. But then the other tries to re-open the connection and the
+ping-pong continues. This problem needs addressing to avoid scalability
+limitations.
+
+The ping-pong can be pevented by electing a leader LNS for a device and
+announcing the decision to others using a C2D message. An LNS receiving such a
+message will then close that connection so that Edge Hub does't continue to
+maintain the connection in the open state through retires. The handling of the
+message needs to know:
```
p-schuler, line 26:
```suggestion
The ping-pong can be prevented by electing a leader LNS for a device through
the function race determination and as a fallback announcing the decision to 
others using a C2D message. An LNS receiving such a message will then close 
that connection so that Edge Hub doesn't continue to maintain the connection 
in the open state through retires. The handling of the message needs to know:
```

p-schuler: @atifaziz - btw, this is unfortunately a good example that the line length limitation does not really work with determining the diff easily. If you need to insert something and have to adjust the alignment, things will end up
being the same.

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,164 @@
...
+- A data message; `DeviceLoaderSynchronizer` loads the twins for multiple
+  devices based on the `DevAddr` and adds them to the cache.
+
+- C2D message that needs to be delivered can trigger a load if the device is
+  not in the cache.
+
+- Cache refreshes can happen on a background thread for a device, causing the
+  need for a connection.
+
+**Note**: We should consider the option to keep building on top of the connection
```
p-schuler, line 49: this was more a comment for you @atifaziz that we may want to list that as an option and consider it for the solution?

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,164 @@
...
+device named A with a `DevAddr` of 1 (from a previous join) using OTAA
+re-joins. Since the device is not in the cache, the LNS issues a request to
+the server function to search for the device on IoT Hub and obtain its primary
+key that's needed to open a device connection. Next, a `LoRaDevice` is
+_minimally initialzed_ for the purpose of loading its twin and the twin data
+requested over the device conneciton. During this time, suppose that another
+device named B sharing the same `DevAddr` as device A but using ABP activation
+sends an uplink message to the LNS. Since the `DevAddr` is not in the cache,
+the LNS will issues a search request to the server function to return all
+devices using the `DevAddr` in question.
```
p-schuler, line 103:
```suggestion
devices using the `DevAddr` in question eventually causing a load of the device twins for the same DevEui.
```

atifaziz: Sorry, this was an older version that should have never been committed and was superseded by 60b5aaffe328f6e3b6dcfeeacdfb8ea02f0bb92b by you.

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,149 @@
...
+   return `Task<DownlinkMessage>` instead of `void` or `Task`. This allows all
+   `Task.Run` uses to be moved to a single and central point when a message is
+   received instead of being littered throughout the code base. The main
+   message loop could also be made responsible for central
+   error-handling/logging and sending of downlink messages when the processing
+   of a request has completed. This would just require regular use of tasks.
+   Next, all `Task`-based operations on `LoRaDevice` could be naturally and
+   _implicitly_ queued and then executed in a mutually exclusive manner.
+
+## Decision
```
p-schuler, line 86: the decision needs to be updated with what we discussed yesterday.

p-schuler:
Why did you decide to remove all the context about the issues we discovered in 5cca96cdd49bfe53b7e85d192a4c472271c95fb0?

atifaziz:
> Why did you decide to remove all the context about the issues we discovered in 5cca96cdd49bfe53b7e85d192a4c472271c95fb0?

It was an accidental commit of an incomplete description, spoke of only one very specific issue (whereas there are many more) and a lot of it seemed redundant with what seems to be already captured in PR #1483 at a higher-level. I think we can cross-reference here.

docs/adr/010_device_request_queue.md
atifaziz: Done with b70734e222bc36fe66f22d6da59c7bfc8dee74dd.

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,154 @@
...
+
+The following cases can trigger operations from different threads:
+
+- An OTAA join message; `DeviceJoinLoader` eventually loads the twins and
+  stores them back up.
+
+- A data message; `DeviceLoaderSynchronizer` loads the twins for multiple
+  devices based on the `DevAddr` and adds them to the cache.
+
+- C2D message that needs to be delivered can trigger a load if the device is
```
danigian, line 42: We are talking about Class C C2D messages here, correct? If yes, let's write it down please

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,154 @@
...
+operations that may be in flight.
+
+Unfortunately, the initialization of a `LoRaDevice` cannot be addressed with
+the same approach since it requires a two-step initialization per its current
+design. That is, a `LoRaDevice` is _minimally initialized_ when created so its
+`InitializeAsync` method can be called to fetch the twin information. Until
+that completes, a `LoRaDevice` is not considered technically ready for being
+put in the cache. Since a `LoRaDevice` cannot be discovered during its
+initialization, it cannot be synchronized with the other operations. To work
+around this,m load operations will be synchronized in the `LoRaDeviceRegistry`
```
danigian, line 101:
```suggestion
around this, load operations will be synchronized in the `LoRaDeviceRegistry`
```

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,154 @@
...
+to ensure there is only ever one operation loading the twins for a particular
+DevEUI. It was also decided that closing the connection after initialization
+is the easiest approach to delay the connection ownership decision to after
+the arrival of the first message.
+
+The approach to simplify the entire flow and processing of a message into a
+simple request-response model will be taken a stretch goal so that the
```
danigian, line 108:
```suggestion
simple request-response model will be taken as stretch goal so that the
```

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,154 @@
...
+- Use of the regular async-await programming model, including error-handling
+  and cancellation that's built-in into tasks.
+
+- Transparent use of queues through tasks in `LoRaDevice` to serialize
+  execution. This subsumes the first approach (discussed in the introductory
+  section) by having implicit rather explicit queuing (at the
+  application-level).
+
+- Greatest potential to simplify tests since assertions can rely on simply
+  return values and exceptions.
```
danigian, line 143:
```suggestion
  return values and exceptions, instead of success/failure notifications.
```

The reader might not know what the current status is for assertions. What do you think about writing it down?

docs/adr/010_device_request_queue.md
```diff
@@ -0,0 +1,154 @@
...
+- It is not mutually exclusive with other approaches explored. For example, by
+  simplifying to the request-response model, it would be even easier to have
+  the LNS operate with a single logical thread if that could further help
+  remove some complexity (without compromising scalability) like locks.
+
+The absence of an explicit queue could be make it more difficult for someone
+to understand the code and choices made if they are not familiar with the
+intricacies of how async-await operates.
+
+[#1480]: https://github.com/Azure/iotedge-lorawan-starterkit/issues/1479
```
danigian, line 154: Is this 1480 or 1479? :)

atifaziz: Thanks & fixed with 7de1eb5bce765d30b66877879a4a55b548519dfa.

docs/adr/010_device_request_queue.md
atifaziz: Thanks. Actually the fix here was that "as" was just missing so I added it with 80be0392ea9700fe185dd718f228b4ff1775fb32.

p-schuler (approved):

docs/adr/010_device_request_queue.md
atifaziz: Clarified with a6e194f43784a80c4c8a96a6d242c75cd6e32a21.

Issue #1489, opened 2022-02-14
[Story] Implement standalone discovery endpoint


bastbu:
## Description

Because of #1484, we will add support for a standalone discovery endpoint. We will continue to maintain the existing `/router-info` discovery endpoints on each LNS. The standalone discovery endpoint will be implemented as decided in the ADR.

## Acceptance criteria

Functionality:
- [x] ASP.NET Core Application that exposes the `router-info` endpoint
- [x] Consumes configuration (e.g. tags) from IoT Hub device twins to assign LBS to LNS
- [x] Uses in-memory round robin distribution to choose the next LNS to which an LBS should connect
- [x] No health check strategy is used
- [x] Can be configured to use a specific certificate
- [ ] (stretch) Supports client authentication. Since this is implemented on the LNS data endpoint in any case, it offers little benefit to have it on the discovery service as well.

Code quality:
- [x] Code is being built and tested during each CI run
- [x] Code style warnings and nullability are enabled
- [x] #1515
- [ ] E2E tests which are disabled on the pipeline and can be run manually before each release (which test using a real LBS) -> tracked in #1527

Observability:
- [x] (stretch) Expose metrics in Prometheus format
- [x] #1512
- [ ] (stretch) Initialize `cloud_roleName` for Application Insights telemetry

Docs:
- [x] How to deploy the discovery endpoint on-premises
- [x] How to configure which LNS are in reach of which LBS
- [x] How to monitor the discovery service endpoint

PR #1507, opened 2022-02-16
Basic version of standalone discovery service

bastbu:
# PR for issue #1489

## What is being addressed

With this PR we introduce a basic version of the standalone discovery service. It does not yet contain all the functionality that it will eventually. A rough overview of the functionalities contained in this PR:

- Dynamic LNS discovery based on twin tags: By setting a `network` tag to identify LBS/LNS that are within the same network boundaries, we can associate LNS that are in reach for a specific LBS. An additional configuration requirement is that each LNS module twin that is discoverable via the discovery service must contain a `hostAddress` desired property that identifies its DNS/IP (including scheme and port)
- Backwards compatibility with the existing discovery endpoint.
- Caching of the list of LNS for a given network for 6 hours (not configurable)
- Support for managed identities/connection strings that can be used for auth against IoT Hub

## How is this addressed

The discovery service is a new module, which is called `LoRaWan.NetworkServerDiscovery`. The LNS module references the discovery service project to reuse the service discovery logic. Apart from unit tests, integration tests test all the components without introducing a dependency on IoT Hub. Testing against a real IoT Hub would complicate the integration test setup for the pipelines and the developer setup, while providing little value other than what the E2E tests are going to cover in the future. Some features that are not yet implemented, but will follow in the future:

- Client authentication support
- Observability (integration with Prometheus and Application Insights)
- E2E tests


LoRaEngine/modules/LoRaWanNetworkSrvModule/Dockerfile.amd64
```diff
@@ -8,6 +8,11 @@ COPY global.json ./
 COPY LICENSE ./
 COPY ./LoRaEngine/modules/LoRaWanNetworkSrvModule/NOTICE.txt ./
 
+WORKDIR /build/LoRaEngine/modules
```
danigian, line 11: Why is this needed? Isn't the dotnet restore executed anyways from the LoRaWanNetworkSrvModule as the Discovery project is being referenced in its csproj?

danigian: Ditto for other dockerfiles

LoRaEngine/modules/LoRaWan.NetworkServerDiscovery/TagBasedLnsDiscovery.cs
```diff
@@ -0,0 +1,152 @@
...
+                                         select HostAddressReader.Read(hostAddressInfo) into uri
+                                         where uri != null
+                                         select uri);
+                    }
+
+                    this.logger.LogInformation("Loaded {Count} LNS candidates for network '{NetworkId}'", results.Count, networkId);
+
+                    // Also cache if no LNS URIs are found for the given network.
+                    // This makes sure that rogue LBS do not cause too many registry operations.
+                    return results;
```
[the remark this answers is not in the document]
bastbu: Good point, I'll add that to the documentation (once I get to that PR)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/Processors/LnsProtocolMessageProcessor.cs
```diff
@@ -47,99 +49,34 @@ internal class LnsProtocolMessageProcessor : ILnsProtocolMessageProcessor
...
-                    this.logger.LogDebug(ex, ex.Message);
-                }
-            }
-            catch (Exception ex) when (ExceptionFilterUtility.False(() => this.logger.LogError(ex, "An exception occurred while processing requests: {Exception}.", ex),
-                                                                    () => this.unhandledExceptionCount?.Add(1)))
-            {
-                throw;
-            }
+                    Scheme = httpContext.Request.IsHttps ? "wss" : "ws",
+                    Port = httpContext.Request.Host.Port ?? 80,
```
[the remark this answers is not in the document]
bastbu: I made the behavior more explicit in 4f3fc6048787014d5dfd4e7382f37fff7fa740da. I am continuing to use the default behavior - if no port is specified, the defaults of the protocol should be used. I also added tests for it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/Dockerfile.amd64
bastbu: I would still need to switch back to copy over the discovery service files after the restore has executed, no? I would be able to get rid of one restore, but would need another workdir switch instead. Or am I missing something?

danigian: Sorry, I was referring explicitly to the dotnet restore. I would avoid operation on rows 12/13 and just switch workdir and then "COPY ./LoRaEngine/modules/LoRaWan.NetworkServerDiscovery LoRaWan.NetworkServerDiscovery"

bastbu: Addressed in 1cd6ab6fe18b6fe55c1f325e37c75f946b0bb6db

danigian (approved):
Tested manually with a Basic Station and everything works as expected.
Thanks!
LGTM

Issue #1501, opened 2022-02-15
Implement device operation synchronization 


p-schuler:
As per the decision in the [ADR], all operations on a device connection must be synchronized and where connections and disconnections can be done in a controllable manner. For example, a request to disconnect should be synchronized with any other operation that is in flight such that the disconnection occurs after the in-flight operation is done.

[ADR]: https://github.com/Azure/iotedge-lorawan-starterkit/pull/1494

## Acceptance Criteria

- [x] All operations on the device client connection are synchronized
- [x] Disconnection is just another operation such that it is naturally synchronized with other operations
- [x] Optionally, disconnection is deferred if there is a _connection activity_ and until all activities are done
- [x] Existing unit tests are adjusted
- [x] New code is unit tested

p-schuler:
@atifaziz pls provide details and ACs.

PR #1517, opened 2022-02-18
Synchronize device operations

atifaziz:
# PR for issue #1501

## What is being addressed

- Synchronization of device operations over the connection
- Avoiding connection closes while the connection is in use by an operation

## How is this addressed

All device operations are synchronized such that only one operation uses the device connection at any given time. The way this is done is that `LoRaDeviceClientConnectionManager` wraps the original `ILoRaDeviceClient` implementation with one that synchronizes around each asynchronous operation such that only one executes at any given time. The private wrapper (called `SynchronizedLoRaDeviceClient`) is what's returned from the `GetClient` method. This strategy allowed most changes to be concentrated closest to the management of the client connection. The following asynchronous operations synchronized are:

- `GetTwinAsync`
- `SendEventAsync`
- `UpdateReportedPropertiesAsync`
- `ReceiveAsync`
- `CompleteAsync`
- `AbandonAsync`
- `RejectAsync`
- `DisconnectAsync` (used to synchronous previously)

The only synchronous method that's not synchronized with the above is `EnusureConnected` since it's not a true operation on the client connection (that is, one that _uses_ the underlying connection). Note also that `DisconnectAsync` used to synchronous previously, but is now asynchronous because it's synchronized with other operations of the client.

The actual synchronization code is implemented in a separate, general and and generic, class called `ExclusiveProcessor<>`. It simply creates a semaphore of 1 that can be awaited. It also raises a number of events that are used by `SynchronizedLoRaDeviceClient` for the purpose of logging. Below is a sample of what such logging looks like:

    Queued "GetTwinAsync" (1)
    Invoking "GetTwinAsync" (1)
    Invoked "GetTwinAsync" (1); status = RanToCompletion, run-time = 00:00:00.0195538, wait-time = 00:00:00.0055340
    Queued "DisconnectAsync" (2)
    Invoking "DisconnectAsync" (2)
    DisconnectAsync = Disconnected
    Invoked "DisconnectAsync" (2); status = RanToCompletion, run-time = 00:00:00.0063517, wait-time = 00:00:00.0001069

The most complex area of `SynchronizedLoRaDeviceClient` is with respect to disconnection and activity management. An _activity_ allows a client connection to be _reserved for multiple operations_. An activity is started via `BeginDeviceClientConnectionActivity` and `SynchronizedLoRaDeviceClient` keeps track of how many activities are outstanding. Each activity is represented by a disposable object, which when disposed, decrements the activity counter. While activities are outstanding on the client connection, calls to `DisconnectAsync` are deferred until the activity counter drops to zero and then a single disconnection is issued to the actual and underlying/wrapped client connection. The activity design used to exist on `LoRaDevice` previously, but is now moved into `SynchronizedLoRaDeviceClient` where it's easier to manage and reason about in one place.

`SynchronizedLoRaDeviceClient` also calls `EnsureConnected` for all asynchronous operations except `DisconnectAsync`. If `EnsureConnected` succeeds, the client is cached with a sliding expiration where it will be disconnected at the time of cache entry eviction. Again, if there are outstanding activities then this disconnection will be deferred until all activities have ended (via disposal of the activity object).

## ~~Pending work~~

The following work is pending before this PR can be put out of draft:

- [x] Address ABP E2E tests that are failing (currently only `LoRaWan.Tests.E2E.ABPTest.Test_ABP_Device_With_Connection_Timeout`)
- [x] Add unit tests (to the extent possible) to exercise the new synchronized operations and deferred disconnection when activities are outstanding
- [x] Re-work some of the tests removed on `LoRaDevice` because the functionality has been moved into `LoRaDeviceClientConnectionManager` and `SynchronizedLoRaDeviceClient`.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
@@ -53,6 +53,8 @@ namespace LoRaWan.NetworkServer
 #pragma warning restore CA1062 // Validate arguments of public methods
         }
 
+#pragma warning restore CA1034 // Nested types should not be visible
+
         private sealed record ProcessingTaskOutcome<TResult> :
             ProcessingOutcome<Task<TResult>>,
             IProcessingOutcome
@@ -66,8 +68,6 @@ namespace LoRaWan.NetworkServer
 
             Task IProcessingOutcome.Task => Result;
         }
-
-#pragma warning restore CA1034 // Nested types should not be visible
 
         public async Task<ProcessingOutcome<TResult>> ProcessAsync<TResult>(T processor, Func<Task<TResult>> function)
         {
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
@@ -78,7 +78,7 @@ namespace LoRaWan.NetworkServer
 
         public async Task<ProcessingOutcome<Task<TResult>>> TryProcessAsync<TResult>(T processor, Func<Task<TResult>> function)
         {
-            if (function == null) throw new ArgumentNullException(nameof(function));
+            ArgumentNullException.ThrowIfNull(function, nameof(function));
 
             var submissionTime = DateTimeOffset.Now;
 
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
@@ -94,7 +94,7 @@ namespace LoRaWan.NetworkServer
         {
             ArgumentNullException.ThrowIfNull(function, nameof(function));
 
-            var submissionTime = DateTimeOffset.Now;
+            var submissionTime = DateTime.UtcNow;
 
             lock (this.queue)
                 this.queue.Add(processor);
@@ -125,9 +125,9 @@ namespace LoRaWan.NetworkServer
                     else
                     {
                         Processing?.Invoke(this, processor);
-                        var startTime = DateTimeOffset.Now;
+                        var startTime = DateTime.UtcNow;
                         var task = await Task.WhenAny(function());
-                        var endTime = DateTimeOffset.Now;
+                        var endTime = DateTime.UtcNow;
                         var outcome = new ProcessingTaskOutcome<TResult>(task, submissionTime, startTime - submissionTime, endTime - startTime);
                         Processed?.Invoke(this, (processor, outcome));
                         return outcome;
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
@@ -19,6 +19,7 @@ namespace LoRaWan.NetworkServer
         private readonly SemaphoreSlim processingLock = new(1);
         private readonly List<T> queue = new();
 
+        public event EventHandler<T>? Submitted;
         public event EventHandler<(T InterruptedProcessor, T InterruptingProcessor)>? Interrupted;
         public event EventHandler<T>? Processing;
         public event EventHandler<(T Processor, IProcessingOutcome Outcome)>? Processed;
@@ -83,6 +84,8 @@ namespace LoRaWan.NetworkServer
 
             lock (this.queue)
                 this.queue.Add(processor);
+
+            Submitted?.Invoke(this, processor);
 
             while (true)
             {
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
@@ -21,7 +21,7 @@ namespace LoRaWan.NetworkServer
 
         public event EventHandler<(T InterruptedProcessor, T InterruptingProcessor)>? Interrupted;
         public event EventHandler<T>? Processing;
-        public event EventHandler<IProcessingOutcome>? Processed;
+        public event EventHandler<(T Processor, IProcessingOutcome Outcome)>? Processed;
 
         public ExclusiveProcessor() : this(ps => ps[0]) { }
 
@@ -112,7 +112,7 @@ namespace LoRaWan.NetworkServer
                         var task = await Task.WhenAny(function());
                         var endTime = DateTimeOffset.Now;
                         var outcome = new ProcessingTaskOutcome<TResult>(task, submissionTime, startTime - submissionTime, endTime - startTime);
-                        Processed?.Invoke(this, outcome);
+                        Processed?.Invoke(this, (processor, outcome));
                         return outcome;
                     }
                 }
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
@@ -14,7 +14,7 @@ namespace LoRaWan.NetworkServer
 
     public sealed class ExclusiveProcessor<T>
     {
-        private readonly Func<IReadOnlyList<T>, T> scheduler;
+        private readonly IScheduler scheduler;
         private readonly IEqualityComparer<T> comparer;
         private readonly SemaphoreSlim processingLock = new(1);
         private readonly List<T> queue = new();
@@ -24,11 +24,25 @@ namespace LoRaWan.NetworkServer
         public event EventHandler<T>? Processing;
         public event EventHandler<(T Processor, IProcessingOutcome Outcome)>? Processed;
 
-        public ExclusiveProcessor() : this(ps => ps[0]) { }
+#pragma warning disable CA1034 // Nested types should not be visible (tightly related)
+        public interface IScheduler
+#pragma warning restore CA1034 // Nested types should not be visible
+        {
+            T SelectNext(IReadOnlyList<T> processes);
+        }
 
-        public ExclusiveProcessor(Func<IReadOnlyList<T>, T> scheduler) : this(scheduler, null) { }
+        public static readonly IScheduler DefaultScheduler = new FifoScheduler();
 
-        public ExclusiveProcessor(Func<IReadOnlyList<T>, T> scheduler, IEqualityComparer<T>? comparer)
+        private sealed class FifoScheduler : IScheduler
+        {
+            public T SelectNext(IReadOnlyList<T> processes) => processes[0];
+        }
+
+        public ExclusiveProcessor() : this(DefaultScheduler) { }
+
+        public ExclusiveProcessor(IScheduler scheduler) : this(scheduler, null) { }
+
+        public ExclusiveProcessor(IScheduler scheduler, IEqualityComparer<T>? comparer)
         {
             this.scheduler = scheduler;
             this.comparer = comparer ?? EqualityComparer<T>.Default;
@@ -97,7 +111,7 @@ namespace LoRaWan.NetworkServer
 
                     lock (this.queue)
                     {
-                        var nextProcessor = this.scheduler(this.queue);
+                        var nextProcessor = this.scheduler.SelectNext(this.queue);
                         if (!this.comparer.Equals(nextProcessor, processor))
                             next = (true, nextProcessor);
                         else
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
@@ -1,0 +1,126 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+#pragma warning disable CA1003 // Use generic event handler instances (suppressed for performance)
+
+namespace LoRaWan.NetworkServer
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public sealed class ExclusiveProcessor<T>
+    {
+        private readonly Func<IReadOnlyList<T>, T> scheduler;
+        private readonly IEqualityComparer<T> comparer;
+        private readonly SemaphoreSlim processingLock = new(1);
+        private readonly List<T> queue = new();
+
+        public event EventHandler<(T InterruptedProcessor, T InterruptingProcessor)>? Interrupted;
+        public event EventHandler<T>? Processing;
+        public event EventHandler<IProcessingOutcome>? Processed;
+
+        public ExclusiveProcessor() : this(ps => ps[0]) { }
+
+        public ExclusiveProcessor(Func<IReadOnlyList<T>, T> scheduler) : this(scheduler, null) { }
+
+        public ExclusiveProcessor(Func<IReadOnlyList<T>, T> scheduler, IEqualityComparer<T>? comparer)
+        {
+            this.scheduler = scheduler;
+            this.comparer = comparer ?? EqualityComparer<T>.Default;
+        }
+
+#pragma warning disable CA1034 // Nested types should not be visible (by design)
+
+        public interface IProcessingOutcome
+        {
+            Task Task { get; }
+            DateTimeOffset SubmissionTime { get; }
+            TimeSpan WaitDuration { get; }
+            TimeSpan RunDuration { get; }
+        }
+
+        public record ProcessingOutcome<TResult>(TResult Result,
+                                                 DateTimeOffset SubmissionTime,
+                                                 TimeSpan WaitDuration,
+                                                 TimeSpan RunDuration)
+        {
+#pragma warning disable CA1062 // Validate arguments of public methods (operators don't usually throw)
+            public static implicit operator TResult(ProcessingOutcome<TResult> outcome) => outcome.Result;
+#pragma warning restore CA1062 // Validate arguments of public methods
+        }
+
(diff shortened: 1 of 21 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
```diff
@@ -1,162 +1,399 @@
...
+                    if (this.logger is { } logger && logger.IsEnabled(LogLevel.Debug))
+                    {
+                        const string message = $"{nameof(BeginDeviceClientConnectionActivity)} (disposal) = {{Activities}}";
+                        logger.LogDebug(message, activities);
+                    }
+
+                    if (activities > 0)
+                        return;
+
+                    _ = await DisconnectAsync(deferred: true);
```
p-schuler, line 312: why would we disconnect if we do not have any more activities? The owning LNS should not drop the connection.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
```diff
@@ -0,0 +1,129 @@
...
+#pragma warning disable CA1003 // Use generic event handler instances (suppressed for performance)
+
+namespace LoRaWan.NetworkServer
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public sealed class ExclusiveProcessor<T>
```
p-schuler, line 15: Is it possible that this does not have any unit tests?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
```diff
@@ -0,0 +1,129 @@
...
+        public event EventHandler<T>? Processing;
+        public event EventHandler<(T Processor, IProcessingOutcome Outcome)>? Processed;
+
+        public ExclusiveProcessor() : this(ps => ps[0]) { }
+
+        public ExclusiveProcessor(Func<IReadOnlyList<T>, T> scheduler) : this(scheduler, null) { }
+
+        public ExclusiveProcessor(Func<IReadOnlyList<T>, T> scheduler, IEqualityComparer<T>? comparer)
+        {
+            this.scheduler = scheduler;
```
p-schuler, line 33: for me it is super hard to read the code easily with the function callback for the scheduler. If I want to understand what schedulers we have implemented, VS can't help me. I need to find all references to the ctor, look at all overloads and end up at the places, where we pass in a function callback for the scheduler. Vs. using a proper interface, or even a virtual method on the processor to allow implement a different scheduler (not my favorite), both would allow to find me all implementations with a single query.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            var outcome = await TryProcessAsync(processor, function);
+            return new ProcessingOutcome<TResult>(await outcome.Result, outcome.SubmissionTime, outcome.WaitDuration, outcome.RunDuration);
+        }
+
+        public async Task<ProcessingOutcome<Task<TResult>>> TryProcessAsync<TResult>(T processor, Func<Task<TResult>> function)
+        {
+            if (function == null) throw new ArgumentNullException(nameof(function));
+
+            var submissionTime = DateTimeOffset.Now;
```
p-schuler, line 83: UtcNow?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
```diff
@@ -0,0 +1,129 @@
...
+                    }
+                    else
+                    {
+                        Processing?.Invoke(this, processor);
+                        var startTime = DateTimeOffset.Now;
+                        var task = await Task.WhenAny(function());
+                        var endTime = DateTimeOffset.Now;
+                        var outcome = new ProcessingTaskOutcome<TResult>(task, submissionTime, startTime - submissionTime, endTime - startTime);
+                        Processed?.Invoke(this, (processor, outcome));
+                        return outcome;
```
p-schuler, line 119: This is used for logging only and hence I think this should all be UTC.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
```diff
@@ -1,162 +1,399 @@
...
 
-        public void Register(LoRaDevice loRaDevice, ILoRaDeviceClient loraDeviceClient)
-        {
-            if (loRaDevice is null) throw new ArgumentNullException(nameof(loRaDevice));
+            private async Task<T> InvokeExclusivelyAsync<T>(bool doesNotRequireOpenConnection,
+                                                            Func<ILoRaDeviceClient, Task<T>> processor,
+                                                            [CallerMemberName] string? callerName = null)
+            {
+                _ = Interlocked.Increment(ref this.pid);
+                return await this.exclusiveProcessor.ProcessAsync(new Process(this.pid, callerName!), async () =>
```
p-schuler, line 208: where do we need that pid? couldn't figure it out.

p-schuler: I saw we print the "id" in the callbacks. Not sure what we will use this for though?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
```diff
@@ -1,162 +1,399 @@
...
+                {
+                    if (!doesNotRequireOpenConnection)
```
p-schuler, line 210: nit: if you change this to "requiresOpenConnection" it becomes more readable.

p-schuler: is this ever needed? what operation would not require an open connection?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
```diff
@@ -1,162 +1,399 @@
...
+
+            private enum DisconnectionResult { Disconnected, Deferred, NotDisconnected }
 
-            if (this.managedConnections.TryRemove(GetConnectionCacheKey(loRaDevice.DevEUI), out var removedItem))
+            private Task<DisconnectionResult> DisconnectAsync(bool deferred) =>
+                InvokeExclusivelyAsync(doesNotRequireOpenConnection: true, async client =>
+                {
+                    DisconnectionResult result;
+
+                    if (Interlocked.Add(ref this.activities, 0) > 0)
```
p-schuler, line 264: not sure what we are trying to do here. The BeginDeviceClientConnectionActivity can be called any time. Why not just compare to 0 here instead? If you are racing, we have the race anyway and even if it is 0 now, it can be 1 on the next line, causing the disconnect to happen. 

Operations will eventually re-open the connection any they are synchronized, so not a big deal but the validation with 0 seems not to require an Interlocked.Add

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
```diff
@@ -1,162 +1,399 @@
...
-                        return managedConnection;
-                    });
+        public IAsyncDisposable BeginDeviceClientConnectionActivity(LoRaDevice loRaDevice)
+        {
+            if (loRaDevice == null) throw new ArgumentNullException(nameof(loRaDevice));
+            return this.clientByDevEui[loRaDevice.DevEUI].BeginDeviceClientConnectionActivity();
         }
 
-        private void OnScheduledDisconnect(object key, object value, EvictionReason reason, object state)
+        private sealed class SynchronizedLoRaDeviceClient : ILoRaDeviceClient, ILoRaDeviceClientSynchronizedOperationEventSource
```
p-schuler, line 143: no unit tests for this?

atifaziz: It's a private class of `LoRaDeviceClientConnectionManager` so it cannot be publicly instantiated. That said, it is unit-tested, albeit indirectly, through the connection manager that returns this client in `GetClient` and the public `ILoRaDeviceClient` members that it implements. See `LoRaDeviceClientConnectionManagerTests`.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
atifaziz: > is this ever needed? what operation would not require an open connection?

`DisconnectAsync`

atifaziz: > nit: if you change this to "requiresOpenConnection" it becomes more readable.

I know what you mean, but a recommended practice that I am following here is that an optional `Boolean` value should default to `false` otherwise it gets very hard to remember (without consulting the source) what's the default behaviour. It does indeed sometimes mean double-negation from a naming/usage perspective.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
atifaziz: For now, this is purely for diagnostics purpose to understand the sequence in which operations were invoked. If you think it's not providing any value and just adding noise to logs then I can remove it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
atifaziz: A `DateTimeOffset` always has an offset (even when printed) so there's no ambiguity (like with `DateTime`) whether it's local time or UTC; the former can be converted to the latter any time desired. The decision to use UTC or not can then be done at the time of logging/printing, but I find it's best to capture the value with full fidelity when initialized.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
atifaziz: [See my related comment](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1517#discussion_r811648861).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
atifaziz: This is only dealing with _deferred_ disconnects (`deferred: true`). It means, someone issued a `DisconnectAsync` while activities were taking place. To avoid interference, the actual disconnection is deferred until all activities are done. I agree that this can mislead to read that a disconnection is done, but it's predicated. I could rename the parameter to `ifDeferred` or `whenDeferred` if it helps?

atifaziz: > The owning LNS should not drop the connection.

The ownership problem is not being solved directly in this PR. It's preparing the terrain to make that easier to fit in a more deterministic setup.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
atifaziz: You are right, it doesn't. I feel this is general/generic _infrastructure code_ that supports `SynchronizedLoRaDeviceClient`. It could have existed directly as part of the client code but was extracted into a separate class since it's _technically_ unrelated and could be explored and reasoned about separately. Since `SynchronizedLoRaDeviceClient` uses this type, it is indirectly getting tested. I wanted to avoid duplicating the tests. According to VS code coverage, 90% of the class is exercised:

| Code                                                    | Covered (% Blocks) |
|---------------------------------------------------------|:------------------:|
| `ExclusiveProcessor.<>c<T>`                             |       100.00%      |
| `ExclusiveProcessor.<ProcessAsync>d__22<T, TResult>`    |       100.00%      |
| `ExclusiveProcessor.<TryProcessAsync>d__23<T, TResult>` |       78.33%       |
| `ExclusiveProcessor.ProcessingOutcome<T, TResult>`      |       66.67%       |
| `ExclusiveProcessor.ProcessingTaskOutcome<T, TResult>`  |       100.00%      |
| `ExclusiveProcessor<T>`                                 |       100.00%      |

Some of the untested parts are due to argument validation and never hitting the case of _interrupted processes_ since we don't take advantage of custom scheduling where one process could be given a higher priority over another; the default behaviour being FIFO:

```c#
public ExclusiveProcessor() : this(ps => ps[0]) { }
```

If you feel strongly about covering this then I can add the unit tests or we can defer that decision to when we see a drastic fall in the coverage of this class (at which point I would even question how much of it is really being used).

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
atifaziz: Refactored the scheduler to an interface with 8a5c52594de7590c4d76ce3c58af6a71999b0eef. Please resolve if the new version is what you had in mind.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
p-schuler: As discussed. I would have chosen a different take on this and preferred more isolated tests around some of the implementations we have. I see that you did do a lot of tests with the ConnectionManager and I trust you have the cases covered.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
p-schuler: After talking to Atif I understand: This was added to support a longer operation with a disconnect in between, like the join case (disconnecting in the factory and then defer that operation until we are done saving the twin). 

Smart! <3

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
p-schuler: As discussed - we should add unit tests for this. Also in light of the re-usability and it is a public class. Resolve this when you opened the item to track it @atifaziz

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
p-schuler: As discussed. All default formatting should be UTC. Using UTC now just ensures the default formatting is correct and no one has to take action to ensure we get the right timestamps in the logs.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
p-schuler: I think you may also need it to ensure uniqueness of the operation? But I might be wrong. It's ok to have that sequence number, however, you may want to rename it so it is obvious, that it is a sequence number for the operations?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
p-schuler: As discussed. I leave it up to you what you want to do with this. I feel it's adding a false sense of isolation. The next line the state can be changed no matter what you do on 264. The decision is always (potentially) outdated and different from the current state.

p-schuler (approved):
Some open items in the comments, otherwise LGTM

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
atifaziz: Done with 09b7879898ad752a143fc8beb5f0a6a63465d582.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
atifaziz: Renamed to `operationSequenceNumber` with 5a3e07d0391031bfd8bf0e1a9f782faa78f675ca.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/ExclusiveProcessor.cs
atifaziz: [Resolved](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1517#discussion_r811874561).

danigian (approved):
Looks great, thanks Atif!

Issue #1490, opened 2022-02-14
[Story] Use standalone discovery endpoint in Quickstart template


bastbu:
## Description

Depends on #1489. 

Automatically deploy the discovery service when using the Quickstart template.

## Acceptance criteria

- [ ] Standalone discovery endpoint is deployed when using Quickstart template
- [ ] Docs of Quickstart are up to date

PR #1520, opened 2022-02-21
Opt-in deployment of discovery service in Quickstart

bastbu:
# PR for issue #1490

## What is being addressed

We add the possibility to deploy the discovery service using the Quickstart template. This deploys a basic, non-production ready discovery service as an App Service. Per default, the standalone discovery service is not deployed when using the Quickstart.

## How is this addressed

We deploy a dedicated App Service Plan (Windows, B1) that hosts the discovery service if the user opts in when using the Quickstart template. The Quickstart template:

- Deploys an App Service Plan
- Deploys an App Service (which runs the discovery service from a package. This package does not yet exist and will only be created with the next release. This is tracked as part of #1521)
- Assigns a role to the App Service managed identity that allows it to query the IoT Hub twins.
- Configures the sample devices to be part of the same "network"

Limitations:

- The user still needs to update the `hostAddress` in the LNS module of the sample edge device to make sure that the discovery service points to the correct DNS/IP of the LNS. This will be documented as part of #1489
- The App Service allows both HTTP and HTTPS per default. This will be documented as well.
- Application Insights integration is made via the `APPINSIGHTS_INSTRUMENTATIONKEY`. This still leaves the user to enable Application Insights on the App Service, which looks incomplete but [does not make a difference when using .NET](https://docs.microsoft.com/en-us/azure/azure-monitor/app/azure-web-apps#enable-application-insights)


Template/azuredeploy.json
```diff
@@ -86,12 +93,14 @@
         "functionAppName": "[concat(parameters('uniqueSolutionPrefix'), 'function')]",
         "gitUsername": "Azure",
         "functionZipBinary": "https://github.com/Azure/iotedge-lorawan-starterkit/releases/download/v2.0.0/function-2.0.0.zip",
+        "discoveryServiceZipBinary": "TODO:replacewithnextrelease",
```
danigian, line 96: Assuming next release is 2.1.0, you could just use 'https://github.com/Azure/iotedge-lorawan-starterkit/releases/download/v2.1.0/discoveryservice-2.1.0.zip'

Otherwise, if we are not yet sure, let's maybe create a checklist of items that need to be changed for next release (inside a GitHub issue item tracking 2.1.0 release itself)

Template/discoveryService.json
```diff
@@ -0,0 +1,106 @@
...
+        },
+        "iotHubName": {
+            "type": "string"
+        },
+        "appInsightsInstrumentationKey": {
+            "type": "string"
+        },
+        "roleNameGuid": {
+            "type": "string",
+            "defaultValue": "[newGuid()]",
```
danigian, line 34:
```suggestion
            "defaultValue": "[guid(resourceGroup().id, "twincontributor")]",
```

Someone might decide to reuse this template and if a new guid is generated, a "Conflict: RoleAssignmentExists" will be thrown.
Please consider usage of [guid function](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string#guid) for scenarios like this one.

Template/azuredeploy.json
bastbu: Agreed, I had already created an item for this and linked it in the PR description: #1521. I will change the title to include v2.1.0 to make it easier to find.

Template/discoveryService.json
bastbu: Nice catch, definitely better that way!

danigian (approved):

PR #1523, opened 2022-02-21
Improve discovery misconfiguration error messages

bastbu:
# PR for issue #1489

## What is being addressed

With the initial version of the standalone discovery endpoint the error messages in case of misconfigured host addresses for LNS were cryptic. With this PR we improve the error messages that are logged/thrown/responded in the presence of misconfigured LNS.

We also fix a bug where URI paths are ignored in the host address specification in configuration.

## How is this addressed

The general approach is to:

- throw if there are no or only misconfigured LNS and provide a helpful error message in that case
- continue and only use the valid LNS configurations in case there is a mix of valid and invalid LNS configurations. The LNS with a misconfigured `hostAddress` are logged and ignored.

All changes rely on the fact that the IoT Hub query always returns a `deviceId` for each element in a query result.

### DIFF
--- a/LoRaEngine/modules/LoRaWan.NetworkServerDiscovery/TagBasedLnsDiscovery.cs
+++ b/LoRaEngine/modules/LoRaWan.NetworkServerDiscovery/TagBasedLnsDiscovery.cs
@@ -27,13 +27,12 @@ namespace LoRaWan.NetworkServerDiscovery
         private static readonly IJsonReader<LnsHostAddressParseResult> HostAddressReader =
             JsonReader.Object(JsonReader.Property("hostAddress",
                                                   from s in JsonReader.String()
-                                                  select Uri.TryCreate(s, default, out var uri) ? uri : null,
+                                                  select Uri.TryCreate(s, UriKind.Absolute, out var uri)
+                                                      && uri.Scheme is "ws" or "wss"
+                                                       ? uri : null,
                                                   (true, null)),
                               JsonReader.Property("deviceId", JsonReader.String()),
-                              (hostAddress, deviceId) => new LnsHostAddressParseResult(hostAddress is { } someHostAddress
-                                                                                                       && someHostAddress.IsAbsoluteUri
-                                                                                                       && someHostAddress.Scheme is "ws" or "wss" ? someHostAddress : null,
-                                                                                       deviceId));
+                              (hostAddress, deviceId) => new LnsHostAddressParseResult(hostAddress, deviceId));
 
         private readonly ILogger<TagBasedLnsDiscovery> logger;
         private readonly IMemoryCache memoryCache;
--- a/LoRaEngine/modules/LoRaWan.NetworkServerDiscovery/TagBasedLnsDiscovery.cs
+++ b/LoRaEngine/modules/LoRaWan.NetworkServerDiscovery/TagBasedLnsDiscovery.cs
@@ -24,11 +24,16 @@ namespace LoRaWan.NetworkServerDiscovery
         private const string NetworkByStationCacheKey = "NetworkByStation";
 
         private static readonly TimeSpan CacheItemExpiration = TimeSpan.FromHours(6);
-        private static readonly IJsonReader<Uri?> HostAddressReader =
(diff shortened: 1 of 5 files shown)


danigian (approved):

LoRaEngine/modules/LoRaWan.NetworkServerDiscovery/TagBasedLnsDiscovery.cs
```diff
@@ -24,11 +24,16 @@ public sealed class TagBasedLnsDiscovery : ILnsDiscovery, IDisposable
...
             JsonReader.Object(JsonReader.Property("hostAddress",
                                                   from s in JsonReader.String()
-                                                  select string.IsNullOrEmpty(s) ? null : new Uri(s),
-                                                  (true, null)));
+                                                  select Uri.TryCreate(s, default, out var uri) ? uri : null,
+                                                  (true, null)),
+                              JsonReader.Property("deviceId", JsonReader.String()),
+                              (hostAddress, deviceId) => new LnsHostAddressParseResult(hostAddress is { } someHostAddress
+                                                                                                       && someHostAddress.IsAbsoluteUri
+                                                                                                       && someHostAddress.Scheme is "ws" or "wss" ? someHostAddress : null,
```
atifaziz, line 35: Can't the host address conditions be moved earlier when parsing it?

```suggestion
                                                  select Uri.TryCreate(s, UriKind.Absolute, out var uri)
                                                      && uri.Scheme is "ws" or "wss"
                                                       ? uri : null,
                                                  (true, null)),
                              JsonReader.Property("deviceId", JsonReader.String()),
                              (hostAddress, deviceId) => new LnsHostAddressParseResult(hostAddress, deviceId));
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/NetworkServerDiscovery/DiscoveryService.cs
```diff
@@ -69,7 +70,13 @@ public async Task HandleDiscoveryRequestAsync(HttpContext httpContext, Cancellat
                                               Id6.FormatOptions.FixedWidth);
 
                         var lnsUri = await this.lnsDiscovery.ResolveLnsAsync(stationEui, cancellationToken);
-                        var url = new Uri($"{lnsUri.Scheme}://{lnsUri.Host}:{lnsUri.Port}{DataEndpointPath}/{stationEui}");
+
+                        // Ensure resilience against duplicate specification of `router-data` and make sure that LNS host address ends with slash
+                        // to make sure that URI composes as expected.
+                        var lnsUriSanitized = Regex.Replace(lnsUri.AbsoluteUri, @"/router-data/?", string.Empty, RegexOptions.IgnoreCase);
```
atifaziz, line 76: Are you sure about `RegexOptions.IgnoreCase`? URL paths are _case-sensitive_.

bastbu: I agree that [the specification](https://www.rfc-editor.org/rfc/rfc7230) says that URL paths are case-sensitive. In practice, some web servers apply case-insensitive path matching (i.e. IIS). What made me choose case-insensitive over case-sensitive matching in this case is Postel's law - the intention of the configuration is clear, we should accept it and avoid adding the `router-data` path twice.

atifaziz: > In practice, some web servers apply case-insensitive path matching (i.e. IIS).

I doubted that IIS doesn't respect the rule. I think the misconception might be coming from the fact that the Windows file system uses case-insensitive file names so when serving static files, a URL identifying a file will serve the file if the path is simply different in case. That said, I agree that we can go Postel's way. I'd suggest to add `RegexOptions.CultureInvariant` as well though.

bastbu: Makes sense! I added this in e639f929e5d1c5a90c365d52bdc2fb5d6956b474

Issue #1532, opened 2022-02-22
[Story] Unify device twin creation in test code


bastbu:
## Description

We have many tests that rely on the creation of a `Microsoft.Azure.Devices.Shared.Twin` with some desired/reported properties (e.g. `Join_With_Subsequent_Unconfirmed_And_Confirmed_Messages`). For many of the possible properties of a twin we have types, e.g. `LoRaWan.DevAddr` for the `"DevAddr"`, `LoRaWan.AppSessionKey` for the `"AppSKey"` property, etc.

Instead of duplicating the code to create these device twins on the call site wherever we need it (and repeating the code for string conversions etc), we should use a central mechanism to create twins for our test code.

## Acceptance criteria

- [ ] Code to create `Microsoft.Azure.Devices.Shared.Twin` is shared between tests and uses the types for the properties when possible.

PR #1539, opened 2022-02-23
Unify test twin creation

bastbu:
# PR for issue #1532

## What is being addressed

In several tests, if we must set up device twins as part of the test setup, we set these twins up in a manner that involves some duplication:

- we duplicate how we represent certain types, e.g. a `DevEui`, `NetworkSessionKey`, etc, in a device twin
- we duplicate how the device twin properties are named (albeit relying on constants)
- we duplicate the twin structures for OTAA/ABP devices

With this PR we introduce a mechanism for test device twin creation. The new mechanism helps us:

- using strong types for device twin properties
- compose similar twin properties by supproting nondestructive mutation
- encapsulating how certain properties are represented in a LoRa device twin

In this PR we introduce the mechanism for only a part of all twins used in the test code, to get early feedback. The rest of the occurrences of the legacy twin creation methods will be replaced in a separate PR.

PR #1548, opened 2022-02-24
Unify test twin setup in `LoRaDeviceTest`

bastbu:
# PR for issue #1532

## What is being addressed

With this PR we use the common mechanism to set up test twins also for the tests in `LoRaDeviceTest`. We also take the chance to improve the test device randomness, since previously there were collisions between different keys, and it was not always clear whether reported or desired properties were expected as part of test assertions.

## How is this addressed

For each test case in `LoRaDeviceTest` we distinguish whether the test asserts on the property *contract*, or whether it requires a twin but has an assertion that is unrelated to the contract. If it asserts the data contract, we keep the existing strategy of setting up the device twins "manually". If it only asserts part of the contract, we start with a common way of creating device twins for OTAA/ABP devices and manually amend the properties on which the test asserts.

Furthermore, we increase the case coverage that we get from our tests by not relying on the `TwinProperty` when creating device twins in `LoRaDeviceTwin`. By hardcoding the `string` property keys, we know that unit tests will break should we change the name of some device twin properties (breaking change detector).

We fix the wrong representation of a `LoRaRegionType` in a twin created via `LoRaDeviceTwin` - even though it is an enum, it's an exception that is serialized as string instead of a number.

### DIFF
--- a/Tests/Common/LoRaDeviceTwin.cs
+++ b/Tests/Common/LoRaDeviceTwin.cs
@@ -80,7 +80,7 @@ namespace LoRaWan.Tests.Common
         public bool? AbpRelaxMode { get; init; }
         public uint? FCntUpStart { get; init; }
         public uint? FCntDownStart { get; init; }
-        public int? FCntResetCounter { get; init; }
+        public uint? FCntResetCounter { get; init; }
         public int? Rx1DROffset { get; init; }
         public DataRateIndex? Rx2DataRate { get; init; }
         public ReceiveWindowNumber? PreferredWindow { get; init; }
@@ -90,30 +90,30 @@ namespace LoRaWan.Tests.Common
 
         public override IEnumerator<KeyValuePair<string, object?>> GetEnumerator()
         {
-            yield return KeyValuePair.Create(TwinProperty.DevEUI, (object?)DevEui);
-            yield return KeyValuePair.Create(TwinProperty.DevAddr, (object?)DevAddr);
-            yield return KeyValuePair.Create(TwinProperty.AppEui, (object?)JoinEui);
-            yield return KeyValuePair.Create(TwinProperty.AppKey, (object?)AppKey);
-            yield return KeyValuePair.Create(TwinProperty.AppSKey, (object?)AppSessionKey);
-            yield return KeyValuePair.Create(TwinProperty.NwkSKey, (object?)NetworkSessionKey);
-            yield return KeyValuePair.Create(TwinProperty.GatewayID, (object?)GatewayId);
-            yield return KeyValuePair.Create(TwinProperty.SensorDecoder, (object?)SensorDecoder);
-            yield return KeyValuePair.Create(TwinProperty.Supports32BitFCnt, (object?)Supports32BitFCnt);
-            yield return KeyValuePair.Create(TwinProperty.ABPRelaxMode, (object?)AbpRelaxMode);
-            yield return KeyValuePair.Create(TwinProperty.FCntUpStart, (object?)FCntUpStart);
-            yield return KeyValuePair.Create(TwinProperty.FCntDownStart, (object?)FCntDownStart);
-            yield return KeyValuePair.Create(TwinProperty.FCntResetCounter, (object?)FCntResetCounter);
-            yield return KeyValuePair.Create(TwinProperty.RX1DROffset, (object?)Rx1DROffset);
-            yield return KeyValuePair.Create(TwinProperty.RX2DataRate, (object?)Rx2DataRate);
-            yield return KeyValuePair.Create(TwinProperty.PreferredWindow, PreferredWindow switch
+            yield return KeyValuePair.Create("DevEUI", (object?)DevEui);
+            yield return KeyValuePair.Create("DevAddr", (object?)DevAddr);
+            yield return KeyValuePair.Create("AppEUI", (object?)JoinEui);
+            yield return KeyValuePair.Create("AppKey", (object?)AppKey);
+            yield return KeyValuePair.Create("AppSKey", (object?)AppSessionKey);
+            yield return KeyValuePair.Create("NwkSKey", (object?)NetworkSessionKey);
+            yield return KeyValuePair.Create("GatewayID", (object?)GatewayId);
+            yield return KeyValuePair.Create("SensorDecoder", (object?)SensorDecoder);
+            yield return KeyValuePair.Create("Supports32BitFCnt", (object?)Supports32BitFCnt);
+            yield return KeyValuePair.Create("ABPRelaxMode", (object?)AbpRelaxMode);
+            yield return KeyValuePair.Create("FCntUpStart", (object?)FCntUpStart);
+            yield return KeyValuePair.Create("FCntDownStart", (object?)FCntDownStart);
+            yield return KeyValuePair.Create("FCntResetCounter", (object?)FCntResetCounter);
+            yield return KeyValuePair.Create("RX1DROffset", (object?)Rx1DROffset);
+            yield return KeyValuePair.Create("RX2DataRate", (object?)Rx2DataRate);
+            yield return KeyValuePair.Create("PreferredWindow", PreferredWindow switch
             {
                 ReceiveWindowNumber.ReceiveWindow1 => 1,
                 ReceiveWindowNumber.ReceiveWindow2 => 2,
                 _ => (object?)null
             });
-            yield return KeyValuePair.Create(TwinProperty.RXDelay, (object?)RxDelay);
-            yield return KeyValuePair.Create(TwinProperty.ClassType, (object?)ClassType);
-            yield return KeyValuePair.Create(TwinProperty.KeepAliveTimeout, (object?)KeepAliveTimeout?.TotalSeconds);
+            yield return KeyValuePair.Create("RXDelay", (object?)RxDelay);
+            yield return KeyValuePair.Create("ClassType", (object?)ClassType);
+            yield return KeyValuePair.Create("KeepAliveTimeout", (object?)KeepAliveTimeout?.TotalSeconds);
         }
     }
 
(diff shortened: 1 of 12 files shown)


Tests/Common/LoRaDeviceTwin.cs
```diff
@@ -40,6 +39,9 @@ public static Twin Create(LoRaDesiredTwinProperties? desiredProperties = null, L
                 {
                     case null:
                         break;
+                    case LoRaRegionType region:
```
atifaziz, line 42: Since the code in this branch is the same as in the last `var obj` case, you could remove this and instead change the `Enum` case to the following instead?

```c#
                    case Enum and not LoRaRegionType:
```

Although I suspect it will get to the `IConvertible` case before `var obj`, which is harmless.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
```diff
@@ -27,14 +28,46 @@ namespace LoRaWan.Tests.Unit.NetworkServer
     /// </summary>
     public class LoRaDeviceTest
     {
+        private static readonly NetworkServerConfiguration Configuration = new NetworkServerConfiguration { GatewayID = "test-gateway" };
+        private static readonly LoRaDesiredTwinProperties OtaaDesiredTwinProperties = new LoRaDesiredTwinProperties
```
atifaziz, line 32: Perhaps we can skip saying `LoRaDesiredTwinProperties` twice since it doesn't add much value, here and elsewhere below?

```suggestion
        private static readonly LoRaDesiredTwinProperties OtaaDesiredTwinProperties = new
```

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
```diff
@@ -106,33 +139,28 @@ public async Task After_Saving_Frame_Counter_Changes_Should_Not_Have_Pending_Cha
         [Fact]
         public async Task When_Initialized_New_OTAA_Device_Should_Have_All_Properties()
         {
-            var appKey = TestKeys.CreateAppKey(0xABC0200000000000, 0x09);
-            var joinEui = new JoinEui(0xABC0200000000009);
-
-            var twin = TestUtils.CreateTwin(
-                desired: new Dictionary<string, object>
+            const string gateway = "mygateway";
```
atifaziz, line 142: Do we need a different gateway identifier here? Why not reuse the one from `OtaaDesiredTwinProperties.GatewayId`?

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
```diff
@@ -1057,5 +867,27 @@ public void BeginDeviceClientConnectionActivity_Delegates_To_Connection_Manager_
             connectionManagerMock.Verify(x => x.BeginDeviceClientConnectionActivity(target),
                                          Times.Exactly(timeoutSeconds > 0 ? 1 : 0));
         }
+
+        private static Twin CreateTwin(Dictionary<string, object> desired = null, Dictionary<string, object> reported = null)
```
atifaziz, line 871: Wouldn't it make sense to use the `LoRaTwinProperties` records here too?

Tests/Common/LoRaDeviceTwin.cs
atifaziz: BTW, I removed this case of `LoRaRegionType` entirely and nothing seems to be breaking?

atifaziz: You can collapse the cases to just one pattern:

```diff
diff --git a/Tests/Common/LoRaDeviceTwin.cs b/Tests/Common/LoRaDeviceTwin.cs
index b7853d97..2d692638 100644
--- a/Tests/Common/LoRaDeviceTwin.cs
+++ b/Tests/Common/LoRaDeviceTwin.cs
@@ -39,13 +39,7 @@ namespace LoRaWan.Tests.Common
                 {
                     case null:
                         break;
-                    case LoRaRegionType region:
-                        target[key] = region.ToString();
-                        break;
-                    case uint:
-                    case int:
-                    case Enum:
-                    case bool:
+                    case uint or int or bool or (Enum and not LoRaRegionType):
                         target[key] = value;
                         break;
                     case IConvertible convertible:
```

atifaziz (commented):
This is a really good start but I feel there are more opportunities for the great stuff you added in PR #1539.

Tests/Common/LoRaDeviceTwin.cs
```diff
@@ -80,40 +82,46 @@ public sealed record LoRaDesiredTwinProperties : LoRaTwinProperties
         public bool? AbpRelaxMode { get; init; }
         public uint? FCntUpStart { get; init; }
         public uint? FCntDownStart { get; init; }
-        public int? FCntResetCounter { get; init; }
+        public uint? FCntResetCounter { get; init; }
         public int? Rx1DROffset { get; init; }
         public DataRateIndex? Rx2DataRate { get; init; }
         public ReceiveWindowNumber? PreferredWindow { get; init; }
         public int? RxDelay { get; init; }
         public char? ClassType { get; init; }
```
atifaziz, line 90: Shouldn't this use `LoRaDeviceClassType`?

Tests/Common/LoRaDeviceTwin.cs
bastbu: I'll apply your suggestion. It's true that nothing breaks, this is just because we can parse both number and string representations into Enums. However, the real code serializes the region as a string, so to stay close to the real behavior we should serialize it as string as well.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
bastbu: Then the assertions on the which is the owning gateway would break. I kept the assertions from the old tests, I will give it another pass and see if I can simplify further.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
bastbu: I don't think so, but maybe we can do better. The `LoRaTwinProperties` simplify the setup of *correct* test twins for us. Some tests in this class have assertions on the data contracts - sometimes by using different representation of properties (e.g. an `int` as a number and as a string once, while expecting that our code can handle both). If we were to use the `LoRaTwinProperties` in all cases, we would not get the same coverage. This is why I kept a `private` helper method to create a twin from these dictionaries containing arbitrary values.

Tests/Common/LoRaDeviceTwin.cs
atifaziz: I figured that was reasoning, which is why I left in the `Enum and not LoRaRegionType` pattern.

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
atifaziz: I see. I looked at the only 3 usages and nothing struck me as testing the various representation, which is why I asked the question. Perhaps I missed a case or two?

Tests/Common/LoRaDeviceTwin.cs
bastbu: Addressed this with 080e6b242d4a21d942b3c6052476229f6246db6c

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
bastbu: Applied for this occurrence and the other `static readonly` occurrences in the same test as part of 0bc8919c90a7ce0c10880fea37c72d4f354b5b4d

Tests/Common/LoRaDeviceTwin.cs
bastbu: Valid point, I propagated the usage of the type to its callers as well (`SimulatedDevice`, `TestDeviceInfo` etc) as part of c6680ad02b397af5e045c9b9d57ed44c56b64d57. Much better now!

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
bastbu: You are right, I mixed things up with other occurrences where I am setting a property directly on the twin. Addressed with c03d6d5b38588ca4dc0b6d8f7da194698e229285

Tests/Unit/NetworkServer/LoRaDeviceTest.cs
bastbu: I replaced the existing usages with a separate test that explicitly tests this behavior in fe45f82239fef39ab3b7fa818a73da9e8e66b3b8

Tests/Common/TestUtils.cs
```diff
@@ -37,7 +36,7 @@ public static class TestUtils
                 NwkSKey = simulatedDevice.LoRaDevice.NwkSKey,
                 GatewayID = simulatedDevice.LoRaDevice.GatewayID,
                 IsOurDevice = true,
-                ClassType = (simulatedDevice.ClassType is 'C' or 'c') ? LoRaDeviceClassType.C : LoRaDeviceClassType.A,
+                ClassType = simulatedDevice.ClassType == LoRaDeviceClassType.C ? LoRaDeviceClassType.C : LoRaDeviceClassType.A,
```
atifaziz, line 39: Can't this be just the following?

```suggestion
                ClassType = simulatedDevice.ClassType,
```

atifaziz (commented):
Ship it! :shipit:

Tests/Common/TestUtils.cs
bastbu: I think it could, since we don't use class B in any case. I just wanted to do the mechanical change here, though. In any case, I think it's probably worth getting rid of the complexity, so I'll apply the suggestion :)

Issue #1496, opened 2022-02-15
Implement basic session affinity handling


p-schuler:
This does track the implementation of basic session affinity for LNS in multi gateway mode.

The ADR 009 describes the target design. 

Acceptance criteria

- [x] Active/Passive tracking on LoRaDevice after function bundler result
- [x] Basic connection management for passive vs. active devices
- [x] Delay on passive device requests
- [x] Unit and Integration tests

Out of scope
- all edge cases regarding connection races
- handling of all passive scenarios (only the main data message scenario is included)
- handling of roaming scenario

PR #1549, opened 2022-02-24
Session affinity handling

danigian:
# PR for issue #1496 

## What is being addressed

We noticed that a connection ping-pong is negatively affecting performances in case multiple IoT Edge Hubs are handling the same leaf device for long periods of time.

The idea behind this PR is to implement session affinity between one device and a IoT Edge Gateway in order to mitigate the connection issues.

Following major changes are being implemented with this PR:
- [x] LoRaDevice class is now keeping track of whether current IoT Edge Gateway is "owning" or not the connection
- [x] The ownership is determined after the "function bundler" execution and purely based on the result of the race towards the shared Redis instance
- [x] LNS now proactively drops any connection for a LoRaDevice which is not being owned by current IoT Edge Gateway.
- [x] LNS is proactively dropping any connection after *all* Join requests
- [x] When a LoRaDevice is not "owned" by current gateway, data requests processing will be delayed in order to guarantee that the IoT Edge Gateway owning current device could reach the function. Delaying and not aborting requests is guaranteeing that "roaming" devices are still handled in case the previous owning gateway is out of reach or currently unavailable for processing.

For more details you can refer to [ADR 010](https://azure.github.io/iotedge-lorawan-starterkit/dev/adr/010_lns_affinity/)

## How is this addressed

- [x] DefaultLoRaDataRequestHandler, LoRaDevice, ILoRaDeviceClient, LoRaDeviceCache, LoRaDeviceRegistry, LoRaDeviceClientConnectionManager and MessageDispatcher are now inheriting from `IAsyncDisposable`. Numerous changes in this PR are related to properly awaiting such disposable classes
- [x] DefaultLoRaDataRequestHandler has been changed to:
    - [x] postpone the `BeginDeviceClientConnectionActivity` call after the Function Bundler execution
    - [x] eventually update "ownership" of LoRaDevice after the Function Bundler execution
    - [x] delay processing (400ms) in case current LNS is not "owning" the LoRaDevice
    - [x] saving twin changes only if current LNS is "owning" the LoRaDevice
- [x] JoinRequestMessageHandler.cs has been changed to forcefully close the connection after any join request
- [x] LoRaDeviceClient.cs has been changed to keep track of active connections (to be incremented when a CreateDeviceClient is invoked and to be decremented when a Disconnect is invoked)
- [x] LoRaDeviceFactory.cs is now closing the client explicitly if we were unable to register or initialize the LoRaDevice
- [x] LoRaDataRequestHandlerConnectionAffinityIntegrationTests has been added for testing the processing delay
- [x] JoinRequestMessageHandlerIntegrationTests is now ensuring that the connection is forcefully closed after a join
- [x] Unit and integration tests were fixed for IAsyncDisposable related changes


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
```diff
@@ -93,7 +93,7 @@ public void Register(LoRaDevice loRaDevice, ILoRaDeviceClient loraDeviceClient)
                         try
                         {
                             using var scope = logger.BeginDeviceScope(client.DevEui);
-                            await client.DisconnectAsync();
+                            await client.DisconnectAsync(CancellationToken.None);
```
danigian, line 96: @p-schuler should we force the disconnection here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
```diff
@@ -73,13 +74,32 @@ private async Task<LoRaDevice> RegisterCoreAsync(IoTHubDeviceInfo deviceInfo, Ca
...
+                // release the loradevice client explicitly. If we were unable to register, or there was already
+                // a connection registered, we will leak this client.
+                await loRaDeviceClient.DisposeAsync();
+
+                if (this.logger.IsEnabled(LogLevel.Debug))
+                {
+                    try
+                    {
+                        // if the created client is registered, release it
+                        if (loRaDeviceClient != ((IIdentityProvider<ILoRaDeviceClient>)this.connectionManager.GetClient(loRaDevice)).Identity)
```
danigian, line 93: @p-schuler should we use ReferenceEquals here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
p-schuler: Yes - I think we should - the point here is that after a certain amount of time we want to disconnect. Since it is a sliding expiration, we will only disconnect if we haven't had any messages from that device for that amount of time. Of course the memory cache makes it non-deterministic, but once we hit that point, even if we own it, we should close it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
p-schuler: That's what will happen (a ref compare). nit, but we can make it explicit so ppl understand.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
danigian: I just realized that this is the synchronized client, so it should be fine as it is.
Agree?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
danigian: Yes, it was a nit. Just asking if it was fine for you, since you wrote that part. Done.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs
p-schuler: yeap - the disconnectAsync on the LoRaDeviceClient(s) is always executed. Only the implementation in LoRaDevice is taking the ownership into consideration. all good.

Tests/Unit/NetworkServer/LoRaDeviceFactoryTest.cs
```diff
@@ -136,6 +141,12 @@ protected override LoRaDevice CreateDevice(IoTHubDeviceInfo deviceInfo)
                 LastDeviceMock = device;
                 return device.Object;
             }
+
+            public override ILoRaDeviceClient CreateDeviceClient(string deviceId, string primaryKey)
+            {
+                LastDeviceClientMock = new Mock<ILoRaDeviceClient>();
+                return LastDeviceClientMock.Object;
+            }
```
p-schuler, line 149:
```suggestion
            public override ILoRaDeviceClient CreateDeviceClient(string deviceId, string primaryKey) => Mock.Of<ILoRaDeviceClient>();
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
```diff
@@ -925,18 +929,25 @@ async Task CoreAsync()
...
 
-        public void Dispose()
+        protected virtual async ValueTask DisposeAsyncCore()
+        {
+            if (this.connectionManager is { } someConnectionManager)
+                await someConnectionManager.ReleaseAsync(this);
+            this.syncSave.Dispose();
+        }
+
+        public async ValueTask DisposeAsync()
```
atifaziz, line 948: Unless I missed it, there doesn't seem to be a test that checks the release on connection manager when the device is disposed.

Tests/Unit/NetworkServer/LoRaDeviceFactoryTest.cs
danigian: I need to assign the Mock to the LastDeviceClientMock in order to do some verifications on it.
I could condense it to the following, but I am not sure if it's still readable enough 
```suggestion
                        public override ILoRaDeviceClient CreateDeviceClient(string deviceId, string primaryKey) => (LastDeviceClientMock = new Mock<ILoRaDeviceClient>()).Object;
```

p-schuler: sorry, overread the fact that it is assigned to a member. no worries then.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDevice.cs
danigian: Good catch. Fixed with 19db0350a45342831cfdc7189135d59f02ef96b1
Resolve if ok

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
```diff
@@ -73,13 +74,32 @@ private async Task<LoRaDevice> RegisterCoreAsync(IoTHubDeviceInfo deviceInfo, Ca
...
+                        // if the created client is registered, release it
+                        if (!ReferenceEquals(loRaDeviceClient, ((IIdentityProvider<ILoRaDeviceClient>)this.connectionManager.GetClient(loRaDevice)).Identity))
+                        {
+                            this.logger.LogDebug("leaked connection found");
+                        }
+                    }
+                    catch (ManagedConnectionException) { }
+                }
+
+                await this.connectionManager.ReleaseAsync(loRaDevice);
```
atifaziz, line 101: nit: Isn't this redundant with the next line, where the device is disposed and which eventually ends up doing the same?

danigian: @p-schuler, I agree with Atif. Did you have anything particular in mind here? Can we get rid of the additional ReleaseAsync?

atifaziz (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs
p-schuler: you can remove it - I believe this was a bit historical from how this developed (having async support on the close connection first and not the LoRaDevice.Dispose)

PR #1551, opened 2022-02-24
Add unit tests for `ExclusiveProcessor<>`

### LINKED ISSUE #1531: [Story] Add unit tests for `ExclusiveProcessor`
## Description

`ExclusiveProcessor<>` added in PR #1517 has [indirect (90+%) code coverage](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1517#discussion_r811666410), but we want unit tests that tests the class directly.

## Acceptance criteria

- [ ] `ExclusiveProcessor<>` is directly tested with unit tests

atifaziz:
# PR for issue #1531

## What is being addressed

Unit tests that directly exercise `ExclusiveProcessor<>` were missing.

## How is this addressed

Unit tests have been added.

### DIFF
--- a/Tests/Unit/NetworkServer/ExclusiveProcessorTests.cs
+++ b/Tests/Unit/NetworkServer/ExclusiveProcessorTests.cs
@@ -1,0 +1,236 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Unit.NetworkServer
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using LoRaWan.NetworkServer;
+    using Xunit;
+    using static MoreLinq.Extensions.IndexExtension;
+    using static MoreLinq.Extensions.ShuffleExtension;
+
+    public class ExclusiveProcessorTests
+    {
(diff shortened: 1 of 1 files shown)


Tests/Unit/NetworkServer/ExclusiveProcessorTests.cs
```diff
@@ -0,0 +1,230 @@
...
+        public async Task ProcessAsync_Is_Fifo_By_Default(int processorCount)
+        {
+            var processorIds = Enumerable.Range(1, processorCount).ToArray();
+            var (submittedList, processedList) = await TestProcessing(this.subject, processorIds);
+
+            Assert.Equal(processedList, submittedList);
+            Assert.Equal(processedList, processorIds);
+        }
+
+        private sealed class LifoScheduler<T> : ExclusiveProcessor<T>.IScheduler
```
bastbu, line 168: [nit] Do you think it would be adding a test to assert at which point the interrupt event is going to be called? I think it might be worth it as "documentation", I had to debug through the code to understand when it's expected.

bastbu (approved):

Tests/Unit/NetworkServer/ExclusiveProcessorTests.cs
atifaziz: Those events are purely for diagnostics and incidentally also help with tests. They are not critical to the overall operation. The other events are not tested either so I think adding for `Interrupted` would make it look somewhat exceptional. I would even opt for removing the even entirely.

Issue #1540, opened 2022-02-24
Unobserved task executions in `DefaultLoRaDataRequestHandler`


bastbu:
## Expected Behavior

We should either log or observe all task executions.

## Current Behavior

In `DefaultLoRaDataRequestHandler`, [L499](https://github.com/Azure/iotedge-lorawan-starterkit/blob/9f3f2fc54ada91ead9c7b220d7b6a97012e547d2/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs#L489) we do not observe the task executions, and we do not log all exceptions (only `OperationCanceledException`s) in the `CompleteAsync`, `AbandonAsync` or `RejectAsync` itself. The same problem is for the unobserved `SendDownstreamAsync`.

We should ensure that we observe or log these exceptions. If we decide to log them, we should keep track of the unobserved exceptions in our custom metric.

PR #1552, opened 2022-02-24
Handle unobserved tasks

bastbu:
# PR for issue #1540

## What is being addressed

When processing messages we do not observe some tasks due to time restrictions to hit receive windows. Some of the exceptions that occur during this go unobserved. 

We defer awaiting these tasks until the end of the processing, but after we sent the downstream messages, we await on all the tasks that we did not observe before.

Note: this change means that we will potentially take longer to process a single message, in case one of the deferred tasks takes a longer time to finish. Given that we serialize message processing per device, this could impact performance.

### DIFF
--- a/Tests/Unit/NetworkServer/ExclusiveProcessorTests.cs
+++ b/Tests/Unit/NetworkServer/ExclusiveProcessorTests.cs
@@ -1,0 +1,230 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Unit.NetworkServer
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using LoRaWan.NetworkServer;
+    using Xunit;
+    using static MoreLinq.Extensions.IndexExtension;
+    using static MoreLinq.Extensions.ShuffleExtension;
+
+    public class ExclusiveProcessorTests
+    {
+        private readonly ExclusiveProcessor<int> subject = new();
+        private readonly DateTimeOffset testStartTime = DateTimeOffset.UtcNow;
+
+        [Fact]
+        public async Task TryProcessAsync_With_Successful_Function_Returns_Outcome()
+        {
+            var outcome = await this.subject.TryProcessAsync(1, () => Task.FromResult(42));
+
+            Assert.True(outcome.SubmissionTime >= this.testStartTime);
+            Assert.NotEqual(TimeSpan.Zero, outcome.WaitDuration);
+            Assert.NotEqual(TimeSpan.Zero, outcome.RunDuration);
+            Assert.True(outcome.Result.IsCompletedSuccessfully);
+            Assert.Equal(42, await outcome.Result);
+        }
+
+        [Fact]
+        public async Task TryProcessAsync_With_Canceled_Function_Returns_Outcome()
+        {
+            using var cts = new CancellationTokenSource();
+            cts.Cancel();
+
+            var outcome = await this.subject.TryProcessAsync(1, () => Task.FromCanceled<int>(cts.Token));
+
+            Assert.True(outcome.SubmissionTime >= this.testStartTime);
+            Assert.NotEqual(TimeSpan.Zero, outcome.WaitDuration);
+            Assert.NotEqual(TimeSpan.Zero, outcome.RunDuration);
+            Assert.True(outcome.Result.IsCanceled);
+        }
+
+        [Fact]
+        public async Task TryProcessAsync_With_Erroneous_Function_Returns_Outcome()
+        {
+#pragma warning disable CA2201 // Do not raise reserved exception types
+            var ex = new ApplicationException();
+#pragma warning restore CA2201 // Do not raise reserved exception types
+
+            var outcome = await this.subject.TryProcessAsync(1, () => Task.FromException<int>(ex));
+
+            Assert.True(outcome.SubmissionTime >= this.testStartTime);
+            Assert.NotEqual(TimeSpan.Zero, outcome.WaitDuration);
+            Assert.NotEqual(TimeSpan.Zero, outcome.RunDuration);
+            Assert.True(outcome.Result.IsFaulted);
+            var aggregateException = outcome.Result.Exception;
+            Assert.NotNull(aggregateException);
+            Assert.Same(ex, Assert.Single(aggregateException.InnerExceptions));
+        }
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -60,14 +59,15 @@ public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
...
             this.c2dMessageTooLong = meter?.CreateCounter<int>(MetricRegistry.C2DMessageTooLong);
-            this.unhandledExceptionCount = meter?.CreateCounter<int>(MetricRegistry.UnhandledExceptions);
         }
 
         public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaRequest request, LoRaDevice loRaDevice)
         {
             if (request is null) throw new ArgumentNullException(nameof(request));
             if (loRaDevice is null) throw new ArgumentNullException(nameof(loRaDevice));
 
+            var deferredTasks = new List<Task>();
```
p-schuler, line 69: If we do not have a downstream message or any other of the edge cases triggering the unobserved tasks, this allocation will never be used. I would suggest to wrap it in a local function and only allocate when we need it.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -143,7 +143,7 @@ public async Task<bool> SendEventAsync(LoRaDeviceTelemetry telemetry, Dictionary
 
                     return true;
                 }
-                catch (OperationCanceledException ex) when (ExceptionFilterUtility.True(() => this.logger.LogError($"could not send message to IoTHub/Edge with error: {ex.Message}")))
+                catch (OperationCanceledException) when (ExceptionFilterUtility.True(() => this.logger.LogError("could not send message to IoTHub/Edge due to timeout.")))
```
p-schuler, line 146: pls always pass the original exception to the logger. The stack traces where the timeout happened can be crucial.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -171,7 +171,7 @@ public async Task<Message> ReceiveAsync(TimeSpan timeout)
 
                 return msg;
             }
-            catch (OperationCanceledException ex) when (ExceptionFilterUtility.True(() => this.logger.LogError($"could not retrieve cloud to device message with error: {ex.Message}")))
+            catch (OperationCanceledException) when (ExceptionFilterUtility.True(() => this.logger.LogError("could not retrieve cloud to device message due to timeout.")))
```
p-schuler, line 174: pls always pass the original exception to the logger. The stack traces where the timeout happened can be crucial.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -201,7 +201,7 @@ private async Task<bool> ExecuteC2DOperationAsync(Message cloudToDeviceMessage,
                 this.logger.LogDebug("done processing '{OperationName}' on cloud to device message, id: '{MessageId}'.", operationName, messageId);
                 return true;
             }
-            catch (OperationCanceledException ex) when (ExceptionFilterUtility.True(() => this.logger.LogError(ex, "'{OperationName}' failed on cloud to device message (id: {MessageId}) with error: '{ExceptionMessage}'.", operationName, messageId, ex.Message)))
+            catch (OperationCanceledException) when (ExceptionFilterUtility.True(() => this.logger.LogError("'{OperationName}' timed out on cloud to device message (id: {MessageId}).", operationName, messageId)))
```
p-schuler, line 204: pls always pass the original exception to the logger. The stack traces where the timeout happened can be crucial.

Tests/Integration/ProcessingTests.cs
```diff
@@ -1697,7 +1525,7 @@ public async Task When_Receiving_Device_Message_And_Loading_Device_Fails_Second_
             await Task.Delay(150);
 
             using var request2 = CreateWaitableRequest(simDevice.CreateUnconfirmedDataUpMessage("2"));
-            messageDispatcher.DispatchRequest(request2);
+            messageDispatcher.Value.DispatchRequest(request2);
```
p-schuler, line 1528: Instead of adding .Value all over we could have initialized a stack variable `var messageDispatcher = messageDispatcherWrapper.Value` for example and keep the IDisposable as a "wrapper".

p-schuler: Otherwise this is a nice refactoring and helps reducing code and maintainability when changes are made in that area.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
bastbu: Done in 6fcc890f3a924576e042a9b5fc4c635d4bed2343 (as well as other occurrences in the same file)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
bastbu: Done in 6fcc890f3a924576e042a9b5fc4c635d4bed2343 (as well as other occurrences in the same file)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: Makes sense, applied here: af9adfef245ee9a06e6851b3b8a610d4209e35c7

Tests/Integration/ProcessingTests.cs
bastbu: Done with 47b3e7ed5b61e34b43221a3a37da0ea73a4fc377.

p-schuler (approved):
LGTM

Issue #1554, opened 2022-02-25
Integration test failure: `The active test run was aborted. Reason: Test host process crashed`


bastbu:
## Expected Behavior

Integration tests should pass consistently

## Current Behavior

Sometimes the integration tests fail with the following exception:

```
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.InvalidOperationException: There is no currently active test.
   at Xunit.Sdk.TestOutputHelper.GuardInitialized() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\TestOutputHelper.cs:line 51
   at Xunit.Sdk.TestOutputHelper.QueueTestOutput(String output) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\TestOutputHelper.cs:line 66
   at LoRaWan.Tests.Common.TestOutputLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter) in /home/runner/work/iotedge-lorawan-starterkit/iotedge-lorawan-starterkit/Tests/Common/TestOutputLogger.cs:line 34
   at Microsoft.Extensions.Logging.Logger`1.Microsoft.Extensions.Logging.ILogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.LoggerExtensions.Log(ILogger logger, LogLevel logLevel, EventId eventId, Exception exception, String message, Object[] args)
   at Microsoft.Extensions.Logging.LoggerExtensions.LogDebug(ILogger logger, String message, Object[] args)
   at LoRaWan.NetworkServer.LoRaDeviceClientConnectionManager.SynchronizedLoRaDeviceClient.<>c__DisplayClass29_0.<<DisconnectAsync>b__0>d.MoveNext() in /home/runner/work/iotedge-lorawan-starterkit/iotedge-lorawan-starterkit/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientConnectionManager.cs:line 290
--- End of stack trace from previous location ---
[...]

PR #1555, opened 2022-02-25
Stabilize test host which crashes due to logging in `TestOutputLogger`

bastbu:
# PR for issue #1554

## What is being addressed

In some cases, logging to a `TestOutputLogger` happens after the test has finished, since the lifetime of some fixtures can be longer than that of a test. If we log e.g. inside a cache eviction of an entry `IMemoryCache`, this will cause a log to the `ITestOutputHelper` after it can be used. This causes the test host to crash.

## How is this addressed

Logging to the `ITestOutputHelper` happens on a best-effort basis. The exceptions in logging due to the `IMemoryCache` eviction will still happen, but we catch them and do not crash the test host.

### DIFF
--- a/Tests/Common/TestOutputLogger.cs
+++ b/Tests/Common/TestOutputLogger.cs
@@ -35,7 +35,16 @@ namespace LoRaWan.Tests.Common
             if (!IsEnabled(logLevel)) return;
 
             var message = formatter(state, exception);
-            this.testOutputHelper.WriteLine(message);
+
+            try
+            {
+                this.testOutputHelper.WriteLine(message);
+            }
+            catch (Exception)
+            {
+                // best-effort logging in case testOutputHelper has already been disposed. Fixes:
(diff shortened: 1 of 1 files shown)


Tests/Common/TestOutputLogger.cs
```diff
@@ -35,7 +35,16 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
             if (!IsEnabled(logLevel)) return;
 
             var message = formatter(state, exception);
-            this.testOutputHelper.WriteLine(message);
+
+            try
+            {
+                this.testOutputHelper.WriteLine(message);
+            }
+            catch (Exception)
```
atifaziz, line 43: What about making this specific to `InvalidOperationException`, which is the [quoted case](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1554) (for now)?

```suggestion
            catch (InvalidOperationException)
```

atifaziz (approved):
Good catch and thanks for fixing this!

Issue #1497, opened 2022-02-15
Implement function based connection state change notification


p-schuler:
There are multiple scenarios, where potentially, a connection owning LNS does not receive a subsequent message from a device it has an open connection:
- temporary blocked path - heavy equipment avoids transmission
- device is roaming

When a owning LNS has an open connection for a particular device, it will not be able to release it if the system is based on message processing from the device only. 

Here we add a detection mechanism to determine, when a connection ownership change is happening and enqueue a C2D message for the LNS to drop the particular device connection.

Acceptance Criteria

- [x] #1528
- [ ] #1529
- [x] edge cases are covered (considering other threads can use the connection for example -> use operations on the device)
  - This is now handled by PR #1517  
- [x] metrics are added to track these operations (forced connection drop via C2D)
- [x] unit test, integration tests and E2E tests covering the scenario
- [ ] #1578

PR #1560, opened 2022-02-28
Will not merge: Feature/1497 function notification to lns

spygi:
# PR for issue #1497 

## What is being addressed

When the winning LNS changes on the function side, we have to inform the losing LNS proactively that it lost the race. The reason why we do this is in order for its Edge Hub to not retry to create the connection which would result in a connection ping-pong.

## How is this addressed

- A new direct method is called from the `DeduplicationExecutionItem` when a connection switch happens.
- When LNS receives it (in `ModuleConnectionHost`), it sets itself as the losing LNS and closes the connection.
- Unit tests added for LNS (`ModuleConnectionHostTest`) and the function side (`MessageDeduplicationTests`), integration tests for Function (`DeduplicationTestWithRedis`) and a simulation test (`SimulatedLoadTest`) for the whole flow.


spygi:
Due to the big git diff, we decided not to merge this PR. Instead it was superseded by #1577

Issue #1561, opened 2022-02-28
MAC commands are serialized "twice"


bastbu:
We serialize MAC command payloads "twice" in `DefaultLoRaDataRequestHandler` by first getting the string representation of a MAC command, and then serializing it.

```csharp
JsonConvert.SerializeObject(payloadData.MacCommands[i].ToString(), Formatting.None)
```

https://github.com/Azure/iotedge-lorawan-starterkit/blob/df6773a4204b492096c4da22c45587ccc2f4b8b8/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs#L734

Issue #1284, opened 2022-01-14
[Story] Add unit tests for MAC commands (`MacCommand` subclasses)


atifaziz:
## Description

Add unit tests for implementations of MAC commands (`MacCommand` subclasses).

## Acceptance criteria

All of the following `MacCommand` subclasses have unit tests with good coverage:

- [x] `DevStatusAnswer`
- [x] `DevStatusRequest`
- [x] `DutyCycleAnswer`
- [x] `DutyCycleRequest`
- [x] `LinkADRAnswer`
- [x] `LinkADRRequest`
- [x] `LinkCheckAnswer`
- [x] `LinkCheckRequest`
- [x] `NewChannelAnswer`
- [x] `NewChannelRequest`
- [x] `RXParamSetupAnswer`
- [x] `RXParamSetupRequest`
- [x] `RXTimingSetupAnswer`
- [x] `RXTimingSetupRequest`
- [x] `TxParamSetupAnswer`
- [x] `TxParamSetupRequest`

PR #1562, opened 2022-02-28
MAC command unit tests

bastbu:
# PR for issue #1284

## What is being addressed

With this PR we add unit tests for the most common methods of MAC commands. This includes tests for:

- The JSON deserialization (if applicable)
- Constructors
- Creation from byte arrays
- The `ToBytes` method

We do not add tests for the `ToString` overload or the JSON serialization, as this is related to #1561 and will be handled as part of that.

## How is this addressed

We use a shared base class to set up the minimum required tests for each MAC command. In addition, each subclass of the new test class can add more detailed tests if there is non-standard behavior.

Unit-testing the MAC commands revealed a couple of bugs:

- Payloads of some CIDs were deserialized as `LinkADRCmd`
- Some minor bugs were present in some constructors
- The `Length` was not JSON-ignored (for serialization, which is currently unused)
- #1575


atifaziz (changes requested):
I noticed there's `MacCommandTest` _and_ `MacCommandTests` (classes are under the same namespace but in different folders of the test project). If these are duplicates or overlap then the former should be merged into the latter or removed if redundant.

bastbu:
@atifaziz Makes sense, I renamed the test class in 681266461d636e3342079d2d86578fcc31e01834

atifaziz (commented):
> The tests for the JSON converter tests general properties of the JSON conversion (handling of lists, case-sensitiveness, illegal CIDs etc), while the new tests assert on the JSON representation of each MAC command (which can be deserialized)

That makes sense. Certainly the overlap in the name had me confused. Thanks for the rename.

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,293 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using global::LoRaTools;
+    using global::LoRaTools.Mac;
+    using global::LoRaTools.Regions;
+    using LoRaWan.Tests.Common;
+    using Newtonsoft.Json;
+    using Xunit;
+
+    public abstract class MacCommandTests
```
atifaziz, line 18: What about making this generic and then constraining the type argument to `MacCommand`, as in:

```c#
public class MacCommandTests<T> where T : MacCommand`
{
    /* ... */
}
```

I think it will simplify a number of things like making `Subject` a `T` and removing the need for a type parameter on `DeserializationTest` and `FromBytesTest`.

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,293 @@
...
+    {
+        /// <summary>
+        /// Gets the CID of the MAC command.
+        /// </summary>
+        public abstract Cid Cid { get; }
```
atifaziz, line 23: These properties should use the `protected` modifier. Also, I don't think they need to be abstract. If `MacCommandTest` has a constructor to take the values of these properties as inputs, which would be the same enforcement as making them abstract here, then the properties could just be non-virtual accessors.

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,293 @@
...
+            private Predicate<LinkADRRequest> Assert => actual => actual.Cid == Cid
+                                                               && actual.DataRate == Subject.DataRate
+                                                               && actual.TxPower == Subject.TxPower
+                                                               && actual.ChMask == Subject.ChMask
+                                                               && actual.ChMaskCntl == Subject.ChMaskCntl
+                                                               && actual.NbRep == Subject.NbRep;
+
+            [Fact]
+            public void Deserializes_Correctly() =>
+                DeserializationTest(Assert, @"{""cid"":3,""dataRate"":1,""txPower"":2,""chMask"":3,""chMaskCntl"":4,""nbRep"":5}");
```
atifaziz, line 159: nit: If `DeserializationTest` called `JsonUtil.Strictify` on the JSON, then all the JSON text could be easier to read (by not needing quoted JSON object members names).

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,293 @@
...
+            public override int Length => 2;
+
+            [Fact]
+            public void Deserializes_Correctly() =>
+                DeserializationTest<RXTimingSetupRequest>(actual => actual.Cid == Cid && actual.Settings == Subject.Settings,
+                                                          @"{""cid"":8,""settings"":1}");
+        }
+
+        [Fact]
+        public void ToBytes_Success() => ToBytes_Internal(Subject, DataBytes);
```
atifaziz, line 274: nit: This and remaining methods are part of the base class `MacCommandTests` and so should be moved before the subclasses. It's somewhat odd to have subclasses sandwiched in the middle of the base class definition.

atifaziz: Is it possible to avoid the base class having tests? I believe this throws off some test runners and produces warnings about duplicate tests.

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,293 @@
...
+        public sealed class NewChannelRequestTests : MacCommandTests
+        {
+            public override Cid Cid => Cid.NewChannelCmd;
+            public override NewChannelRequest Subject => new NewChannelRequest(1, 2, 3, 4);
+            public override IReadOnlyList<byte> DataBytes => new byte[] { 1, 2, 0, 0, 0b110100 };
+            public override int Length => 6;
+
+            [Fact]
+            public void Deserializes_Correctly() =>
+                DeserializationTest<NewChannelRequest>(actual => actual.Cid == Cid
```
atifaziz, line 212: There might be a trick to avoid these equality tests by simply returning an anonymous object given a command. Since anonymous types implement equality, you could then use a regular `Assert.Equal` to compare the subject and the actual value. It would also have the benefit of more helpful assertion failure messages than `Assert.True` does.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRRequest.cs
```diff
@@ -47,8 +47,9 @@ public LinkADRRequest(ushort datarate, ushort txPower, ushort chMask, ushort chM
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LinkADRRequest"/> class. For tests to serialize from byte.
+        /// Intended for test use only.
```
atifaziz, line 50: We should avoid the practice of constructors for testing purposes only.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MACCommandJsonConverter.cs
```diff
@@ -90,6 +86,9 @@ void GetValue(string propertyName, out JToken value)
                         serializer.Populate(item.CreateReader(), cmd);
                         return cmd;
                     }
+                    case Cid.Zero:
+                    case Cid.One:
```
atifaziz, line 90: 0 and 1 are invalid CIDs. We should consider removing these from `Cid` entirely.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/LinkADRAnswer.cs
```diff
@@ -33,7 +33,8 @@ public LinkADRAnswer(byte powerAck, bool dataRateAck, bool channelMaskAck)
         {
             Cid = Cid.LinkADRCmd;
             Status |= (byte)((byte)(powerAck & 0b00000011) << 2);
-            Status |= (byte)((byte)(dataRateAck ? 1 << 1 : 0 << 1) | (byte)(channelMaskAck ? 1 : 0));
+            Status |= (byte)((dataRateAck ? 1 : 0) << 1);
+            Status |= (byte)(channelMaskAck ? 1 : 0);
```
atifaziz, line 37: nit: if we are going to _improve this_, might as well go the whole way:

```c#
            Status = unchecked((byte)(((powerAck & 0b00000011) << 2)
                                      | (dataRateAck ? 2 : 0)
                                      | (channelMaskAck ? 1 : 0)));
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MACCommandJsonConverter.cs
bastbu: I had to keep `Cid.Zero` due to CA1008, but I renamed it to `None` to make it clearer where it comes from. Addressed with b09fe9b16edb8b8ce78be4b8cb6e618b41fa81de

Tests/Unit/LoRaTools/MacCommandTests.cs
bastbu: Done in c6fa017c9295979c98b31f8f3a9c954a3cf56b99

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MACCommandJsonConverter.cs
atifaziz: Since we are using the CID values from the spec directly as the enumeration member values, `Cid.None` doesn't bring much value (in fact, it has the downside that we can't rely on `Enum.IsDefined` to valid the CID). I would opt to disable CA1008 as we've done in other cases like `Bandwidth`, `ReceiveWindowNumber`, `SpreadingFactor`, etc.

Tests/Unit/LoRaTools/MacCommandTests.cs
bastbu: Addressed this in 973733e27f11acbf4ac0e2e12ed5326e4f435f9c

Tests/Unit/LoRaTools/MacCommandTests.cs
bastbu: Addressed this with 973733e27f11acbf4ac0e2e12ed5326e4f435f9c. As discussed, we are keeping the tests inheriting from a base class for now.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/MACCommandJsonConverter.cs
bastbu: Addressed this with 76af7d871d2a8d99577995ee69b2615398b0835f. This caught one instance where we logged a wrong CID, this is now also fixed as part of the same commit.

Tests/Unit/LoRaTools/MacCommandTests.cs
bastbu: Nice trick, applied with 78ee5aa5d208c6bbe74ab064f6ae84de1a6c3e20

Tests/Unit/LoRaTools/MacCommandTests.cs
bastbu: c676cd90bd84fc9909b847803d7bd7b9662946d8

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -167,35 +162,34 @@ public sealed class LinkAdrAnswerTests : MacCommandTests<LinkADRAnswer>
...
-                                                                && !actual.CHMaskAck,
-                                                         bytes => new LinkADRAnswer(new ReadOnlySpan<byte>(bytes)),
-                                                         DataBytes);
+        public void FromBytes_Success() => FromBytesTest(actual => new { actual.Cid, actual.PowerAck, actual.DRAck, actual.CHMaskAck },
+                                                         bytes => new LinkADRAnswer(new ReadOnlySpan<byte>(bytes)));
     }
 
     public sealed class LinkAdrRequestTests : MacCommandTests<LinkADRRequest>
     {
+        private static readonly Func<LinkADRRequest, object> Transform = actual => new
```
atifaziz, line 171: nit: suggest to rename all `actual` to `cmd` or `command` since it doesn't necessarily represent any actual from tests (sometimes it represents the subject).

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,292 @@
...
+        public void Init_Throws_When_Eirp_Is_Invalid() =>
+            Assert.Throws<ArgumentException>(() => new TxParamSetupRequest(new DwellTimeSetting(false, false, 16)));
+
+        public static TheoryData<DwellTimeSetting, byte> ToBytes_Theory_Data() => TheoryDataFactory.From(new (DwellTimeSetting, byte)[]
+        {
+                (new DwellTimeSetting(false, false, 0), 0b0000_0000),
+                (new DwellTimeSetting(true, false, 0), 0b0010_0000),
+                (new DwellTimeSetting(false, true, 0), 0b0001_0000),
+                (new DwellTimeSetting(false, false, 13), 0b0000_1101),
+                (new DwellTimeSetting(true, true, 15), 0b0011_1111),
```
atifaziz, line 90: nit: over-indented:

```suggestion
            (new DwellTimeSetting(false, false, 0), 0b0000_0000),
            (new DwellTimeSetting(true, false, 0), 0b0010_0000),
            (new DwellTimeSetting(false, true, 0), 0b0001_0000),
            (new DwellTimeSetting(false, false, 13), 0b0000_1101),
            (new DwellTimeSetting(true, true, 15), 0b0011_1111),
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/Mac/Cid.cs
```diff
@@ -3,17 +3,17 @@
 
 namespace LoRaTools
 {
+#pragma warning disable CA1008 // Enums should have zero value
```
atifaziz, line 6: Consider adding a justification:

```suggestion
#pragma warning disable CA1008 // Enums should have zero value (but values reflect CID from spec which has none)
```

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,292 @@
...
+    {
+        private static readonly Func<LinkADRRequest, object> Transform = actual => new
+        {
+            actual.Cid,
+            actual.DataRate,
+            actual.TxPower,
+            actual.ChMask,
+            actual.ChMaskCntl,
+            actual.NbRep
+        };
```
atifaziz, line 179: Curious about 2 things:

- Why is this exceptionally expressed differently than in all other tests?
- Why isn't is just a regular method, like below?

```suggestion
        private static object Transform(LinkADRRequest actual) => new
        {
            actual.Cid,
            actual.DataRate,
            actual.TxPower,
            actual.ChMask,
            actual.ChMaskCntl,
            actual.NbRep
        };
```

bastbu: What do you mean by expressed differently? I didn't inline it, since it has two references. All other "transform" methods have only a single reference. Will apply your suggestion though.

Tests/Unit/LoRaTools/MacCommandTests.cs
bastbu: Done in 9f2977087017f9d730427546c87734e38aa2e3ee

Tests/Unit/LoRaTools/MacCommandTests.cs
```diff
@@ -0,0 +1,293 @@
...
+    {
+        public DevStatusAnswerTests() :
+            base(Cid.DevStatusCmd,
+                 new DevStatusAnswer(1, 2),
+                 new byte[] { 1, 2 })
+        { }
+
+        [Fact]
+        public void FromBytes_Success() => FromBytesTest(command => new { command.Cid, command.Battery, command.Margin },
+                                                         bytes => new DevStatusAnswer(new ReadOnlySpan<byte>(bytes)));
```
atifaziz, line 118: nit: you can use `AsSpan()` here and elsewhere for similar cases:

```suggestion
                                                         bytes => new DevStatusAnswer(bytes.AsSpan()));
```

atifaziz (approved):
I think the MAC command tests use an inheritance model that seems to be more reflective of theory data (e.g. all the protected properties of `MacCommandTests<>` are never accessed by its subclasses and base class calls seem like data) so would be best to see if it can be modeled more naturally that way. Otherwise LGTM and thanks for spotting and fixing all the bugs through the addition of these tests.

Tests/Unit/LoRaTools/MacCommandTests.cs
bastbu: That's indeed better, done in 9b13c9464b2059b1e6da20fc58dce7586574bc31

bastbu:
OTAA test succeeded after retry in https://github.com/Azure/iotedge-lorawan-starterkit/runs/5403248757?check_suite_focus=true

PR #1563, opened 2022-02-28
Mention LNS discovery in deployment scenarios

bastbu:
## What is being addressed

We link the LNS discovery documentation from the deployment scenario overview.

### DIFF
--- a/docs/user-guide/deployment-scenarios.md
+++ b/docs/user-guide/deployment-scenarios.md
@@ -24,6 +24,8 @@ updates](station-firmware-upgrade.md) for example.
 
 ## Scalability & Availability
 
+There are multiple strategies that you can take to improve the availability of the starter kit listed below. For all the scenarios below, you can further increase the availability of a concentrator by using a highly available [LNS discovery service](lns-discovery.md).
+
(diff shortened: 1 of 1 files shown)


danigian (approved):

docs/user-guide/deployment-scenarios.md
```diff
@@ -24,6 +24,8 @@ updates](station-firmware-upgrade.md) for example.
 
 ## Scalability & Availability
 
+There are multiple strategies that you can take to improve the availability of the starter kit listed below. For all these scenarios, you can further increase the availability of a concentrator by using a highly available [LNS discovery service](lns-discovery.md).
```
p-schuler, line 27: it's not for all scenarios. you can certainly enable discovery for a single LNS deployment, but that would not make much sense. I would mention this is only for multi LNS deployments (where LBS can reach multiple LNS)

p-schuler (approved):
fixing the suggestion to use this everywhere and then LGTM

docs/user-guide/deployment-scenarios.md
bastbu: Done in b205e5a1111c951ef75588799329ae3699c8b2d7

Issue #1498, opened 2022-02-15
Manage connection correctly when resetting device


p-schuler:
When resetting the device, we do update the twins (resetting the framecounter). If we are losing the race, we should not open the connection. We should try to avoid the processing if we are the losing gateway, for that we would need to get past that call to determine the winner before making the call.

p-schuler:
@danigian

PR #1565, opened 2022-02-28
Deferring reset frame counter twin changes

danigian:
# PR for issue #1498

## What is being addressed

When resetting the device, we updated the twins even before the validation of the request. With the new session affinity changes, if a LNS is losing the race, it should not open any connection.

## What changed

- [x] Both FrameCounterUpdateStrategy (Multi and single gw) are not immediately updating the FCnt but only resetting in memory state, for later deferred update of the fields
- [x] DefaultLoRaDataRequestHandler has been changed to eventually update the twin only after establishing if current LNS is owning the connection. In addition we got rid of "stationEuiChanged" previously and wrongfully used for forcing update the twin.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/MultiGatewayFrameCounterUpdateStrategy.cs
```diff
@@ -24,12 +24,7 @@ public async Task<bool> ResetAsync(LoRaDevice loRaDevice, uint fcntUp, string ga
 
             loRaDevice.ResetFcnt();
 
-            if (await InternalSaveChangesAsync(loRaDevice, force: true))
-            {
-                return await this.loRaDeviceAPIService.ABPFcntCacheResetAsync(loRaDevice.DevEUI, fcntUp, gatewayId);
-            }
-
-            return false;
+            return await this.loRaDeviceAPIService.ABPFcntCacheResetAsync(loRaDevice.DevEUI, fcntUp, gatewayId);
```
p-schuler, line 27: now it could happen we reset the function but not the twins. what could be the side effect of that? I don't think I see a problem with this.

Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
```diff
@@ -156,12 +156,12 @@ public ConcentratorDeduplicationDataMessagesIntegrationTests(ITestOutputHelper t
...
-        [DeduplicationTestData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", DeduplicationMode.Drop, expectedFrameCounterResets: 1, expectedBundlerCalls: 1, expectedFrameCounterDownCalls: 1, expectedMessagesUp: 1, expectedMessagesDown: 1, expectedTwinSaves: 1)] // duplicate
-        [DeduplicationTestData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", DeduplicationMode.Mark, expectedFrameCounterResets: 1, expectedBundlerCalls: 1, expectedFrameCounterDownCalls: 1, expectedMessagesUp: 2, expectedMessagesDown: 1, expectedTwinSaves: 2)] // soft duplicate
-        [DeduplicationTestData("11-11-11-11-11-11-11-11", "22-22-22-22-22-22-22-22", DeduplicationMode.None, expectedFrameCounterResets: 1, expectedBundlerCalls: 1, expectedFrameCounterDownCalls: 1, expectedMessagesUp: 2, expectedMessagesDown: 1, expectedTwinSaves: 2)] // soft duplicate
+        [DeduplicationTestData("11-11-11-11-11-11-11-11", "11-11-11-11-11-11-11-11", DeduplicationMode.Drop, expectedFrameCounterResets: 1, expectedBundlerCalls: 2, expectedFrameCounterDownCalls: 2, expectedMessagesUp: 1, expectedMessagesDown: 2, expectedTwinSaves: 4)] // resubmission with drop
```
p-schuler, line 159: why do we get double the twin safes?

p-schuler (changes requested):
I would like to understand the change in the amount of twin save from 2 to 4 in the integration tests

Tests/Integration/ConcentratorDeduplicationDataMessagesIntegrationTests.cs
danigian: What we are asserting is the call to SaveChangesToDeviceAsync in the DefaultLoRaDataRequestHandler and not the actual UpdateReportedPropertiesAsync on the Client.

We had a different count before, just because the original code was not invoking SaveChangesToDeviceAsync but directly SaveChangesAsync of LoRaDevice.

Let's discuss more if needed offline

Issue #1564, opened 2022-02-28
Throttle device/client disposals


atifaziz:
## Description

Currently, `Parallel.ForEachAsync` is used to dispose devices & clients, but `Parallel.ForEachAsync` is better-suited for CPU-bound operations. An alternative (and potentially general) implementation would simply throttle `IAsyncDisposable.DisposeAsync` calls, most of which could complete on a single logical thread for very quick disposals (completing synchronously).

## Acceptance criteria

- [ ] `Parallel.ForEachAsync` is no longer used
- [ ] Unit tests are added for an alternative

PR #1566, opened 2022-02-28
Throttle disposal of devices & clients

atifaziz:
# PR for issue #1564

## What is being addressed

Throttled disposal of clients & devices without using `Parallel.ForEachAsync` that's better-suited for CPU-bound operations.

## How is this addressed

Added an extension method for disposing a sequence `IAsyncDisposable` objects with a concurrency argument.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DisposableExtensions.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DisposableExtensions.cs
@@ -1,0 +1,61 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+namespace LoRaWan.NetworkServer
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    internal static class DisposableExtensions
+    {
+        /// <remarks>
+        /// In general <see cref="IAsyncDisposable.DisposeAsync"/> implementations are not
+        /// expected to throw exceptions. If any of the disposable objects throw exceptions then
+        /// the behaviour of this method is undefined.
+        /// </remarks>
+        public static async ValueTask DisposeAllAsync(this IEnumerable<IAsyncDisposable> disposables, int concurrency)
+        {
+            ArgumentNullException.ThrowIfNull(disposables, nameof(disposables));
+            if (concurrency <= 0) throw new ArgumentOutOfRangeException(nameof(concurrency), concurrency, null);
+
+            var capacity = disposables switch
+            {
+                ICollection<IAsyncDisposable> collection => collection.Count,
+                IReadOnlyCollection<IAsyncDisposable> collection => collection.Count,
+                _ => (int?)null,
+            };
+
+            if (capacity is 0)
+                return;
+
+            using var semaphore = new SemaphoreSlim(concurrency);
+
+            var tasks = capacity is { } someCapacity ? new List<Task>(someCapacity) : new List<Task>();
+            tasks.AddRange(disposables.Select(DisposeAsync));
+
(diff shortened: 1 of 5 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DisposableExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (concurrency <= 0) throw new ArgumentOutOfRangeException(nameof(concurrency), concurrency, null);
+
+            var capacity = disposables switch
+            {
+                ICollection<IAsyncDisposable> collection => collection.Count,
+                IReadOnlyCollection<IAsyncDisposable> collection => collection.Count,
+                _ => (int?)null,
+            };
+
+            if (capacity is 0)
```
danigian, line 33: Capacity was defined as a nullable integer before, therefore we should check
```suggestion
            if (capacity is null or 0)
```

Or am I missing something?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DisposableExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+                return;
+
+            using var semaphore = new SemaphoreSlim(concurrency);
+
+            var tasks = capacity is { } someCapacity ? new List<Task>(someCapacity) : new List<Task>();
```
danigian, line 38: Why is this initial capacity needed here?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DisposableExtensions.cs
atifaziz: If `capacity` is zero then we know `disposables` is empty. If it's `null`, it means that `disposables` is a true sequence whose count cannot be known in advance so we must iterate it. This `if` just an optimization for when we know for sure the sequence is some empty collection. Hope that clears up any smoke.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DisposableExtensions.cs
atifaziz: It's an optimisation to avoid having the list to grow its internal storage dynamically as more elements/tasks are added to it. If we can tell it the initial capacity in advance and it can allocate the right-sized array (list's backing storage) from the start.

Tests/Unit/NetworkServer/DisposableExtensionsTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            var ex = await Assert.ThrowsAsync<ArgumentNullException>(async () =>
+                await DisposableExtensions.DisposeAllAsync(null!, 0));
+
+            Assert.Equal("disposables", ex.ParamName);
+        }
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(-1)]
+        public async Task DisposeAllAsync_With_Invalid_Concurrency_Throws(int concurrency)
```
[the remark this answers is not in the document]
atifaziz: Concurrency argument is defined as `int concurrency`. It's not nullable.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DisposableExtensions.cs
danigian: Let's add this as comment please

atifaziz: With e0f7ae9944194b7f27172b0b2ecdd925d0678611, I also added a test to cover this condition.

Issue #1556, opened 2022-02-25
DeferredTasks handling has to be changed to throw an aggregate exception


danigian:

danigian:
@bastbu, @atifaziz let's discuss this on monday if there's any question

atifaziz:
I prototyped what a helper for this could look like, including max concurrency and cancellation support:

```c#
static class Extensions
{
    public static async Task<(T, Task<TResult>)[]>
        WhenAllCompleted<T, TResult>(this IEnumerable<T> source,
                                     Func<T, CancellationToken, Task<TResult>> function,
                                     int maxConcurrency,
                                     CancellationToken cancellationToken)
    {
        ArgumentNullException.ThrowIfNull(source, nameof(source));
        ArgumentNullException.ThrowIfNull(function, nameof(function));
        if (maxConcurrency <= 0) throw new ArgumentOutOfRangeException(nameof(maxConcurrency), maxConcurrency, null);

        using var semaphore = new SemaphoreSlim(maxConcurrency);
        var pendingTaskList = new List<Task<(int, T, Task<TResult>)>>(source is ICollection { Count: var count } ? count : 4);

        var i = 0;
        foreach (var item in source)
            pendingTaskList.Add(RunAsync(i++, item));

        var results = new (T, Task<TResult>)[pendingTaskList.Count];

        while (pendingTaskList.Count > 0)
        {
            var completedTask = await Task.WhenAny(pendingTaskList).ConfigureAwait(false);
            pendingTaskList.Remove(completedTask);
            (i, var item, var task) = await completedTask.ConfigureAwait(false);
            results[i] = (item, task);
        }
        
        return results;
        
        async Task<(int, T, Task<TResult>)> RunAsync(int i, T item)
        {
            try
            {	        
                await semaphore.WaitAsync().ConfigureAwait(false);
                return (i, item, await Task.WhenAny(function(item, cancellationToken)).ConfigureAwait(false));
            }
            finally
            {
                semaphore.Release();
            }
        }
    }
}
```

atifaziz:
We could do this more simply with using `Task.WhenAny` on the task returned by `Task.WhenAll` since awaiting on that doesn't throw and we would get access to the aggregate of exceptions collected by the task return from `Task.WhenAll`.

PR #1567, opened 2022-02-28
Handle deferred tasks correctly

bastbu:
# PR for issue #1556

## What is being addressed

With this PR, we change the default message processing to handle the deferred tasks more correctly. With the existing code, there were two problems (mainly):

- `Task.WhenAll` throws the first exception if multiple children fail, which means that we lose some insights into what goes wrong.
- When a secondary task fails and the main processing fails, the deferred task failure hides the original exception

## How is this addressed

With this PR we handle the abovementioned shortcomings. The idea is that an exception in the "main" processing logic is always prioritized over exceptions in secondary tasks. Secondary task failures are always logged. Furthermore, we may only dispose the client connection activity after the secondary tasks finished executing. If there are any secondary task failures, we only log these failures, but do not consider processing as failed.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/TaskExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+                    result.Add(ex);
+
+                if (task.IsFaulted && task.Exception is { } someException)
+                {
+                    result.Add(someException.InnerExceptions switch
+                    {
+                        { Count: 1 } => someException.InnerExceptions[0],
+                        _ => someException
+                    });
+                }
```
atifaziz, line 38: Some patterns can be merged and the switch expression simplified to a conditional:

```suggestion
                if (task is { IsFaulted: true, Exception: { } aggregateException })
                    result.Add(aggregateException.InnerExceptions is { Count: 1 } exceptions ? exceptions[0] : aggregateException);
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/TaskExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+            var result = new List<Exception>();
+
+            foreach (var task in tasks)
+            {
+                if (!task.IsCompleted)
+                {
+                    throw new InvalidOperationException("All tasks must have completed.");
+                }
+
+                if (task.IsCompletedSuccessfully) continue;
```
atifaziz, line 26: nit: formatting:

```suggestion
                if (task.IsCompletedSuccessfully)
                    continue;
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/TaskExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    internal static class TaskExtensions
+    {
+        public static IList<Exception> GetExceptions(this IList<Task> tasks)
+        {
+            var result = new List<Exception>();
+
+            foreach (var task in tasks)
+            {
+                if (!task.IsCompleted)
```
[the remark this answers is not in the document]
bastbu: Good point, addressed this with ea5cacedaae73ea87a0f415edf3d9f50f0e1570b

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -64,514 +65,541 @@ public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
...
+            try
+            {
+                ExceptionDispatchInfo edi = null;
+                LoRaDeviceRequestProcessResult result = null;
 
-            var payloadFcntAdjusted = LoRaPayloadData.InferUpper32BitsForClientFcnt(payloadFcnt, loRaDevice.FCntUp);
-            this.logger.LogDebug($"converted 16bit FCnt {payloadFcnt} to 32bit FCnt {payloadFcntAdjusted}");
+                try
+                {
+                    result = await ProcessAsync();
```
atifaziz, line 81: Worth adding a comment here that this method modifies and updates `edi` and `result` since it's not obvious. Ideally we could have them as `out` arguments but those are forbidden on async methods/functions.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -64,514 +65,541 @@ public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
 
         public async Task<LoRaDeviceRequestProcessResult> ProcessRequestAsync(LoRaRequest request, LoRaDevice loRaDevice)
         {
-            if (request is null) throw new ArgumentNullException(nameof(request));
-            if (loRaDevice is null) throw new ArgumentNullException(nameof(loRaDevice));
+            ArgumentNullException.ThrowIfNull(request, nameof(request));
+            ArgumentNullException.ThrowIfNull(loRaDevice, nameof(loRaDevice));
 
             List<Task> deferredTasks = null;
```
atifaziz, line 71: I feel this method is a misnamed. It says the tasks are deferred, which is not true. They run concurrent with the main processing of the data request. I would call this `forkedTasks` and review related names.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -64,514 +65,541 @@ public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
...
 
-            var frameCounterStrategy = this.frameCounterUpdateStrategyProvider.GetStrategy(loRaDevice.GatewayID);
-            if (frameCounterStrategy == null)
-            {
-                this.logger.LogError($"failed to resolve frame count update strategy, device gateway: {loRaDevice.GatewayID}, message ignored");
-                return new LoRaDeviceRequestProcessResult(loRaDevice, request, LoRaDeviceRequestFailedReason.ApplicationError);
-            }
+                if (edi is { } someEdi)
+                {
+                    if (deferredTaskException is { } someDeferredTaskException)
```
atifaziz, line 106: Doesn't having this nested here mean that no errors will be logged if one or more of the forked tasks failed unless the main processing also failed? Shouldn't we be logging irrespective? That is, as long as `deferredTaskException` isn't `null`, which was established earlier?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -64,514 +65,541 @@ public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
...
+                        this.logger.LogError(someDeferredTaskException, "Processing of deferred tasks failed.");
 
-            // Contains the Cloud to message we need to send
-            IReceivedLoRaCloudToDeviceMessage cloudToDeviceMessage = null;
+                    someEdi.Throw();
+                }
+                else if (deferredTaskException is { } someDeferredTaskException)
+                {
+                    throw someDeferredTaskException;
```
atifaziz, line 113: Do we want to fail the data request processing if any of the forked tasks failed? This means that if data request processing succeeded then we'll mask it with a errors from forked tasks.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
```diff
@@ -64,514 +65,541 @@ public class DefaultLoRaDataRequestHandler : ILoRaDataRequestHandler
...
-            this.logger.LogDebug($"converted 16bit FCnt {payloadFcnt} to 32bit FCnt {payloadFcntAdjusted}");
+                try
+                {
+                    result = await ProcessAsync();
+                }
+#pragma warning disable CA1031 // Do not catch general exception types (exception is rethrown)
+                catch (Exception ex)
+#pragma warning restore CA1031 // Do not catch general exception types
+                {
+                    edi = ExceptionDispatchInfo.Capture(ex);
```
atifaziz, line 87: I am not sure this is needed and unless I am overlooking something, doesn't all of this simply boil down to the following? That is, when the data request processing finishes (successfully or otherwise), we simply log errors in forked tasks (if any).

```c#
List<Task> deferredTasks = null;
IAsyncDisposable deviceConnectionActivity = null;

try
{
    return await ProcessAsync();
}
finally
{
    if (deviceConnectionActivity is { } someDeviceConnectionActivity)
        await someDeviceConnectionActivity.DisposeAsync();

    if (deferredTasks is { } someDeferredTasks)
    {
        _ = await Task.WhenAny(Task.WhenAll(someDeferredTasks));

        var exception = someDeferredTasks.GetExceptions() switch
        {
            { Count: 1 } exs => exs[0],
            { Count: > 1 } exs => new AggregateException(exs),
            _ => null
        };

        if (exception is { } someException)
            this.logger.LogError(someException, "Processing of deferred tasks failed.");
    }
}
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: This becomes obsolete with 8a82a1f3a5182e059c5e29f4cc08c7495d6eb485. The commit makes the closure obvious, I think there is no further comment needed.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: This was simplified with 8a82a1f3a5182e059c5e29f4cc08c7495d6eb485

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: Changed with 8a82a1f3a5182e059c5e29f4cc08c7495d6eb485, will also update the PR description

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: Valid point, I changed the semantics of secondary task failures (these do not fail the message processing) in 8a82a1f3a5182e059c5e29f4cc08c7495d6eb485 and adapted the tests accordingly. Will update the PR description.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/DefaultLoRaDataRequestHandler.cs
bastbu: Renamed to `secondaryTasks` in 6b2d2025bb35c8620b01388c0439d3877d81acf4

atifaziz (approved):
LGTM. :shipit:

Issue #1568, opened 2022-02-28
Make session affinity timeout configurable


p-schuler:
The delay we are using when giving the owning LNS an edge on the leader election race should be configurable. 

We may want to be able to disable the behavior in the future (for example, when decoupling from edge).

Currently the timeout is fixed to 400ms - the timeout should be driven by the LNS twins. If <= 0 we will skip the delay. If none specified, we use a default of 400ms.

- [x] configuration is added to the LNS desired properties
- [x] tests validating disabling of the feature as well as adjusting the timeout

PR #1571, opened 2022-03-01
Connection affinity processing delay is configurable

MaggieSalak:
# PR for issue #1568 

## What is being addressed

Processing delay to be used for the LNS which is not owning the connection for a device is configurable via device twin of the Network Server.

## How is this addressed

- A new configuration option was added for the LNS to allow setting the delay (in milliseconds) which defaults to 400 ms.
- The configured delay is updated from desired twin properties on LNS module initialization and on desired properties update.
- A new method was added to `DefaultLoRaDataRequestHandler` to determine if processing delay should happen or not. If configured delay is <= 0 then the delay is considered disabled.
- Before delaying the processing in `DefaultLoRaDataRequestHandler`, we first check if delay is enabled or not. If delay is disabled, the device connection will not be dropped in `DefaultLoRaDataRequestHandler` and the `IsConnectionOwner` flag will not be changed to `false`.
- Unit and integration tests added to cover the changes.


spygi:
Maybe out of scope for this PR but shouldn't the delay also be a configuration on azuredeploy/createEdgeDevice.json when we create the devices on IoTHub? And also added on the quickstart documentation.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -204,6 +207,22 @@ private bool TryUpdateConfigurationWithDesiredProperties(TwinCollection desiredP
             return false;
         }
 
+        private void UpdateProcessingDelayWithDesiredProperties(TwinCollection desiredProperties)
```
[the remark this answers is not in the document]
MaggieSalak: The main difference between the delay configuration and the other method is that the latter expects the config of the facade function to be present. A lack of an update is treated and logged as an error, whereas for the delay this update optional. That's why I wanted to split the two cases. We can of course handle both in the same method and only treat the facade update failure as an error, let me know what you think.

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
```diff
@@ -115,6 +114,48 @@ public async Task On_Desired_Properties_Incorrect_Update_Should_Not_Update(strin
...
+            });
+
+            await moduleClient.OnDesiredPropertiesUpdate(new TwinCollection(input), null);
+            Assert.Equal(processingDelay, networkServerConfiguration.ProcessingDelayInMilliseconds);
+        }
+
+        [Theory]
+        [InlineData("{ ProcessingDelayInMilliseconds: '' }")]
+        [InlineData("{ ProcessingDelay: 200 }")]
+        public async Task On_Desired_Properties_Incorrect_Update_Should_Not_Update_Processing_Delay(string twinUpdate)
```
spygi, line 146: Shall we add tests covering also the unhappy path on init module for the delay property (like the one here for the properties update)?

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
```diff
@@ -115,6 +114,48 @@ public async Task On_Desired_Properties_Incorrect_Update_Should_Not_Update(strin
...
+        public async Task On_Desired_Properties_Correct_Update_Should_Update_Processing_Delay(int processingDelay)
+        {
+            var networkServerConfiguration = new NetworkServerConfiguration();
+            var classCMessageSender = Mock.Of<IClassCDeviceMessageSender>();
+            var loRaDeviceRegistry = Mock.Of<ILoRaDeviceRegistry>();
+            var loRaModuleClientFactory = Mock.Of<ILoRaModuleClientFactory>();
+
+            await using var moduleClient = new ModuleConnectionHost(networkServerConfiguration, classCMessageSender, loRaModuleClientFactory, loRaDeviceRegistry, loRaDeviceApiServiceBase, NullLogger<ModuleConnectionHost>.Instance, TestMeter.Instance);
+
+            Assert.Equal(400, networkServerConfiguration.ProcessingDelayInMilliseconds);
```
spygi, line 132: [nit] Did you consider making 400 a constant so that if we change the value we wouldn't need to fix these tests?

Tests/Integration/LoRaDataRequestHandlerConnectionAffinityIntegrationTests.cs
```diff
@@ -64,55 +64,89 @@ public LoRaDataRequestHandlerConnectionAffinityIntegrationTests(ITestOutputHelpe
...
             }
         }
 
+        [Theory]
+        [InlineData(null, true)]
+        [InlineData(400, true)]
+        [InlineData(1500, true)]
+        [InlineData(0, false)]
+        [InlineData(-1000, false)]
+        public void IsProcessingDelayEnbled(int? processingDelay, bool processingDelayEnabled)
```
spygi, line 142: [nit] This is more of a unit test I would say.

MaggieSalak: That's what I thought as well but the test is using `TestDefaultLoRaRequestHandler` which is part of the integration tests project and requires setting up a number of dependencies before it can be used in the actual test code so I would argue it belongs in the integration tests.

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
MaggieSalak: For sure we can add a test for module init with incorrect delay setting, but just to make sure we're on the same page, this will not be an 'unhappy path' scenario, since in this case the module init should succeed and the default delay would be used.

MaggieSalak: Added new test in df7188bbce978d98caceb7410658ed48769a7b43

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
MaggieSalak: Done in df7188bbce978d98caceb7410658ed48769a7b43

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
spygi: Agree, meant "unhappy" as in "the config is incorrect". Thanks for the test

Tests/Integration/LoRaDataRequestHandlerConnectionAffinityIntegrationTests.cs
```diff
@@ -64,55 +64,89 @@ public LoRaDataRequestHandlerConnectionAffinityIntegrationTests(ITestOutputHelpe
...
+
             // act
             _ = await this.dataRequestHandlerMock.Object.ProcessRequestAsync(this.loraRequest, this.deviceMock.Object);
 
             // assert
-            Assert.Equal(keptConnection, this.deviceMock.Object.IsConnectionOwner);
+            Assert.Equal(canProcess || !processingDelayEnabled, this.deviceMock.Object.IsConnectionOwner);
 
-            if (keptConnection)
+            if (canProcess)
```
spygi, line 113: [nit] What do you think about changing the assertions to this? It might be a bit easier to understand. Unfortunately Github doesn't allow to suggest on removed lines. 

```suggestion
                        this.deviceMock.Verify(x => x.BeginDeviceClientConnectionActivity(), canProcess ? Times.Once : Times.Never);

            if (canProcess || (!canProcess && !processingDelayEnabled))
            {
                this.deviceMock.Verify(x => x.CloseConnectionAsync(CancellationToken.None, It.IsAny<bool>()), Times.Never);
                this.dataRequestHandlerMock.Verify(x => x.SaveChangesToDeviceAsyncAssert(), Times.Once);
            }
            else
            {
                this.deviceMock.Verify(x => x.CloseConnectionAsync(CancellationToken.None, false), Times.Once);
                this.dataRequestHandlerMock.Verify(x => x.SaveChangesToDeviceAsyncAssert(), Times.Never);
            }
```

Tests/Integration/LoRaDataRequestHandlerConnectionAffinityIntegrationTests.cs
```diff
@@ -64,55 +64,89 @@ public LoRaDataRequestHandlerConnectionAffinityIntegrationTests(ITestOutputHelpe
...
+        [InlineData(true, false)]
+        [InlineData(false, false)]
+        [InlineData(null, true)]
+        public async Task When_Gateway_Does_Not_Own_Connection_It_Should_Delay_If_Delay_Enabled(bool? connectionOwner, bool processingDelayEnabled)
         {
             // arrange
             this.deviceMock.Object.IsConnectionOwner = connectionOwner;
 
+            if (!processingDelayEnabled)
+                ServerConfiguration.ProcessingDelayInMilliseconds = 0;
```
spygi, line 78: [nit] Here we are setting the configuration in order to ensure that `IsProcessingDelayEnabled()` will return the right result. However if we change the implementation (for example so that 0 is considered as delay enabled) then the tests would need to be fixed. Since we have the `dataRequestHandlerMock` shouldn't we maybe use it here as well to mock out the implementation of the `IsProcessingDelayEnabled()`?

Tests/Integration/LoRaDataRequestHandlerConnectionAffinityIntegrationTests.cs
MaggieSalak: Done in 65f0f573b3322e0c3d80db70f4cd35a317973521, thanks for the good suggestion:)

Tests/Integration/LoRaDataRequestHandlerConnectionAffinityIntegrationTests.cs
MaggieSalak: Makes sense, done in 5ff5499bc0c9f0c480246e663cbed29d0c4fe500

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -204,6 +207,22 @@ private bool TryUpdateConfigurationWithDesiredProperties(TwinCollection desiredP
...
+            {
+                return;
+            }
+
+            var reader = new TwinCollectionReader(desiredProperties, this.logger);
+            if (reader.TryRead<int>(Constants.ProcessingDelayKey, out var processingDelay))
+            {
+                this.logger.LogDebug($"Updating processing delay for LNS {this.networkServerConfiguration.GatewayID} " +
+                    $"to {processingDelay} from desired properties of the module twin");
+                this.networkServerConfiguration.ProcessingDelayInMilliseconds = processingDelay;
```
p-schuler, line 222: shouldn't we add range checking here? If the configured delay in the twin is < 0 for example, I would expect to use the default. 

And didn't we discuss that we should use the default, if it is not specified? so they need to opt out? otherwise, they need to take action to get the higher scalability.

p-schuler: or fallback to 0

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -204,6 +207,22 @@ private bool TryUpdateConfigurationWithDesiredProperties(TwinCollection desiredP
...
+        {
+            if (desiredProperties is null)
+            {
+                return;
+            }
+
+            var reader = new TwinCollectionReader(desiredProperties, this.logger);
+            if (reader.TryRead<int>(Constants.ProcessingDelayKey, out var processingDelay))
+            {
+                this.logger.LogDebug($"Updating processing delay for LNS {this.networkServerConfiguration.GatewayID} " +
```
p-schuler, line 220: no need to add the GatewayID, the gatewayId is included in all entries, it's the host name.

p-schuler: can we use structured logging for the delay. That allows to query for the entry when it is for example > 400.

MaggieSalak:
> Maybe out of scope for this PR but shouldn't the delay also be a configuration on azuredeploy/createEdgeDevice.json when we create the devices on IoTHub? And also added on the quickstart documentation.

@p-schuler any comments on this?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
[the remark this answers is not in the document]
MaggieSalak: As discussed offline, merged the two methods into one

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
MaggieSalak: Makes sense, will add a check there. Regarding values <0 I think it's safer to use the default but don't have a strong opinion on it.

MaggieSalak: Done in 6fad78bcfa16585a84159b122f35a6dae0a5e1ad, using default delay if provided value is <0.

spygi (approved):

Issue #1573, opened 2022-03-02
`ObjectDisposedException` & other exceptions require `DeviceClient` recreation


danigian:
Sometimes C# IoT Hub Device Client is, under the hood, disposing some internal components while we could still need those.

Unfortunately, when a situation like this happens, we need to dispose the entire client and recreate a new instance.

We need to handle this scenarios and retry before eventually throwing.

Example stacktrace:

```
System.ObjectDisposedException:
   at Microsoft.Azure.Devices.Client.AuthenticationWithTokenRefresh+<GetTokenAsync>d__19.MoveNext (Microsoft.Azure.Devices.Client, Version=1.39.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at Microsoft.Azure.Devices.Client.Transport.AmqpIot.AmqpIotCbsTokenProvider+<GetTokenAsync>d__3.MoveNext (Microsoft.Azure.Devices.Client, Version=1.39.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
[...]

PR #1574, opened 2022-03-02
Recreate `DeviceClient` & retry operation on some errors

atifaziz:
# PR for issue #1573

## What is being addressed

Under certain conditions, the device client enters into a state where the only recourse is to re-create the client.

## How is this addressed

Whenever execution of a device client operation throws `ObjectDisposedException` or `InvalidOperationException`, the device client is re-created and the operation retried 3 times before giving up.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientExtensions.cs
```diff
@@ -0,0 +1,120 @@
...
+                    try
+                    {
+                        _ = this.client.EnsureConnected();
+                        return await operation(this.client, arg1, arg2, cancellationToken);
+                    }
+                    catch (Exception ex)
+                        when ((ex is ObjectDisposedException
+                                  || (ex is InvalidOperationException ioe
+                                      && ioe.Message.StartsWith("This operation is only allowed using a successfully authenticated context.", StringComparison.OrdinalIgnoreCase)))
+                              && ExceptionFilterUtility.True(() => this.logger?.LogWarning(ex, @"Device client operation ""{Operation}"" failed due to: {Error}", operationName, ex.GetBaseException().Message)))
```
danigian, line 73: Can you please explicitely state here that the operation will be retryied?

atifaziz: Added with 8babf8a84e80d52ebf2bd225cb90d9d4fd882cdd.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientExtensions.cs
```diff
@@ -0,0 +1,125 @@
...
+                    {
+                        _ = this.client.EnsureConnected();
+                        return await operation(this.client, arg1, arg2, cancellationToken);
+                    }
+                    catch (Exception ex)
+                        when ((ex is ObjectDisposedException
+                                  || (ex is InvalidOperationException ioe
+                                      && ioe.Message.StartsWith("This operation is only allowed using a successfully authenticated context.", StringComparison.OrdinalIgnoreCase)))
+                              && ExceptionFilterUtility.True(() =>
+                                     this.logger?.LogWarning(ex, @"Device client operation ""{Operation}"" (attempt {Attempt}/"
```
p-schuler, line 75: Since we are handling the exception, let's not log a warning. As long as we can handle it, we are ok and if we can't (due to the retries) we will log it up in the stack and we know we ran out of retries. 

I was contemplating if a metric of the retries would make sense but I think it does not add much value.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientExtensions.cs
```diff
@@ -0,0 +1,125 @@
...
+                            throw;
+                    }
+                }
+            }
+
+            private delegate Task<TResult> Operation<in T1, in T2, TResult>(ILoRaDeviceClient client, T1 arg1, T2 arg2, CancellationToken cancellationToken);
+
+            private static class Operations
+            {
+                public static readonly Operation<Void, Void, Twin> GetTwin = (client, _, _, cancellationToken) => client.GetTwinAsync(cancellationToken);
```
p-schuler: We had a chat on this one: I felt it was a lot of added complexity to avoid the additional allocations on the closure, but runtime wise, it is a good option. Atif and I agreed, using a static function to avoid capturing any context and avoid additional allocations.

p-schuler (approved):
Nice design with the extension method and wrapping of the client instance.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientExtensions.cs
atifaziz: Addressed with f845f60e700f0eef570879b71c3f951e0f62308e.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClientExtensions.cs
atifaziz: I agree and so with 43b1b882938e30d1eaf06be1572f501505cbd98a, I (logically) reverted to the initial and simpler implementation I had in 37575aa475152e55b172d379cd20b15954ea317f.

PR #1577, opened 2022-03-02
Function notifies losing LNS with a direct method

spygi:
# PR for issue #1497 

## What is being addressed

When the winning LNS changes on the function side, we have to inform the losing LNS proactively that it lost the race. The reason why we do this is in order for its Edge Hub to not retry to create the connection which would result in a connection ping-pong.

## How is this addressed

- A new direct method is called from the `DeduplicationExecutionItem` when a connection switch happens.
- When LNS receives it (in `ModuleConnectionHost`), it sets itself as the losing LNS and closes the connection.
- Unit tests added for LNS (`ModuleConnectionHostTest`) and the function side (`MessageDeduplicationTests`), integration tests for Function (`DeduplicationTestWithRedis`) and a simulation test (`SimulatedLoadTest`) for the whole flow.

## Verification

[Simulation test in the CI](https://github.com/Azure/iotedge-lorawan-starterkit/runs/5412518902?check_suite_focus=true)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs
```diff
@@ -262,6 +262,7 @@ public bool EnsureConnected()
 
         private DeviceClient CreateDeviceClient()
         {
+            this.logger.LogDebug("CreatingDeviceClient");
```
danigian, line 265: We don't need this anymore 👍

Tests/Simulation/SimulatedLoadTests.cs
```diff
@@ -81,6 +82,25 @@ public async Task Single_ABP_Simulated_Device()
             AssertMessageAcknowledgement(device, messageCount);
         }
 
+        [Fact]
+        public async Task Ensures_Disconnect_Happens_For_Losing_Gateway_When_Connection_Switches()
+        {
+            // arrange
+            var device = new SimulatedDevice(TestFixtureSim.Device1003_Simulated_ABP, simulatedBasicsStation: new[] { this.simulatedBasicsStations.First() }, logger: this.logger);
+            await SendConfirmedUpstreamMessages(device, 1);
+
```
danigian, line 91: Let's wait 15 seconds between the two confirmed messages and log something that is helping understanding what's happening there

danigian: In addition, let's send at least 3 messages to both gateways instead of one and let's verify that the connection does not switch before than what we expect

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
```diff
@@ -47,7 +58,8 @@ internal async Task<DuplicateMsgResult> GetDuplicateMessageResultAsync(DevEui de
             {
                 if (await deviceCache.TryToLockAsync())
                 {
-                    // we are owning the lock now
+                    logger?.LogDebug("Obtained the lock to execute deduplication for device '{DevEui}' and frame counter up '{FrameCounterUp}'.", devEUI, clientFCntUp);
```
p-schuler, line 61: I think this is too verbose. I would suggest removing this.

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
```diff
@@ -69,9 +81,48 @@ internal async Task<DuplicateMsgResult> GetDuplicateMessageResultAsync(DevEui de
 
                         if (updateCacheState)
                         {
+                            var previousGateway = cachedDeviceState.GatewayId;
+
                             cachedDeviceState.FCntUp = clientFCntUp;
                             cachedDeviceState.GatewayId = gatewayId;
                             _ = deviceCache.StoreInfo(cachedDeviceState);
+
+                            logger?.LogDebug("Previous connection owner was '{PreviousConnectionOwner}', current message was received from '{Gateway}'", previousGateway, gatewayId);
```
p-schuler, line 90: Also here I'm not sure we want to log this. This means every function invocation to the bundler will now log this, when in debug mode. I would suggest to limit the logging to:
1. connection ownership assignment (first time -> no entry, or no gateway id)
2. connection ownership changes

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
```diff
@@ -69,9 +81,48 @@ internal async Task<DuplicateMsgResult> GetDuplicateMessageResultAsync(DevEui de
...
+                            if (previousGateway != gatewayId)
+                            {
+                                var loraC2DMessage = new LoRaCloudToDeviceMessage()
+                                {
+                                    DevEUI = devEUI,
+                                    Fport = FramePort.AppMin,
+                                    MessageId = Guid.NewGuid().ToString()
+                                };
+
+                                using var scope = logger?.BeginScope(new Dictionary<string, object> { [MessageIdKey] = loraC2DMessage.MessageId });
```
p-schuler, line 100: why do we create a scope here? The scope makes sense for me on a higher level in the stack to track common properties, like the DevEui or the calling LNS id.

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
```diff
@@ -69,9 +81,48 @@ internal async Task<DuplicateMsgResult> GetDuplicateMessageResultAsync(DevEui de
...
+                                logger?.LogDebug("Invoking direct method on LNS '{PreviousConnectionOwner}' to drop connection for device '{DevEUI}'", previousGateway, devEUI);
+
+                                var method = new CloudToDeviceMethod(LoraKeysManagerFacadeConstants.CloudToDeviceDropConnection, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1));
+                                _ = method.SetPayloadJson(JsonConvert.SerializeObject(loraC2DMessage));
+
+                                logger?.LogDebug("Payload constructed with device '{DevEui}' for LNS '{PreviousConnectionOwner}'", loraC2DMessage.DevEUI, previousGateway);
```
p-schuler, line 106: too much logging here as well. Why not simply write a single statement to indicate that the connection owner changed and we sent a message to the previous owner? -> after we invoked.

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
p-schuler: every log statement contains the DevEui. @bastbu was looking into adding logging scopes to the function API. Is this overlapping?

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
```diff
@@ -69,9 +81,48 @@ internal async Task<DuplicateMsgResult> GetDuplicateMessageResultAsync(DevEui de
...
+
+                                try
+                                {
+                                    var res = await this.serviceClient.InvokeDeviceMethodAsync(previousGateway, LoraKeysManagerFacadeConstants.NetworkServerModuleId, method);
```
p-schuler, line 110: I think it would be interesting to have a metric tracking the connection ownership changes.

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
```diff
@@ -69,9 +81,48 @@ internal async Task<DuplicateMsgResult> GetDuplicateMessageResultAsync(DevEui de
...
+                                    logger?.LogDebug("Invoked direct method on LNS '{PreviousConnectionOwner}'; status '{Status}'", previousGateway, res?.Status);
+
+                                    if (res == null || !HttpUtilities.IsSuccessStatusCode(res.Status))
```
p-schuler, line 113: I would expect this to throw if we get back anything but a 2xx? Is this not the case?

p-schuler, line 113: nit: I doubt we need to validate for null here. It should throw, if it can't create a result.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -149,6 +159,47 @@ private async Task<MethodResponse> ClearCacheAsync()
             return new MethodResponse((int)HttpStatusCode.OK);
         }
 
+        private async Task<MethodResponse> DropConnectionAsync(MethodRequest methodRequest)
+        {
+            this.forceClosedConnections.Add(1);
+            ReceivedLoRaCloudToDeviceMessage c2d = null;
+
+            this.logger.LogDebug("Will deserialize '{JsonData}'.", methodRequest.DataAsJson);
```
p-schuler, line 167: too verbose

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -149,6 +159,47 @@ private async Task<MethodResponse> ClearCacheAsync()
...
+            {
+                c2d = JsonSerializer.Deserialize<ReceivedLoRaCloudToDeviceMessage>(methodRequest.DataAsJson, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
+            }
+            catch (Exception ex) when (ex is ArgumentNullException or JsonException)
+            {
+                this.logger.LogError(ex, "Unable to parse Json for direct method '{MethodName}' for device '{DevEui}', message id '{MessageId}'", methodRequest.Name, c2d?.DevEUI, c2d?.MessageId);
+                return new MethodResponse((int)HttpStatusCode.BadRequest);
+            }
+
+            using var cts = methodRequest.ResponseTimeout.HasValue ? new CancellationTokenSource(methodRequest.ResponseTimeout.Value) : null;
```
p-schuler, line 179: no need to create the cts, if the validation fails on the next line. Hence, create it after you validate it has a valid DevEUI

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -26,9 +26,12 @@ public sealed class ModuleConnectionHost : IAsyncDisposable
         private readonly LoRaDeviceAPIServiceBase loRaDeviceAPIService;
         private readonly ILogger<ModuleConnectionHost> logger;
         private readonly Counter<int> unhandledExceptionCount;
+        private readonly Counter<int> forceClosedConnections;
         private ILoraModuleClient loRaModuleClient;
         private readonly ILoRaModuleClientFactory loRaModuleClientFactory;
 
+        public const string DroppedConnectionLog = "Device connection was dropped ";
```
p-schuler, line 33: nit: I would prefer `closed` vs. `dropped` - dropped sounds unexpected.

p-schuler (changes requested):
The logging is too verbose and I think we should add a metric tracking the ownership changes. That would allow an operator to track aggressive connection changes and alert on them.

Generally: When Mark strategy is configured, I think we could end up in a situation, where every message (worst case) does switch the connection. We could end up in a scenario, where we have a lot of direct methods going to the edge to request the closing of the connection. I'm not sure this is a concern really, as we make it clear that Mark should not be used in general, but on only for specific devices.

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
spygi: I created the scope to track all logs associated with that message id but since we will remove all logs but 1, I will remove the scope as well. Btw I don't think every log statement in the function contains the DevEui (at least I don't see it)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
spygi: Done in d489a749d8a517f62eb764a7522091ec36527b1c

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
spygi: Removed in d489a749d8a517f62eb764a7522091ec36527b1c

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
spygi: Done in d489a749d8a517f62eb764a7522091ec36527b1c

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
spygi: removed in d489a749d8a517f62eb764a7522091ec36527b1c

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
spygi: done in d489a749d8a517f62eb764a7522091ec36527b1c

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
spygi: d489a749d8a517f62eb764a7522091ec36527b1c

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
spygi: done in d489a749d8a517f62eb764a7522091ec36527b1c

Tests/Simulation/SimulatedLoadTests.cs
spygi: Done in 46a820a8afc9985f807bd2dae09074846f5218a4

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
```diff
@@ -273,10 +274,79 @@ public async Task OnDirectMethodCall_ClearCache_When_Correct_Should_Work()
...
+            mockedDevice.Verify(x => x.CloseConnectionAsync(It.IsAny<CancellationToken>(), true), Times.Once);
+        }
+
+        public static TheoryData<byte[]> DropConnectionInvalidMessages =>
+            new TheoryData<byte[]>
+                {
+                    null,
+                    Array.Empty<byte>(),
+                    Encoding.UTF8.GetBytes("{{\"DevEui\": null, \"Fport\": 1}}"), // invalid DevEui
+                    Encoding.UTF8.GetBytes("{{\"DevEui\": \"{ new DevEui(0) }\", \"Fport\": 1, \"MessageId\": 123}}") // invalid message id
```
atifaziz, line 308: Did you mean to have an interpolated string here?

```suggestion
                Encoding.UTF8.GetBytes($"{{\"DevEui\": \"{ new DevEui(0) }\", \"Fport\": 1, \"MessageId\": 123}}")); // invalid message id
```

What concerns me further is that the tests are passing nonetheless even though the theory data is completely invalid JSON in the first place. Unfortunately, the reliance on just the status code being `HttpStatusCode.BadRequest` means that once cannot know if the individual cases of invalid `DevEui` or `MessageId` are really being exercised. I would recommend to make the tests break before making them go green and relying on another strategy (either returning something in the response of sniffing the log).

I also think the theory data should be simple JSON strings and the conversion to bytes be done within the test method. The problem with bytes is that it's not easy to discern in the text UI what's being tested:

```
OnDirectMethodCall_DropConnection_Should_Return_Bad_Request_When_Invalid_Message(c2d: [123, 123, 34, 68, 101, ...])
OnDirectMethodCall_DropConnection_Should_Return_Bad_Request_When_Invalid_Message(c2d: [123, 123, 34, 68, 101, ...])
OnDirectMethodCall_DropConnection_Should_Return_Bad_Request_When_Invalid_Message(c2d: [])
OnDirectMethodCall_DropConnection_Should_Return_Bad_Request_When_Invalid_Message(c2d: null)
```

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
```diff
@@ -273,10 +274,79 @@ public async Task OnDirectMethodCall_ClearCache_When_Correct_Should_Work()
...
+
+            // assert
+            loRaDeviceRegistry.VerifyAll();
+            mockedDevice.Verify(x => x.CloseConnectionAsync(It.IsAny<CancellationToken>(), true), Times.Once);
+        }
+
+        public static TheoryData<byte[]> DropConnectionInvalidMessages =>
+            new TheoryData<byte[]>
+                {
+                    null,
```
atifaziz, line 305: Did you mean to test with a _null reference_ or a _JSON null_, as in `Encoding.UTF8.GetBytes("null")` in the latter case?

Tests/Integration/DeduplicationWithRedisIntegrationTests.cs
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Integration
```
atifaziz, line 4: Would be best to enable nullable context for new code files:

```suggestion
#nullable enable

namespace LoRaWan.Tests.Integration
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -149,6 +157,45 @@ private async Task<MethodResponse> ClearCacheAsync()
...
 
+        private async Task<MethodResponse> CloseConnectionAsync(MethodRequest methodRequest)
+        {
+            ReceivedLoRaCloudToDeviceMessage c2d = null;
+
+            try
+            {
+                c2d = JsonSerializer.Deserialize<ReceivedLoRaCloudToDeviceMessage>(methodRequest.DataAsJson, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
+            }
+            catch (Exception ex) when (ex is ArgumentNullException or JsonException)
```
atifaziz, line 168: Don't think we should be mixing the two cases: `ArgumentNullException` means there's no JSON data and `JsonException` means the JSON syntax or data is invalid.

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -68,13 +68,19 @@ public override void Configure(IFunctionsHostBuilder builder)
                 .AddSingleton<IFunctionBundlerExecutionItem, ADRExecutionItem>()
                 .AddSingleton<IFunctionBundlerExecutionItem, PreferredGatewayExecutionItem>()
                 .AddSingleton<LoRaDevAddrCache>();
+
+            if (!string.IsNullOrEmpty(configHandler.AppInsightsInstrumentation))
+            {
+                _ = builder.Services.AddApplicationInsightsTelemetry(configHandler.AppInsightsInstrumentation);
```
p-schuler, line 74: are you sure you need the key here? I think there is an overload that does not take a key and will try to use the default configuration? Did you try that?

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
spygi: Added with b62e4551917b5286664bceb2e455928169fe1355 and a160bad449fd96f3c9baf528c8d01784b105d1b6

LoRaEngine/LoraKeysManagerFacade/FunctionsBundler/DeduplicationExecutionItem.cs
```diff
@@ -69,17 +87,52 @@ internal async Task<DuplicateMsgResult> GetDuplicateMessageResultAsync(DevEui de
...
+                            }
                         }
                     }
                     else
                     {
                         // initialize
                         isDuplicate = false;
                         var state = deviceCache.Initialize(clientFCntUp, clientFCntDown);
-                        logger?.LogDebug("initialized state for {id}:{gwid} = {state}", devEUI, gatewayId, state);
+                        logger?.LogDebug("Connection owner for {DevEui} set to {GatewayId}; state {State}", devEUI, gatewayId, state);
```
p-schuler, line 135: I think Bastian is adding the DevEui to the scopes?

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
spygi: I tried now, it seems to be working actually 👍 will remove the code around getting the key.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
```diff
@@ -149,6 +157,45 @@ private async Task<MethodResponse> ClearCacheAsync()
...
+            using var scope = this.logger.BeginDeviceScope(c2d.DevEUI);
+
+            var loRaDevice = await this.loRaDeviceRegistry.GetDeviceByDevEUIAsync(c2d.DevEUI.Value);
+            if (loRaDevice == null)
+            {
+                this.logger.LogError("Could not retrieve LoRa device; message id '{MessageId}'", c2d.MessageId);
+                return new MethodResponse((int)HttpStatusCode.NotFound);
+            }
+
+            loRaDevice.IsConnectionOwner = false;
```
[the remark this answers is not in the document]
p-schuler: tbh - it means at least one of them is closed - minimizing the effect of the ping pong. I would keep it this way. I think the whole strategy around != Drop probably needs a revisit and think through.

p-schuler (approved):
Approved - pls work with Atif on his comments to close this out.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs
spygi: Changed with 0be1f5b3cda6ab3abbfb0bb1b65006befb7139a5

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
```diff
@@ -273,10 +274,79 @@ public async Task OnDirectMethodCall_ClearCache_When_Correct_Should_Work()
...
+        public async Task OnDirectMethodCall_DropConnection_Should_Work_As_Expected()
+        {
+            // arrange
+            var networkServerConfiguration = new NetworkServerConfiguration();
+            var classCMessageSender = new Mock<IClassCDeviceMessageSender>(MockBehavior.Strict);
+            var loRaDeviceRegistry = new Mock<ILoRaDeviceRegistry>(MockBehavior.Strict);
+            var devEui = new DevEui(0);
+            var mockedDevice = new Mock<LoRaDevice>(null, devEui, null);
+            _ = loRaDeviceRegistry.Setup(x => x.GetDeviceByDevEUIAsync(devEui)).ReturnsAsync(mockedDevice.Object);
+            var c2d = $"{{\"DevEui\": \"{ devEui }\", \"Fport\": 1, \"MessageId\": \"{ Guid.NewGuid() }\"}}";
```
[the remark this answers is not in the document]
spygi: agree 0be1f5b3cda6ab3abbfb0bb1b65006befb7139a5

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
spygi: As discussed, changed with 0be1f5b3cda6ab3abbfb0bb1b65006befb7139a5

Tests/Unit/NetworkServer/ModuleConnectionHostTest.cs
spygi: Good catch, thanks! Fixed with 0be1f5b3cda6ab3abbfb0bb1b65006befb7139a5

Issue #1584, opened 2022-03-03
[Task] Use logging scopes in Facade


bastbu:
We should add logging scopes when applicable on endpoints in the Facade function.

PR #1588, opened 2022-03-03
Use logging scopes in Facade wherever possible

bastbu:
# PR for issue #1584

## What is being addressed

We add logging scopes for Stations, DevAddr and Dev EUIs wherever possible in the Facade function


p-schuler (commented):
I see you added the scopes, what I miss is that we ensure the scope is maintained when exceptions are handled (when clause). Can you do a pass on that as well?

p-schuler (approved):
LGTM - Thanks for fixing this - will help with logs correlation between LNS and the function.

bastbu:
Addressed the instances of the scope logging in exceptions with 6131889b28704e20e17f873360b109fbabbadc63

PR #1603, opened 2022-03-04
Update API version

p-schuler:
Update the API version of the function to reflect on the changes (direct method call)


danigian (approved):
Approving as [e2e tests](https://github.com/Azure/iotedge-lorawan-starterkit/runs/5421440233) were invoking new function successfully

PR #1607, opened 2022-03-06
Create Device Registry interface + Add IoT Hub Registry Manager Impl

kbeaugrand:
## What is being addressed

This PR is a first PR to introduce separate implementation for DeviceRegistry (preparing IoT Central Impl).

## How is this addressed

I created the ``IDeviceRegistryManager`` interface and the ``IoTHubRegistryManager`` for the default implementation.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubRegistryManager.cs
```diff
@@ -0,0 +1,105 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices;
+    using Microsoft.Azure.Devices.Shared;
+
+    public sealed class IoTHubRegistryManager : IDeviceRegistryManager
+    {
+        private readonly RegistryManager instance;
+
+        public static IDeviceRegistryManager From(RegistryManager instance)
```
danigian, line 15: In order to avoid all the CA2000 warning suppressions, you might want to give the task of instantiating the RegistryManager to the IoTHubRegistryManager itself.

As an example, the changes would look like the following

```csharp
public static IDeviceRegistryManager CreateWithProvider(Func<RegistryManager> registryManagerProvider)
{
    return registryManagerProvider == null
        ? throw new ArgumentNullException(nameof(registryManagerProvider))
        : (IDeviceRegistryManager)new IoTHubRegistryManager(registryManagerProvider);
}

private IoTHubRegistryManager(Func<RegistryManager> registryManagerProvider)
{
    this.instance = registryManagerProvider() ?? throw new InvalidOperationException("RegistryManager provider provided a null RegistryManager.");
}
```

and any attempt to instantiate a IoTHubRegistryManager would look like:
```csharp
IoTHubRegistryManager.CreateWithProvider(() => RegistryManager.Create(hostName, new ManagedIdentityCredential()))
```
or
```csharp
IoTHubRegistryManager.CreateWithProvider(() => RegistryManager.CreateFromConnectionString(iotHubConnectionString));
```

@Mandur, wdyt about this alternative approach?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IDeviceRegistryManager.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace LoRaTools
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices;
+    using Microsoft.Azure.Devices.Shared;
+
+    public interface IDeviceRegistryManager : IDisposable
```
danigian, line 12: Let's move the IDisposable to the implementations of the interface

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubRegistryManager.cs
```diff
@@ -0,0 +1,105 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaTools
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices;
+    using Microsoft.Azure.Devices.Shared;
+
+    public sealed class IoTHubRegistryManager : IDeviceRegistryManager
```
danigian, line 11: As per other comment, let's move IDisposable here

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
```diff
@@ -44,7 +45,7 @@ public override void Configure(IFunctionsHostBuilder builder)
 
 #pragma warning disable CA2000 // Dispose objects before losing scope
             // Object is handled by DI container.
-            _ = builder.Services.AddSingleton(RegistryManager.CreateFromConnectionString(iotHubConnectionString));
+            _ = builder.Services.AddSingleton<IDeviceRegistryManager>(IoTHubRegistryManager.From(RegistryManager.CreateFromConnectionString(iotHubConnectionString)));
```
danigian, line 48: If other suggestions are implemented, CA2000 suppression can be removed and this can be inlined with other builder.Services.Add* calls below on row 56 and so on

Tests/Unit/IoTHubRegistryManagerTests.cs
```diff
@@ -0,0 +1,358 @@
...
+        private readonly MockRepository mockRepository;
+        private readonly Mock<RegistryManager> mockRegistryManager;
+
+        public IoTHubRegistryManagerTests()
+        {
+            this.mockRepository = new MockRepository(MockBehavior.Strict);
+            this.mockRegistryManager = this.mockRepository.Create<RegistryManager>();
+        }
+
+        private IDeviceRegistryManager CreateManager()
```
danigian, line 25: This should create a IoTHubRegistryManager, since we are unit testing the class.

Consider making the constructor in IoTHubRegistryManager internal so it can be used from here

Tests/Unit/IoTHubRegistryManagerTests.cs
```diff
@@ -0,0 +1,358 @@
...
+        {
+            return IoTHubRegistryManager.From(this.mockRegistryManager.Object);
+        }
+
+        [Fact]
+        public async Task AddConfigurationAsync()
+        {
+            // Arrange
+            var manager = this.CreateManager();
```
danigian, line 34: All the instances returned from CreateManager are implementing IDisposable, therefore we should change all the occurrencies to:
```suggestion
            using var manager = this.CreateManager();
```
in order to ensure a proper disposal

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubRegistryManager.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        private IoTHubRegistryManager(RegistryManager instance)
+        {
+            this.instance = instance;
+        }
+
+        public Task<Configuration> AddConfigurationAsync(Configuration configuration)
+        {
+            return this.instance.AddConfigurationAsync(configuration);
+        }
```
danigian, line 28: This and all other similar rows can be simplified to
```csharp
public Task<Configuration> AddConfigurationAsync(Configuration configuration)
   => this.instance.AddConfigurationAsync(configuration);
```

danigian (changes requested):

LoRaEngine/LoraKeysManagerFacade/FacadeStartup.cs
kbeaugrand: Tried but CA200 still present on this statement.

Tests/Unit/IoTHubRegistryManagerTests.cs
kbeaugrand: We are here unit testing the ``IoTHubRegistryManager`` class that depends on the ``RegistryManager`` class. So we mock the  ``RegistryManager```?

Tests/Unit/IoTHubRegistryManagerTests.cs
kbeaugrand: done

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubRegistryManager.cs
kbeaugrand: Done

kbeaugrand:
@danigian, 

Thank you for your suggestions. I made the changes you requested.

Tests/Unit/IoTHubRegistryManagerTests.cs
danigian: My point is more that the CreateManager method should actually return a IoTHubRegistryManager and not an IDeviceRegistryManager. I agree with mocking the actual RegistryManager class behind it, as you are already doing.

```csharp
private IoTHubRegistryManager CreateManager() =>
  new IoTHubRegistryManager(() => this.mockRegistryManager.Object);
```

Doing this, you can also validate the proper disposal of the mockRegistryManager, by setting up in the constructor the mock as follows
```csharp
this.mockRegistryManager.Protected().Setup("Dispose", ItExpr.Is<bool>(m => true));
```

In addition, you will need to adjust all the this.CreateManager calls to something like the following for the AddConfigurationAsync test method
```csharp
[Fact]
public async Task AddConfigurationAsync()
{
    // Arrange
    using (var manager = this.CreateManager())
    {
        var configuration = new Configuration("testConfiguration");

        this.mockRegistryManager.Setup(c => c.AddConfigurationAsync(It.Is<Configuration>(x => x == configuration)))
            .ReturnsAsync(configuration);

        // Act
        var result = await manager.AddConfigurationAsync(configuration);

        // Assert
        Assert.Equal(configuration, result);
    }
    this.mockRepository.VerifyAll();
}
```

kbeaugrand: Ok Got it !

kbeaugrand:
@danigian any update ?

danigian (commented):
Thanks for reacting to my comments @kbeaugrand

I will now trigger another full ci run before merging, but otherwise LGTM.

Will approve as soon as the CI is green

danigian (approved):
Full CI passed (except for one known transient error)

Please go ahead and merge! 👍

danigian:
> > Full CI passed (except for one known transient error)
> > Please go ahead and merge! 👍
> 
> Unfortunatly I'm not able to merge while the check didn't pass...

Did it for you! Thanks for the contribution!

PR #1611, opened 2022-03-08
Fix Universal Decoder NPM package lock file version

atifaziz:
## What is being addressed

The Docker files for Universal Decoder uses Node 14:

https://github.com/Azure/iotedge-lorawan-starterkit/blob/3d1c00eb90ac0e5c88c709a199c9e47d69275b38/Samples%2FUniversalDecoder%2FDockerfile.amd64#L1

The NPM version is 6, but since the lock file version is 2, it produces the following warning:

    This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2.
    I'll try to do my best with it!

## How is this addressed

This PR fixes the lock file version to 1 by running it through NPM v6 and saving back the lock file.

---
PS The bulk of the diff is due to the packages section that is gone.


atifaziz:
[As can be seen from the CI run](https://github.com/Azure/iotedge-lorawan-starterkit/runs/5461921982?check_suite_focus=true#step:6:1), the warning no longer shows up:

    npm install
    shell: /usr/bin/bash -e {0}
    npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.3.2 (node_modules/fsevents):
    npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

    added 457 packages from 333 contributors and audited 459 packages in 9.883s

    38 packages are looking for funding
    run `npm fund` for details

    found 0 vulnerabilities

danigian (commented):
What about moving to Node 16 (LTS) and to npm 8?
We could also decide to make the Dockerfile use node:16-bullseye-slim (based on bullseye to match what we have for LNS and LBS modules)

danigian (approved):
Approving this to fix the warning, but please have a look at the other comment for actually creating a story for upgrading the node version and docker images

atifaziz:
> …, but please have a look at the other comment for actually creating a story for upgrading the node version and docker images

Tracked in #1612.

Issue #1612, opened 2022-03-08
[Story] Upgrade Universal Decoder sample to use latest Node LTS


atifaziz:
## Description

The Universal Decoder sample is pinned to Node v14 in the Docker files, which will reach end-of-life in about a year. It should be updated to use the latest LTS sometime before then.

## Acceptance criteria

- [ ] Universal Decoder is updated to use latest LTS of Node

atifaziz:
Re-opening because this was accidentally closed by incorrect association with PR #1613.

Mandur:
As the universal decoder was removed I close the issue

PR #1613, opened 2022-03-08
Build decoders list for static loading in Universal Decoder

atifaziz:
# PR for issue #1612

## What is being addressed

The potential for the Universal Decoder to load and execute arbitrary JavaScript code disguised as a decoder.

## How is this addressed

Instead of globbing for JavaScript files, this PR updates `copy-codecs.js` to build a static list of decoder modules as it copies and patches them from the TTN repo. The static list goes into the `./codecs`/`./codecs/index.js` module and an example of its content would be:

```js
module.exports = {
  "klax": require("./alpha-omega-technology/klax.js"),
  "aqs": require("./aquascope/aqs.js"),
  "bvs": require("./aquascope/bvs.js"),
  "kfr": require("./aquascope/kfr.js"),
  "wwd": require("./aquascope/wwd.js"),
  /* ... */
  "a3": require("./uradmonitor/a3.js"),
  "industrial": require("./uradmonitor/industrial.js"),
  "mousetrap": require("./xignal/mousetrap.js")
};
```

The `decoders.js` module is now loaded statically at run-time via `require('./decoders')` _if it exists_. The `decoders.js` module is ignored for version control since it is dynamically maintained.

Note that some JavaScript never contained the expected `decodeUplink` function and therefore would have failed to load at run-time previously at the following patched line:

```js
module.exports={decodeUplink};
```

`copy-codecs.js` now sniffs top-level function declarations for `decodeUplink` and includes a decoder in the list only if such a declaration is found.

One performance downside is that all decoders are now loaded (which can take 2-3 seconds) even if a tiny subset is used/tested, but performance is not a concern since Universal Decoder is just sample code for illustrating how the decoders from the TTN repo could be used with this kit.

### DIFF
--- a/Samples/UniversalDecoder/tools/copy-codecs.js
+++ b/Samples/UniversalDecoder/tools/copy-codecs.js
@@ -3,12 +3,15 @@
 const glob = require('glob');
 const fs = require('fs');
 const fse = require('fs-extra');
+const path = require('path');
+const esprima = require('esprima');
 
 var args = process.argv.slice(2);
 const srcDir = args[0] || './node_modules/lorawan-devices/vendor';
 const dstDir = args[1] || './codecs';
+const indexFilePath = path.join(dstDir, '../decoders.js');
 
-glob.sync(`**/*`,
+const index = glob.sync(`**/*`,
   {
     cwd: srcDir,
     nodir: true,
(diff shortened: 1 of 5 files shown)


marcgs:
Thanks for the fix @atifaziz ! Good job!

Samples/UniversalDecoder/tools/copy-codecs.js
```diff
@@ -25,9 +28,18 @@ glob.sync(`**/*`,
     fse.copySync(srcPath, dstPath);
 
     if (f.endsWith(".js")) {
-      fs.appendFile(dstPath, '\nmodule.exports={decodeUplink};', function (err) {
-        if (err) throw err;
-        console.log(`Patching ${dstPath}`);
-      });
+      const tree = esprima.parseScript(fs.readFileSync(srcPath).toString());
+      if (tree.body.find(n => n.type === 'FunctionDeclaration' && n.id.name === 'decodeUplink')) {
```
marcgs, line 32: nit: a comment explaining the logic here would help

Samples/UniversalDecoder/tools/copy-codecs.js
```diff
@@ -3,12 +3,15 @@
 const glob = require('glob');
 const fs = require('fs');
 const fse = require('fs-extra');
+const path = require('path');
+const esprima = require('esprima');
 
 var args = process.argv.slice(2);
 const srcDir = args[0] || './node_modules/lorawan-devices/vendor';
 const dstDir = args[1] || './codecs';
+const indexFilePath = path.join(dstDir, '../decoders.js');
```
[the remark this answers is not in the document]
atifaziz: Done with 1cb67e31086c40735983c1b83ae9535310a6bec1.

Samples/UniversalDecoder/tools/copy-codecs.js
atifaziz: Added with 2515958db46883d17e2b936cda4ca3b7a5792b0f.

Issue #1624, opened 2022-03-23
[ADR] Decoupling LoRaWAN Network Server from IoT Edge


danigian:
The ADR should:

- Describe a deployment scenario being enabled by decoupling 
- Describe how to differentiate "edge" LNSs from "cloud" LNSs in mixed scenarios
- Provide additional information on abstracting and replacing ModuleClient related dependencies, i.e.:
    - Direct methods (currently used for Clearing Cache, forcefully closing DeviceClient connections, handling class-c c2d messages)
    - Twin updates (currently used for updating few configuration items, without restarting the entire LNS)
- Provide information on a "Function" endpoint for clearing cache of all "cloud" LNSs

PR #1626, opened 2022-03-25
ADR - Decouple LoRaWan Network Server from IoT Edge

danigian:
# PR for issue #1624

## What is being addressed

This PR is adding an ADR on how to decouple LoRaWan Network Server from IoT Edge.

As a follow-up, some work items will be created for implementing the requested changes


docs/adr/012_decouple_from_edge.md
```diff
@@ -0,0 +1,147 @@
...
+  - Closing DeviceClient connections for avoiding "ping-pong" as described in [ADR 010](./010_lns_affinity.md)
+  - Sending Cloud to Device messages for Class-C devices
+
+Except for ModuleConnectionHost, the other difference is concerning the ability to "proxy" leaf-device messages through Edge Hub. By default the "ENABLE_GATEWAY" environment variable is set to true, we need to make sure that **ENABLE_GATEWAY can't be set to true when not running as Edge module**
+
+### Differentiate Edge from Cloud deployments
+
+As of v2.1.0, the LNS code is already able to understand whether it is running as IoT Edge Module or not. The availability of the "IOTEDGE_APIVERSION" environment variable is checked. This variable is injected by the "Edge Agent" component at the moment of module creation and startup, therefore if the variable is not there we can safely assume that the LNS is not running in a Edge environment.
+
+When a leaf device message is sent and processed, the device twin gets updated with the "GatewayID", currently being set to the "IOTEDGE_DEVICEID" environment variable.
```
p-schuler, line 63: Can we rely on that? What if they remove this or change it? Should we have an explicit variable we provision to indicate the deployment model?

docs/adr/012_decouple_from_edge.md
```diff
@@ -0,0 +1,147 @@
...
+- for sending the LNS a command to close the device client connection for such leaf device
+- for sending a C2D message for a Class-C device through a "direct method invocation"
+
+We need to **conditionally set the "GatewayID" parameter depending on whether the LNS is running on Edge or not**
+
+The proposal is to:
+
+- Prefix the "GatewayID" with "standalone-" for non-Edge deployments
+- Use the "Hostname" environment variable as a default for GatewayID field
+
```
p-schuler, line 76: nit: cloud as a prefix?

docs/adr/012_decouple_from_edge.md
danigian: To be honest, IOTEDGE_APIVERSION has been there since the very beginning of IoT Edge, I am not expecting them to deprecate this soon. At the same time, it's not something that is meant for "public" usage, therefore we might consider adding an additional variable to be used for identifying a "CLOUD_DEPLOYMENT"

docs/adr/012_decouple_from_edge.md
```diff
@@ -0,0 +1,147 @@
...
+Depending on the target cloud service, you might have different ways to manage configuration in a centralized way.
+
+If using Azure Kubernetes Service, a configmap to be applied to the desired cluster might work for you.
+If using Azure App Service, you might just set the environment variables in the configuration section of the application.
+
+While these solutions are great for "static" configuration (i.e.: those values that require a restart for being changed), the LNS is coded in such a way that some configuration might be changed dynamically without having to restart the entire network server.
+
+The proposal is to **make use of "[Azure App Configuration](https://docs.microsoft.com/en-us/azure/azure-app-configuration/overview)"** to centrally manage application settings and keep [dynamic configuration using poll model](https://docs.microsoft.com/en-us/azure/azure-app-configuration/enable-dynamic-configuration-aspnet-core?tabs=core5x)
+
+### Direct method invocation
```
p-schuler, line 92: consider this a stretch - alternative is to have the configuration changes requiring restarts.

docs/adr/012_decouple_from_edge.md
```diff
@@ -0,0 +1,147 @@
...
+
+As previously mentioned, IoT Edge is allowing the functionality of "direct method invocations" giving the possiblity to remotely invoke some functions on the LNS.
+
+When decoupling LNS from Azure IoT Edge, there are two possibilities available for handling the scenarios covered by direct method invocation.
+
+For both mechanisms the "Facade" Azure Function bundled with the Starter Kit, has to differentiate between a "edge" or "cloud" deployment of LNS. The proposal is to **differentiate the mechanism used for remotely invoke methods depending on the "LNS device id"**. This means that an Azure IoT Edge device will still be using the direct method invocations provided by IoT Hub whereas the "cloud" deployments will use a new, different, mechanism.
+
+#### Expose REST endpoints on the LNS
```
p-schuler, line 100: this is repeating the problem from above. I would suggest to create a dedicated section, that talks about how to address the LNS in the 2 different deployment models and then link to that.

docs/adr/012_decouple_from_edge.md
```diff
@@ -0,0 +1,148 @@
...
+
+One alternative might be to expose some REST endpoints on the LNS. This is the easiest approach to take and would simplify the "manual" invocation of such methods, even though this is not in scope of this document.
+
+The scenarios covered by these method invocations are meant to target exactly one desired LNS instance and not any of them.
+
```
p-schuler, line 106: also here, addressing a particular LNS is the overall problem.

docs/adr/012_decouple_from_edge.md
```diff
@@ -0,0 +1,148 @@
...
+This scenario brings in the possibility of using one (or multiple) Azure Kubernetes Services cluster for running one (or multiple) instances of LoRaWAN Network Server.
+
+While keeping in mind that a LoRa Basics Station can be connected only to a single LNS instance at the same time, with such scenario it could be anyways possible to increase high availability or scalability.
+
+Ideally, the LoRa Basics Station could point to a ["vNext" discovery service](./009_discovery.md) that will be able to balance the load across multiple Azure Kubernetes Service clusters, or in an "active-passive" cluster configuration, only point to the "active" one.
+
+In addition, the load of the potentially multiple LNS instances in the same single cluster could be balanced by routing based on the number of "stations" already connected to the server instances (by leveraging the already existing metric being exported in Prometheus format)
+
+## Summary of the required and proposed changes
+
```
p-schuler, line 141: We have a lot of proposals. Let's turn them into decisions and, with pros and cons and elaborate why we make them. Also, add the alternative to have a fixed deployment model pls and it's pros and cons.

docs/adr/012_decouple_from_edge.md
danigian: I will make the easier way of doing it the preferred way and put the Azure App Configuration as an alternative

docs/adr/012_decouple_from_edge.md
```diff
@@ -52,77 +52,105 @@ The main difference comes to "ModuleConnectionHost", which is responsible for:
...
+The decision is to:
+
+- Prefix the **"GatewayID" with "cloud-"** for non-Edge deployments
+- Use the **"Hostname" environment variable as a default** for GatewayID field
+
+An alternative considered is to identify the difference in deployment model by querying IoT Hub for availability of a "IoT Edge" capable device and cache that result.
+
+An advantage of such alternative is the fact that there's no need to use any prefix for GatewayID which would also not tamper the observability part.
+
+A disadvantage, though, comes from the request quota limits on IoT Hub Device Registry operations which could be cause of throttling and instability for other pieces of the LNS itself.
```
p-schuler, line 105: we are querying for all devices and cache them to avoid too many operations. Aren't those already in that list? Couldn't we fetch it from there?

p-schuler: another alternative would be to indicate the deployment model through the API itself. Make it explicit. The parameter could be optional and default to false (edge mode), avoiding breaking existing deployments.

docs/adr/012_decouple_from_edge.md
```diff
@@ -52,77 +52,105 @@ The main difference comes to "ModuleConnectionHost", which is responsible for:
...
+
+#### Direct method invocation
 
 As previously mentioned, IoT Edge is allowing the functionality of "direct method invocations" giving the possiblity to remotely invoke some functions on the LNS.
 
-When decoupling LNS from Azure IoT Edge, there are two possibilities available for handling the scenarios covered by direct method invocation.
+When decoupling LNS from Azure IoT Edge, there are different possibilities for handling the scenarios covered by direct method invocation.
+
+As stated in previous paragraphs, the "Facade" Azure Function has to differentiate between LNS deployment modes depending on the "LNS device id".
```
p-schuler, line 113: don't repeat the solution here, but link to the section that discusses it. Otherwise we have this all over the place and if we do not end up using the device id, we have to change it in multiple places.

docs/adr/012_decouple_from_edge.md
```diff
@@ -52,77 +52,105 @@ The main difference comes to "ModuleConnectionHost", which is responsible for:
...
+The decision is to:
 
-The scenarios covered by these method invocations are meant to target exactly one desired LNS instance and not any of them.
+- Reuse [Azure Cache for Redis](https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-overview) instance that it's already being deployed in the solution as it allows the usage of [Redis Pub/Sub functionality](https://redis.io/docs/manual/pubsub/)
+- As a stretch goal, useful for high scale and testing scenarios, implement a new function for clearing cache of all the LNS subscribed to the topic
 
-As an example, let's assume the scenario where we want to send a cloud-to-device message to a specific class-c leaf device. In this scenario we need to target the exact LNS which is managing the connection with the LoRa Basics Station in reach of the class-c device we want to target.
+An alternative to "Azure Cache for Redis" for the Pub/Sub functionality is to use "Azure Service Bus". While this would allow higher resiliency, the alternative has been discarded to avoid provisioning another additional service.
```
p-schuler, line 124: as long as we abstract the pub/sub implementation, the actual service should be transparent.

danigian: Agree, will clarify that the goal is abstracting it but still choose Redis in this phase

danigian: Done in b539b62d586c7ec203a11ee4490ed7e2511da5d7

docs/adr/012_decouple_from_edge.md
danigian: regarding the second comment, I tried to write down your suggestion with f8b475b5fb543ecf0cb5d8b51a58fa92c3339859
Please feel free to add any advantage/disadvantage directly

p-schuler: I didn't check what exactly we are putting into the cache, could be that we needed to refactor a bit and get the right info into the cache for the LNSs. If we would query a list of the edge devices, we could query that cache entry for the existence of the gw and if it is not in there, we know it's not an edge device. Simply a list of all edge devices that is refreshed would probably be good enough?

p-schuler (approved):
Let's add the modification of the cache (the way we refresh device information in the bg) as an option to consider. If they can be queried the same way (only changed devices) I think it would be very efficient. You can simply check if a gatewayName exists in that list and the list is refreshing in the background.

PR #1653, opened 2022-04-25
Fix CI Install Step

Mandur:
## What is being addressed

The E2E CI currently [fails](https://github.com/Azure/iotedge-lorawan-starterkit/runs/6087248496?check_suite_focus=true) because of missing dependencies.

## How is this addressed

Adding a command with -E to install missing dependencies seems to [fix the issue](https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/2216221423)


Mandur:
CI here https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/2218361698

danigian (approved):

PR #1680, opened 2022-05-10
 Exclude greenme from universal decoder

Mandur:
Currently the Universal decode CI is broken because of a new decoder from greenme that makes usage of Node classes. We are using the [Esprima ](https://esprima.org/) library to ensure that the node script from the decoder is valid javascript and it seems [esprima doesn't support classes as of now](https://github.com/jquery/esprima/issues/1971).

The proposal is to add an exclusion list to skip the file using unsupported attributes for now. The list might be revisited in the future. The current fix was tested locally and seems to fix the issue

### DIFF
--- a/Samples/UniversalDecoder/tools/copy-codecs.js
+++ b/Samples/UniversalDecoder/tools/copy-codecs.js
@@ -11,12 +11,12 @@ const srcDir = args[0] || './node_modules/lorawan-devices/vendor';
 const dstDir = args[1] || './codecs';
 const indexFilePath = path.join(dstDir, 'index.js');
 // cube.js is ignored as esprima doesn't support class parsing https://github.com/jquery/esprima/issues/1971
-const ignoreList = ["greenme/cube.js"];
 const index = glob.sync(`**/*`,
   {
     cwd: srcDir,
     nodir: true,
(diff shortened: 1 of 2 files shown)


Samples/UniversalDecoder/tools/copy-codecs.js
```diff
@@ -10,7 +10,8 @@ var args = process.argv.slice(2);
 const srcDir = args[0] || './node_modules/lorawan-devices/vendor';
 const dstDir = args[1] || './codecs';
 const indexFilePath = path.join(dstDir, 'index.js');
-
+// cube.js is ignored as esprima doesn't support class parsing https://github.com/jquery/esprima/issues/1971
+const ignoreList = ["greenme/cube.js"];
```
atifaziz, line 14: I believe that using the `!(pattern|pattern|pattern)` with [`glob`](https://github.com/isaacs/node-glob) on the following line will give you the same effect.

atifaziz (approved):

Samples/UniversalDecoder/tools/copy-codecs.js
Mandur: Amended thank you :)

Issue #1637, opened 2022-03-28
Extract ClearCacheAsync, CloseConnectionAsync, SendCloudToDeviceMessageAsync from ModuleConnectionHost


danigian:
Acceptance criteria:

- [x] ClearCacheAsync, CloseConnectionAsync, SendCloudToDeviceMessageAsync are not directly implemented as part of ModuleConnectionHost, but extracted into a "CloudControl" class
- [x] The methods should not anymore accept a "MethodRequest" input but, where needed, a ReceivedLoRaCloudToDeviceMessage object
- [x] The methods should all take into account a Cancellation Token
- [x] All the previously required dependencies in ModuleConnectionHost for the methods above are removed
- [x] Tests are properly adjusted

PR #1695, opened 2022-05-16
feat: split module connection host

bastbu:
# PR for issue #1637

## What is being addressed

With this PR we extract logic for operations on a LNS, which is currently tightly coupled to the `ModuleClient` to a separate class. This change allows us to call that logic from either the Module Client connection or via another mechanism

## How is this addressed

The changes are mostly mechanical, tests are refactored a bit (reduced duplication) but no additional coverage is added.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/LnsOperation.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/LnsOperation.cs
@@ -46,7 +46,7 @@ namespace LoRaWan.NetworkServer.BasicsStation
         {
             if (!string.IsNullOrEmpty(json))
             {
-                ReceivedLoRaCloudToDeviceMessage c2d = null;
+                ReceivedLoRaCloudToDeviceMessage c2d;
 
                 try
                 {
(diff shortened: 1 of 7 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/LnsOperation.cs
```diff
@@ -0,0 +1,123 @@
...
+{
+    using System.Diagnostics.Metrics;
+    using System.Net;
+    using System.Text.Json;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using LoRaTools;
+    using Microsoft.Extensions.Logging;
+
+    internal interface ILnsOperation
```
p-schuler, line 14: nit: this represents any operations that can be triggered server side. Maybe we can use a name that represents that better: ILnsRemoteOperation, ILnsRemoteRequest, ILnsRemoteCall?

p-schuler (approved):
Great work ensuring re-usability. Just a small nit on the naming.

danigian (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/LnsOperation.cs
bastbu: Done in a104ae286a968e8188763bd849ddc6315a8b1202

Issue #1694, opened 2022-05-16
Add Edge devices in cache, in order to identify whether target LNS is "edge" or "cloud" based


danigian:
Acceptance criteria:

- [x] Add functionality to get all IoT Edge device to discriminate whether to use Direct Methods or Pub/Sub Functionality
- [x] Unit tests are added

Issue #1634, opened 2022-03-28
Handle alternative path for InvokeDeviceMethodAsync in SendCloudToDeviceMessage.SendMessageViaDirectMethodAsync


danigian:
Acceptance criteria:

- [x] SendCloudToDeviceMessage is handling the scenario covered by SendMessageViaDirectMethodAsync for cloud based devices
- [x] The message published on the IChannelPublisher has to be published on the channel named as the GatewayId (i.e. preferredGatewayID variable)
- [x] The value of the published message should be the json serialization of the already existing LoRaCloudToDeviceMessage
- [x] The new path is covered by at least one unit test in SendCloudToDeviceMessageTest

Issue #1633, opened 2022-03-28
Handle alternative path for InvokeDeviceMethodAsync in DeduplicationExecutionItem.GetDuplicateMessageResultAsync


danigian:
Acceptance criteria:

- [x] GetDuplicateMessageResultAsync in DeduplicationExecutionItem is handling the CloudToDeviceCloseConnection for cloud based devices
- [x] The message published on the IChannelPublisher has to be published on the channel named as GatewayId (i.e. previousGateway variable)
- [x] The value of the published message should be the json serialization of the already existing LoRaCloudToDeviceMessage
- [x] The new path is covered by at least one unit test in MessageDeduplicationTests

PR #1697, opened 2022-05-17
Adding EdgeDeviceGetter and unit tests

danigian:
# PR for issue #1694

## What is being addressed

This PR is adding functionality to get all the IoT Edge devices in Azure Function to discriminate whether to use Direct Methods or Pub/Sub Functionality (#1633, #1634)

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
@@ -74,8 +74,10 @@ namespace LoraKeysManagerFacade
             return isEdgeDevice;
         }
 
+        private static string RedisLnsDeviceCacheKey(string lnsId) => $"lnsInstance-{lnsId}";
+
         private bool MarkDeviceAsNonEdge(string lnsId)
-            => this.cacheStore.ObjectSet(lnsId,
+            => this.cacheStore.ObjectSet(RedisLnsDeviceCacheKey(lnsId),
                                          new DeviceKind(isEdge: false),
                                          TimeSpan.FromDays(1),
                                          onlyIfNotExists: true);
@@ -86,7 +88,7 @@ namespace LoraKeysManagerFacade
             var twins = await GetEdgeDevicesAsync(cancellationToken);
             foreach (var t in twins)
             {
-                _ = this.cacheStore.ObjectSet(t.DeviceId,
+                _ = this.cacheStore.ObjectSet(RedisLnsDeviceCacheKey(t.DeviceId),
                                               new DeviceKind(isEdge: true),
                                               TimeSpan.FromDays(1),
                                               onlyIfNotExists: true);
--- a/LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
@@ -1,0 +1,79 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoraKeysManagerFacade
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using LoRaTools;
+    using Microsoft.Azure.Devices.Shared;
+    using Microsoft.Extensions.Logging;
+
+    internal class EdgeDeviceGetter
+    {
+        private readonly IDeviceRegistryManager registryManager;
+        private readonly ILoRaDeviceCacheStore cacheStore;
+        private readonly ILogger<EdgeDeviceGetter> logger;
+
+        public EdgeDeviceGetter(IDeviceRegistryManager registryManager,
+                                ILoRaDeviceCacheStore cacheStore,
+                                ILogger<EdgeDeviceGetter> logger)
+        {
+            this.registryManager = registryManager;
+            this.cacheStore = cacheStore;
+            this.logger = logger;
+        }
+
+        private async Task<List<Twin>> GetEdgeDevicesAsync()
+        {
+            this.logger.LogDebug("Getting Azure IoT Edge devices");
+            var q = this.registryManager.CreateQuery("SELECT * FROM devices where capabilities.iotEdge = true");
+            var twins = new List<Twin>();
+            do
+            {
+                twins.AddRange(await q.GetNextAsTwinAsync());
+            } while (q.HasMoreResults);
+            return twins;
+        }
+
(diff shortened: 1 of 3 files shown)


LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,98 @@
...
+            var q = this.registryManager.CreateQuery("SELECT * FROM devices where capabilities.iotEdge = true");
+            var twins = new List<Twin>();
+            do
+            {
+                twins.AddRange(await q.GetNextAsTwinAsync());
+            } while (q.HasMoreResults);
+            return twins;
+        }
+
+        internal async Task<bool> IsEdgeDeviceAsync(string lnsId)
```
p-schuler, line 40: We should make sure that we support a `CancellationToken`

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                _ = this.cacheStore.ObjectSet(t.DeviceId,
+                                              new DeviceKind(isEdge: true),
+                                              TimeSpan.FromDays(1),
+                                              onlyIfNotExists: true);
+            }
+        }
+    }
+
+    internal class DeviceKind
```
p-schuler, line 90: Do we need to keep track of `IsEdge`? Looks like the query is only returning items that are edge. Hence, can't we simply store the Lns ids as a list in redis and keep that as a validation that it is an edge device?

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
danigian: Unfortunately the underlying RegistryManager does not support cancellation tokens, would you anyways add a token to be cancelled which is not taking in consideration the lack of RegistryManager?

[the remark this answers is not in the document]
p-schuler: You are right of course!

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+            finally
+            {
+                _ = cacheStore.LockRelease(keyLock, owner);
+            }
+            return isEdgeDevice;
+        }
+
+        private bool MarkDeviceAsNonEdge(string lnsId)
+            => this.cacheStore.ObjectSet(lnsId,
```
p-schuler, line 71: MarkDeviceAsNonEdge does not indicate any conditional execution. However, it is conditional and only will execute if it is not in cache. We are also not evaluating the return value when calling the method. I think the name should reflect the conditional aspect and the caller should actually emit the error if it did not update it, as that would mean, we actually had a race with the lock (potentially releasing it too early).

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,98 @@
...
+                                         TimeSpan.FromDays(1),
+                                         onlyIfNotExists: true);
+
+        private async Task RefreshEdgeDevicesCacheAsync()
+        {
+            this.logger.LogDebug("Refreshing Azure IoT Edge devices cache");
+            var twins = await GetEdgeDevicesAsync();
+            foreach (var t in twins)
+            {
+                _ = this.cacheStore.ObjectSet(t.DeviceId,
```
p-schuler, line 82: I would recommend using a prefix for the key. The cache is shared. Adding a prefix will help avoiding conflicts.

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,98 @@
...
+            do
+            {
+                twins.AddRange(await q.GetNextAsTwinAsync());
+            } while (q.HasMoreResults);
+            return twins;
+        }
+
+        internal async Task<bool> IsEdgeDeviceAsync(string lnsId)
+        {
+            var keyLock = $"{nameof(EdgeDeviceGetter)}-lock";
```
p-schuler, line 42: can be a const.

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,98 @@
...
+            var owner = nameof(EdgeDeviceGetter);
```
p-schuler, line 43: also this can be a const on the type.

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,98 @@
...
+            var isEdgeDevice = false;
+            try
+            {
+                if (await cacheStore.LockTakeAsync(keyLock, owner, TimeSpan.FromSeconds(10)))
+                {
+                    var findInCache = () => this.cacheStore.GetObject<DeviceKind>(lnsId);
+                    if (findInCache() is null)
+                    {
+                        await RefreshEdgeDevicesCacheAsync();
```
p-schuler, line 52: One risk here is that if we have a lot of instances of the lns running, it could be that all of them would cause a cache refresh. We may want to introduce a timestamp of the last refresh to mitigate that risk.

p-schuler (approved):
Looks good! The concern with the throttling on the registry manager should be mitigated.

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
danigian: Done

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
danigian: Logging an error now in case of issues

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
danigian: Done

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
danigian: Done with 281263591b5844d0b0cdb98a04948560e320ddfd
Please check if ok 👍

Issue #1629, opened 2022-03-28
Make ProcessingDelayInMilliseconds configurable from environment variables


danigian:
All the configuration parameters, except the ProcessingDelayInMilliseconds, are currently configurable via environment variables. It is therefore needed to also allow the configuration of such parameter from environment variables in a edge-decoupled scenario.

Acceptance criteria:

- [x] ProcessingDelayInMilliseconds is configurable from environment variables (PROCESSING_DELAY_IN_MS), with a default on DefaultProcessingDelayInMilliseconds (400ms)
- [x] Unit tests are properly adjusted

Issue #1628, opened 2022-03-28
When not running as Edge module, ENABLE_GATEWAY can't be set to true


danigian:
Acceptance criteria:

- [x] Modify NetworkServerConfiguration in such a way that a NotSupportedException is thrown when RunningAsIoTEdgeModule is false and ENABLE_GATEWAY is set to true
- [x] A unit test is covering the new code

Issue #1627, opened 2022-03-28
Introduce CLOUD_DEPLOYMENT environment variable


danigian:
As per ADR 012, we need to identify whether a LNS is running standalone or on a edge device

Acceptance criteria:
- [x] NetworkServerConfiguration.cs is changed to grab "RunningAsIoTEdgeModule" from "CLOUD_DEPLOYMENT" environment variable instead of IOTEDGE_APIVERSION
- [x] A test is making sure that BasicsStationNetworkServer.RunServerAsync does not invoke ModuleConnectionHost.CreateAsync when CLOUD_DEPLOYMENT variable is set to false or not set
- [x] When #1638 is ready, if CLOUD_DEPLOYMENT is set to true, the hosted service needs to be injected instead of the ModuleConnectionHost

Issue #1010, opened 2021-12-07
[Story] Create `NetworkServerConfiguration` from `IConfiguration` instead of environment variables


bastbu:
## Description

With using ASP.NET Core for the LNS we get the configuration extensions, which per default add environment variables to the sources of configuration. Currently, `NetworkServerConfiguration` is still created directly from the environment variables in `CreateFromEnvironmentVariables`. We should rely on `IConfiguration` instead of environment variables directly. This would allow integration tests that run in parallel, without possible interference between environment variables of parallel test instances. 

## Acceptance criteria

- [ ] Create `NetworkServerConfiguration` from `IConfiguration` instead of the environment variables

PR #1698, opened 2022-05-17
OCW2022 Decouple LoRaWAN Starter Kit from IoTEdge

andrewDoing:
# PR for issues #1627 #1628 #1629 

## What is being addressed

- Introduce CLOUD_DEPLOYMENT environment variable
- When not running as Edge module, ENABLE_GATEWAY can't be set to true
- Make ProcessingDelayInMilliseconds configurable from environment variables

## How is this addressed

- Added and modified environment variables with accompanying unit tests.

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
@@ -79,7 +79,6 @@ namespace LoRaWan.NetworkServer.BasicsStation
                         .AddHttpClient()
                         .AddApiClient(NetworkServerConfiguration, ApiVersion.LatestVersion)
                         .AddSingleton(NetworkServerConfiguration)
-                        .AddSingleton<ModuleConnectionHost>()
                         .AddSingleton<ILnsRemoteCall, LnsRemoteCall>()
                         .AddSingleton<ILoRaDeviceFrameCounterUpdateStrategyProvider, LoRaDeviceFrameCounterUpdateStrategyProvider>()
                         .AddSingleton<IDeduplicationStrategyFactory, DeduplicationStrategyFactory>()
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
@@ -79,7 +79,6 @@ namespace LoRaWan.NetworkServer.BasicsStation
                         .AddHttpClient()
                         .AddApiClient(NetworkServerConfiguration, ApiVersion.LatestVersion)
                         .AddSingleton(NetworkServerConfiguration)
-                        .AddSingleton<ModuleConnectionHost>()
                         .AddSingleton<ILoRaDeviceFrameCounterUpdateStrategyProvider, LoRaDeviceFrameCounterUpdateStrategyProvider>()
                         .AddSingleton<IDeduplicationStrategyFactory, DeduplicationStrategyFactory>()
                         .AddSingleton<ILoRaADRStrategyProvider, LoRaADRStrategyProvider>()
@@ -124,6 +123,10 @@ namespace LoRaWan.NetworkServer.BasicsStation
 
             if (NetworkServerConfiguration.ClientCertificateMode is not ClientCertificateMode.NoCertificate)
                 _ = services.AddSingleton<IClientCertificateValidatorService, ClientCertificateValidatorService>();
+            if (NetworkServerConfiguration.RunningAsIoTEdgeModule)
+            {
+                _ = services.AddSingleton<ModuleConnectionHost>();
+            }
         }
 
 #pragma warning disable CA1822 // Mark members as static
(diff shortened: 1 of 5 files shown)


Tests/Unit/NetworkServer/ConfigurationTest.cs
```diff
@@ -39,5 +39,31 @@ public void Should_Setup_Allowed_Dev_Addresses_Correctly(string inputAllowedDevA
...
+        [InlineData("500")]
+        [InlineData("x")]
+        public void ProcessingDelayIsConfigurable(string processing_delay)
+        {
+            var envVariables = new[] { ("PROCESSING_DELAY_IN_MS", processing_delay) };
+
+            try
+            {
+                foreach (var (key, value) in envVariables)
+                    Environment.SetEnvironmentVariable(key, value);
```
bastbu, line 52: Note that this can interfere with other tests running in parallel (xUnit runs unit tests from different test classes in parallel per default) - consider using `IConfiguration` to isolate this and be independent of side effects.

Tests/Unit/NetworkServer/BasicsStation/BasicsStationNetworkServerStartupTests.cs
```diff
@@ -27,5 +30,64 @@ public void All_Dependencies_Are_Registered_Correctly()
...
+
+        [Theory]
+        [InlineData(true, true)]
+        [InlineData(true, false)]
+        [InlineData(false, true)]
+        [InlineData(false, false)]
+        public void EnableGatewayTrue_IoTModuleFalse_IsNotSupported(bool cloud_deployment, bool enable_gateway)
+        {
+            // arrange
+            Environment.SetEnvironmentVariable("CLOUD_DEPLOYMENT", cloud_deployment.ToString());
```
bastbu, line 73: Same comment as above about the side effects of using env variables. In addition, we should clean up the environment variables at the end of the test.

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -144,11 +147,15 @@ public static NetworkServerConfiguration CreateFromEnvironmentVariables()
...
+            config.ProcessingDelayInMilliseconds = envVars.GetEnvVar("PROCESSING_DELAY_IN_MS", Constants.DefaultProcessingDelayInMilliseconds);
+            config.RunningAsIoTEdgeModule = !envVars.GetEnvVar("CLOUD_DEPLOYMENT", false);
             config.IoTHubHostName = envVars.GetEnvVar("IOTEDGE_IOTHUBHOSTNAME", string.Empty);
             config.GatewayHostName = envVars.GetEnvVar("IOTEDGE_GATEWAYHOSTNAME", string.Empty);
-            config.EnableGateway = envVars.GetEnvVar("ENABLE_GATEWAY", config.EnableGateway);
+            config.EnableGateway = envVars.GetEnvVar("ENABLE_GATEWAY", config.EnableGateway) is bool enable_gateway
+                                   && !config.RunningAsIoTEdgeModule
+                                   && enable_gateway
+                                   ? throw new NotSupportedException("ENABLE_GATEWAY cannot be true if RunningAsIoTEdgeModule is false.")
+                                   : enable_gateway;
```
bastbu, line 158: [nit] Consider making the side effect here obvious - some parts of the logic are always true (e.g. `is bool enable_getway` is just used to assign a variable, but as `bool`s are value types it will always "match")

```suggestion
            config.EnableGateway = envVars.GetEnvVar("ENABLE_GATEWAY", true);
            if (!config.RunningAsIoTEdgeModule && config.EnableGateway)
            {
                throw new NotSupportedException("ENABLE_GATEWAY cannot be true if RunningAsIoTEdgeModule is false.");
            }
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -125,6 +125,10 @@ public void ConfigureServices(IServiceCollection services)
 
             if (NetworkServerConfiguration.ClientCertificateMode is not ClientCertificateMode.NoCertificate)
                 _ = services.AddSingleton<IClientCertificateValidatorService, ClientCertificateValidatorService>();
+            if (NetworkServerConfiguration.RunningAsIoTEdgeModule)
+            {
+                _ = services.AddSingleton<ModuleConnectionHost>();
```
bastbu, line 130: On L82 we unconditionally register the `ModuleConnectionHost` (I think this came from a merge conflict). Let's make sure to remove it there.

Tests/Unit/NetworkServer/BasicsStation/BasicsStationNetworkServerStartupTests.cs
```diff
@@ -27,5 +30,64 @@ public void All_Dependencies_Are_Registered_Correctly()
...
+            {
+                Assert.NotNull(result);
+            }
+        }
+
+        [Theory]
+        [InlineData(true, true)]
+        [InlineData(true, false)]
+        [InlineData(false, true)]
+        [InlineData(false, false)]
```
bastbu, line 69:
```suggestion
        [CombinatorialData]
```

Tests/Unit/NetworkServer/BasicsStation/BasicsStationNetworkServerStartupTests.cs
danigian: In addition to this, I would move this test only to [ConfigurationTest.cs](https://github.com/Azure/iotedge-lorawan-starterkit/pull/1698/files#diff-0db7d82aa6361a0cec2585b35c85345c82ceacfc335bc5eba9528d518da19ba9)

It is not affecting startup (injection of different services)

Tests/Unit/NetworkServer/ConfigurationTest.cs
danigian: We can't use IConfiguration as the NetworkServerConfiguration is directly parsing environment variables. Let's keep as @andrewdoing did

bastbu: Linking the story with the required change: #1010

bastbu (approved):
LGTM

danigian (approved):

Issue #1636, opened 2022-03-28
Define IChannelSubscriber interface and provide RedisChannelSubscriber implementation for it


danigian:
Acceptance criteria:

- [x] IChannelSubscriber is defined with a SubscribeAsync method, accepting a "channel" string input and a Action input that defines the handler to be executed for new messages
- [x] RedisChannelSubscriber implementation of IChannelSubscriber is provided
- [x] RedisChannelSubscriber should have a constructor that accepts a ISubscriber from StackExchange.Redis and a ILogger
- [x] New integration tests for RedisChannelSubscriber are created. The tests should make use of existing RedisFixture

PR #1699, opened 2022-05-17
Redis subscriber implementation

bastbu:
# PR for issue #1636

## What is being addressed

With this PR we introduce a component that can subscribe to `LnsRemoteCall`s published on a Redis pub/sub channel.

## How this is handled

We introduce a shared class that represents a remote call, which can be transferred across the wire. In the future, there will be two paths to handle these `LnsRemoteCall`s - via direct methods or via a Redis channel subscription. We use a single channel per LNS for this PR, as Redis does not have a limit on pub/sub channels and this allows us to not need filter logic on the LNS side.

### DIFF
--- a/Tests/Integration/RedisRemoteCallListenerTests.cs
+++ b/Tests/Integration/RedisRemoteCallListenerTests.cs
@@ -1,0 +1,60 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace LoRaWan.Tests.Integration
+{
+    using System;
+    using System.Threading.Tasks;
+    using LoRaWan.NetworkServer;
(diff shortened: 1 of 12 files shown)


Tests/Integration/RedisRemoteCallListenerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        private readonly RedisRemoteCallListener subject;
+
+        public RedisRemoteCallListenerTests(RedisFixture redisFixture)
+        {
+            this.redis = redisFixture.Redis;
+            this.subject = new RedisRemoteCallListener(this.redis);
+        }
+
+        [Fact]
+        public async Task Subscribe_Rceives_Message()
```
danigian, line 27: nit: typo  "receives"

danigian (approved):

Issue #1632, opened 2022-03-28
Define IChannelPublisher and provide a RedisChannelPublisher implementation for it


danigian:
Acceptance criteria:

- [ ] IChannelPublisher is defined with a PublishAsync method, accepting a "channel" string input and a "value" string input
- [ ] RedisChannelPublisher implementation of IChannelPublisher is provided
- [ ] RedisChannelPublisher should have a constructor that accepts a ISubscriber from StackExchange.Redis and a ILogger
- [ ] New integration tests for RedisChannelPublisher are created. The tests should make use of existing RedisFixture

PR #1700, opened 2022-05-17
OCW2022 Decouple LoRaWAN Starter Kit from IoTEdge - 1632

mkcomer:
# PR for issue #1632 

## What is being addressed

Subscriber Pub/Sub Functionality for "cloud" LNS deployments

## How is this addressed

- Added IChannelPublisher and RedisChannelPublisher implementation
- RedisChannelPublisher Tests 
- Leverages LnsRemoteCall for message


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LnsRemoteCall.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
```
danigian, line 1: Putting this file in LoRaTools is the right thing to do.

Could you please remove the other instance we have in LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LnsRemoteCall.cs please?

LoRaEngine/LoraKeysManagerFacade/ RedisChannelPublisher.cs
```diff
@@ -0,0 +1,31 @@
...
+            this.redis = redis;
+            this.logger = logger;
+            this.subscriber = this.redis.GetSubscriber();
+        }
+
+        public async Task PublishAsync(string channel, LnsRemoteCall lnsRemoteCall)
+        {
+            this.logger.LogDebug("Publishing message to channel '{Channel}'.", channel);
+            _ = await this.subscriber.PublishAsync(channel, JsonSerializer.Serialize(lnsRemoteCall));
+        }
```
danigian, line 29:
```suggestion
        private readonly ILogger logger;

        public RedisChannelPublisher(ConnectionMultiplexer redis, ILogger<RedisChannelPublisher> logger)
        {
            this.redis = redis;
            this.logger = logger;
        }

        public async Task PublishAsync(string channel, LnsRemoteCall lnsRemoteCall)
        {
            this.logger.LogDebug("Publishing message to channel '{Channel}'.", channel);
            _ = await this.redis.GetSubscriber().PublishAsync(channel, JsonSerializer.Serialize(lnsRemoteCall));
        }
```

We can simplify this by not storing the "GetSubscriber" result object (see [here](https://stackexchange.github.io/StackExchange.Redis/Basics.html#using-redis-pubsub) it is a "cheap pass-thru object that does not need to be stored")

danigian (commented):
Couple of small changes and we are good to go!

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/LnsRemoteCall.cs
mkcomer: removed!

Tests/Unit/LoRaTools/LnsRemoteCallTests.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
```
bastbu, line 1: There's an identical test file in a different directory, we should remove it from there (similar to removing the `LnsRemoteCall` duplciated file).

bastbu (approved):
LGTM, thanks for the contribution!

PR #1706, opened 2022-05-18
fix: Track unhandled exception

bastbu:
## What is being addressed

We ensure that a currently untracked exception is tracked.


p-schuler (approved):
LGTM! Thanks for fixing this.

Issue #1708, opened 2022-05-18
Flaky device cache unit test


p-schuler:
The LoRaDeviceCacheTest.When_Device_Inactive_It_Is_Removed fails occasionally. 

There is a problem with the ordering of releasing the semaphore and actually removing the device. The semaphore should only be released once we actually removed the device from the cache.

PR #1709, opened 2022-05-18
Bug/1708 flaky lo ra device cache test test

p-schuler:
# PR for issue #1708

## What is being addressed

Fixing flaky device cache tests.

## How is this addressed

- fix ordering of semaphore release and device removal
- refactor timings and increase timeout as it was too aggressive

### DIFF
--- a/Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
+++ b/Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
@@ -471,7 +471,7 @@ namespace LoRaWan.Tests.Unit.NetworkServer
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Tests/Unit/NetworkServer/LoRaDeviceCacheTest.cs
```diff
@@ -474,4 +471,11 @@ protected override ValueTask DisposeAsync(bool dispose)
             }
         }
     }
+        internal static class LoRaDeviceCacheOptionsExtensions
```
bastbu, line 474: [nit] alignment seems off

bastbu (approved):

danigian (approved):

Issue #1712, opened 2022-05-18
[Story] Add Clear Cache endpoint to Azure Function


danigian:
## Description

We should have an Azure Function that when invoked is:

- [x] Getting all Edge Devices and invoking a Direct method to clear cache on those
- [x] Getting all Redis "channels" for pub/sub and publish a clear cache message on those channels

PR #1716, opened 2022-05-19
Adding Clear Cache endpoint to Facade function

danigian:
# PR for issue #1712 

## What is being addressed

This PR is introducing a endpoint for Azure Facade Function for clearing cache of both IoT Edge and non-edge devices.

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
@@ -58,9 +58,10 @@ namespace LoraKeysManagerFacade
         internal async Task ClearLnsCacheInternalAsync(CancellationToken cancellationToken)
         {
             this.logger.LogInformation("Clearing device cache for all edge and Pub/Sub channel based Network Servers.");
+            // Edge device discovery for invoking direct methods
             var edgeDevices = await this.edgeDeviceGetter.ListEdgeDevicesAsync(cancellationToken);
             var tasks = edgeDevices.Select(e => InvokeClearViaDirectMethodAsync(e, cancellationToken)).ToArray();
-
+            // Publishing a single message for all cloud based LNSes
             await PublishClearMessageAsync();
             await Task.WhenAll(tasks);
         }
--- a/LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
@@ -13,7 +13,6 @@ namespace LoraKeysManagerFacade
     using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Mvc;
     using Microsoft.Azure.Devices;
-    using Microsoft.Azure.Devices.Common.Exceptions;
     using Microsoft.Azure.WebJobs;
     using Microsoft.Azure.WebJobs.Extensions.Http;
     using Microsoft.Extensions.Logging;
@@ -60,6 +59,10 @@ namespace LoraKeysManagerFacade
             this.logger.LogInformation("Clearing device cache for all edge and Pub/Sub channel based Network Servers.");
             // Edge device discovery for invoking direct methods
             var edgeDevices = await this.edgeDeviceGetter.ListEdgeDevicesAsync(cancellationToken);
+            if (this.logger.IsEnabled(LogLevel.Debug))
+            {
+                this.logger.LogDebug("Invoking clear cache direct method for following devices: {deviceList}", string.Join(',', edgeDevices));
+            }
             var tasks = edgeDevices.Select(e => InvokeClearViaDirectMethodAsync(e, cancellationToken)).ToArray();
             // Publishing a single message for all cloud based LNSes
             await PublishClearMessageAsync();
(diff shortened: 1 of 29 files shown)


LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
```
bastbu, line 3: Let's make sure to enable nullability on all new files

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
```diff
@@ -0,0 +1,99 @@
...
+            await ClearLnsCacheInternalAsync(cancellationToken);
+
+            return new AcceptedResult();
+        }
+
+        internal async Task ClearLnsCacheInternalAsync(CancellationToken cancellationToken)
+        {
+            this.logger.LogInformation("Clearing device cache for all edge and Pub/Sub channel based Network Servers.");
+            var edgeDevices = await this.edgeDeviceGetter.ListEdgeDevicesAsync(cancellationToken);
+            var tasks = edgeDevices.Select(e => InvokeClearViaDirectMethodAsync(e, cancellationToken)).ToArray();
```
bastbu, line 60: [nit] consider adding a comment that this is for edge devices, while L62 is for cloud LNS.

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
```diff
@@ -0,0 +1,99 @@
...
+    using LoRaTools;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.Azure.Devices;
+    using Microsoft.Azure.Devices.Common.Exceptions;
+    using Microsoft.Azure.WebJobs;
+    using Microsoft.Azure.WebJobs.Extensions.Http;
+    using Microsoft.Extensions.Logging;
+
+    public class ClearLnsCache
```
bastbu, line 19:
```suggestion
    public sealed class ClearLnsCache
```

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
bastbu, line 60: [nit] consider adding some logging here as this is a sideeffectful invocation.

Tests/Unit/LoraKeysManagerFacade/ClearLnsCacheTest.cs
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
```
bastbu, line 3: [nit] Let's enable the nullable annotation context here.

LoRaEngine/LoraKeysManagerFacade/IEdgeDeviceGetter.cs
```diff
@@ -3,11 +3,13 @@
...
 namespace LoraKeysManagerFacade
 {
+    using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
 
     public interface IEdgeDeviceGetter
     {
         Task<bool> IsEdgeDeviceAsync(string lnsId, CancellationToken cancellationToken);
+        Task<IEnumerable<string>> ListEdgeDevicesAsync(CancellationToken cancellationToken);
```
bastbu, line 13: Consider using an `ICollection` to express that this is a finite collection. If it is a streaming response, consider using `IAsyncEnumerable` instead.

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            try
+            {
+                var res = await this.serviceClient.InvokeDeviceMethodAsync(lnsId,
+                                                                           LoraKeysManagerFacadeConstants.NetworkServerModuleId,
+                                                                           new CloudToDeviceMethod(LoraKeysManagerFacadeConstants.ClearCacheMethodName),
+                                                                           cancellationToken);
+                if (HttpUtilities.IsSuccessStatusCode(res.Status))
+                {
+                    this.logger.LogInformation("Cache cleared for {gatewayID} via direct method", lnsId);
```
bastbu, line 82: [nit] and more of a question: I remember that in the past this was flagged by static code analysis - it required that the identifiers for structured logging are starting with a capital case. Do you know if something changed there?

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
```diff
@@ -0,0 +1,99 @@
...
+                }
+
+                this.logger.LogError("Direct method call to {gatewayID} failed with {statusCode}. Response: {response}",
+                                     lnsId,
+                                     res.Status,
+                                     res.GetPayloadAsJson());
+            }
+            catch (DeviceNotFoundException ex)
```
bastbu, line 90: Why hide the exception here instead of relying on the top-level error handling/Function error handling? Wouldn't it be better to let make it visible to the user via the status code that this failed?

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
```diff
@@ -0,0 +1,99 @@
...
+                var res = await this.serviceClient.InvokeDeviceMethodAsync(lnsId,
+                                                                           LoraKeysManagerFacadeConstants.NetworkServerModuleId,
+                                                                           new CloudToDeviceMethod(LoraKeysManagerFacadeConstants.ClearCacheMethodName),
+                                                                           cancellationToken);
+                if (HttpUtilities.IsSuccessStatusCode(res.Status))
+                {
+                    this.logger.LogInformation("Cache cleared for {gatewayID} via direct method", lnsId);
+                }
+
+                this.logger.LogError("Direct method call to {gatewayID} failed with {statusCode}. Response: {response}",
```
bastbu, line 85: Should we throw here as well instead of logging to indicate an internal server error?

LoRaEngine/LoraKeysManagerFacade/LoraKeysManagerFacadeConstants.cs
```diff
@@ -11,6 +11,7 @@ internal static class LoraKeysManagerFacadeConstants
         internal const string TwinProperty_DevAddr = "DevAddr";
         internal const string TwinProperty_NwkSKey = "NwkSKey";
         internal const string NetworkServerModuleId = "LoRaWanNetworkSrvModule";
+        internal const string ClearCacheMethodName = "clearcache";
```
bastbu, line 14: Instead of introducing this, should we consider sharing it between the LNS and the Facade? Otherwise, we should have a test that breaks if we change the value of this constant.

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        internal async Task ClearLnsCacheInternalAsync(CancellationToken cancellationToken)
+        {
+            this.logger.LogInformation("Clearing device cache for all edge and Pub/Sub channel based Network Servers.");
+            var edgeDevices = await this.edgeDeviceGetter.ListEdgeDevicesAsync(cancellationToken);
+            var tasks = edgeDevices.Select(e => InvokeClearViaDirectMethodAsync(e, cancellationToken)).ToArray();
+
+            await PublishClearMessageAsync();
+            await Task.WhenAll(tasks);
```
bastbu, line 63: If the synchronous part of a `InvokeClearViaDirectMethodAsync` fails (currently the first operation is already async, so it's not yet a problem) we do not publish the message on Redis. Consider yielding the tasks initially to make it explicit that they will always attempt to execute no matter what.

LoRaEngine/LoraKeysManagerFacade/ClearLnsCache.cs
danigian: No, I don't. I don't see any error either

bastbu (approved):

LoRaEngine/LoraKeysManagerFacade/LoraKeysManagerFacadeConstants.cs
danigian: Unfortunately the constants are already spread between the two projects. I would keep the refactoring out of this PR

Issue #1553, opened 2022-02-25
Decouple LNS from IoT Edge


p-schuler:
The LoRaWAN starter kit utilizes IoT edge to host and manage the LNS and the decoders. There are multiple advantages for this model:
- deployment can be easily managed through Azure IoT Hub
- processing of message can be kept on prem - at the edge
- communication from the IoT hub to the LNS - direct method
- offline capabilities and network resiliency are built into edge

If a deployment requires local processing using edge is the best option. 

With the switch to use LBS to operate the concentrator, we started decoupling the concentrator from the LNS. This opens up deployment models where the LNS is hosted in the cloud and starts handling messages from different sites/locations. Edge is the wrong technology to put into the cloud - obviously - hence to extend the reach of the starter kit we should decouple the dependency on IoT edge and allow to deploy the same code directly into the cloud and scale it up.

The module client interactions will need to be replaced and abstracted in the current implementation. The interactions with the twins are limited to the API function key - which we can completely get rid off and use managed identity.

[ModuleConnectionHost](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/ModuleConnection/ModuleConnectionHost.cs)

The direct method calls would need to be exposed differently - HTTP calls or decoupling with SB or EventHub - the means of how you can call direct methods, would be different based on the deployment model - however, that's generally hidden from the caller in general through the function exposing APIs.

The deployment needs to be updated to be flexible to not use edge but deploy into a service in the cloud (AppService for example). The endpoint we expose needs to support web sockets (LBS connects to web sockets). 

DoD:
1. Dependency on Edge Hub is abstracted in the code (#1627)
2. Deployment template allows to deploy either using edge hub or a cloud service (#1639)
3. Configuration of the LNS specific values is abstracted away from twins to standard configuration (#1629)
4. Direct method calls are abstracted and the service can receive input / requests for the matching operations (#1630)
5. E2E tests are executed against the new deployment model
6. Any code not edge related is re-used

Tasks: 
- [x] #1624
- [x] Decouple code from module client & IoT twins
- [x] Update deployment model
- [x] Configuration management

bastbu:
We need to think about how we support the standalone discovery service for cloud LNS. Currently we rely on querying the network tags from the Edge Device modules to associate LNS and LBS within the same network. @danigian

[...]

PR #1746, opened 2022-06-29
Running LoRaWAN Network Server decoupled from IoT Edge

danigian:
# PR for issue #1553 

## What is being addressed

This PR is bringing into dev branch all the work related to [Decouple LNS from Edge milestone](https://github.com/Azure/iotedge-lorawan-starterkit/milestone/26)

## Validation:
- E2E Run with temporary changes to only target cloud deployment [here](https://github.com/Azure/iotedge-lorawan-starterkit/runs/7153938469) (the only failing test Test_OTAA_Confirmed_And_Unconfirmed_Message_With_Custom_RX1_DR_Offset_MultiGw is to be fixed by #1770)


LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
```diff
@@ -0,0 +1,125 @@
...
+        public EdgeDeviceGetter(IDeviceRegistryManager registryManager,
+                                ILoRaDeviceCacheStore cacheStore,
+                                ILogger<EdgeDeviceGetter> logger)
+        {
+            this.registryManager = registryManager;
+            this.cacheStore = cacheStore;
+            this.logger = logger;
+        }
+
+#pragma warning disable IDE0060 // Remove unused parameter. Kept here for future improvements of RegistryManager
```
Mandur, line 31: Do we want to have a work item to keep this under radar?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
```diff
@@ -124,6 +126,15 @@ public void ConfigureServices(IServiceCollection services)
 
             if (NetworkServerConfiguration.ClientCertificateMode is not ClientCertificateMode.NoCertificate)
                 _ = services.AddSingleton<IClientCertificateValidatorService, ClientCertificateValidatorService>();
+
+            _ = NetworkServerConfiguration.RunningAsIoTEdgeModule
+                ? services.AddSingleton<ModuleConnectionHost>()
```
Mandur, line 131: nit: should we harmonize the name ModuleControlHost ? (Could be done later)

Mandur (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/BasicsStation/BasicsStationNetworkServerStartup.cs
danigian: Will keep as it is for now 👍

LoRaEngine/LoraKeysManagerFacade/EdgeDeviceGetter.cs
danigian: We should not need a work item here. Ideally if a future version of the RegistryManager will support cancellation tokens, as soon as we upgrade the dependency, the build should fail as we are not propagating the cancellation token (as per https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2016)

Issue #1751, opened 2022-07-04
docs: missing environment variables for cloud only configuration


danigian:
In docs published with #1717 FACADE_SERVER_URL and FACADE_AUTH_CODE environment variables are missing from cloud only environment, but they are mandatory

PR #1717, opened 2022-05-19
Add documentation about cloud-based LNS configuration

bastbu:
# PR for issue #1705

## What is being addressed

Documents how LNS can be run in the cloud.

PR #1752, opened 2022-07-04
docs: adding missing environment variables for cloud only configuration

danigian:
# PR for issue #1751

## What is being addressed

In docs published with #1717 FACADE_SERVER_URL and FACADE_AUTH_CODE environment variables are missing from cloud only environment, but they are mandatory

### DIFF
--- a/docs/user-guide/lns-configuration.md
+++ b/docs/user-guide/lns-configuration.md
@@ -53,8 +53,10 @@ environment variables:
 
 | Environment variable name | Description | Mandatory |
 | --- | --- | --- |
-| CLOUD_DEPLOYMENT | Switch to indicate that the LNS is deployed as a standalone cloud-based instance | true |
-| ENABLE_GATEWAY | Gateway mode is only applicable to IoT Edge-based deployments. Must be set to `false` or unset. | false |
-| IOTHUBHOSTNAME | Host name of the IoT Hub to which the LNS should connect to. | true |
-| HOSTNAME | Identifier of the LNS. | true |
-| REDIS_CONNECTION_STRING | Connection string used to connect to the deployed Redis instance. | true |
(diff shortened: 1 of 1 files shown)


docs/user-guide/lns-configuration.md
```diff
@@ -53,8 +53,10 @@ environment variables:
...
-| ENABLE_GATEWAY | Gateway mode is only applicable to IoT Edge-based deployments. Must be set to `false` or unset. | false |
-| IOTHUBHOSTNAME | Host name of the IoT Hub to which the LNS should connect to. | true |
-| HOSTNAME | Identifier of the LNS. | true |
-| REDIS_CONNECTION_STRING | Connection string used to connect to the deployed Redis instance. | true |
+| CLOUD_DEPLOYMENT | Switch to indicate that the LNS is deployed as a standalone cloud-based instance | Yes (must be set to true) |
+| ENABLE_GATEWAY | Gateway mode is only applicable to IoT Edge-based deployments. Must be set to `false` or unset. | No |
+| IOTHUBHOSTNAME | Host name of the IoT Hub to which the LNS should connect to. | Yes |
+| HOSTNAME | Identifier of the LNS. | Yes |
+| REDIS_CONNECTION_STRING | Connection string used to connect to the deployed Redis instance. | Yes |
+| FACADE_SERVER_URL | Azure Facade function url, e.g. "https://deployedfunction.azurewebsites.net/api" | Yes |
```
bastbu, line 61: Do we really want to duplicate the variables? On L. 49 - 51 we mention that we have some "shared" variables, should we maybe consider extracting them to a "common" section?

bastbu (approved):

docs/user-guide/lns-configuration.md
danigian: I know that it might be redundant and potentially error prone on our side when we are making changes
At the same time, as a user I would probably jump directly to the table and definitely forgot about some shared but mandatory variables.

Let's keep them split for now, we can eventually rewrite the documentation in future versions

Issue #1764, opened 2022-07-04
Component Governance is broken


roel4ez:
The Component Governance pipeline is broken because we are using docker images directly from docker hub (docker.io).

We must:

- [x] Ensure all images which are not available on mcr.microsoft.com are manually uploaded to out *.azurecr.io
- [x] Ensure all images are pulled from either mcr.microsoft.com or our own *.azurecr.io
    - [ ] LBS
    - [ ] UniversalDecoderSample

PR #1766, opened 2022-07-04
Use internal docker images instead of public docker.io

roel4ez:
# PR for issue #1764 

## What is being addressed

Use internal docker images, hosted on either mcs.microsoft.com or *.azurecr.io

## How is this addressed

Use specific references instead of default docker.io

**Note** This PR also deletes CayenneSample, as this is no longer needed.


LoRaEngine/modules/LoRaBasicsStationModule/Dockerfile.amd64
```diff
@@ -1,6 +1,6 @@
 #This docker builds a container for the LoRa Basics station on amd64 architecture
 
-FROM amd64/debian:buster as build
+FROM lorae2etestsacr.azurecr.io/amd64/debian:buster as build
```
danigian, line 3: Instead of hardcoding this, which is not accessible by anyone outside of our team, let's do it differently.

```suggestion
ARG SOURCE_CONTAINER_REGISTRY_ADDRESS=docker.io
FROM $SOURCE_CONTAINER_REGISTRY_ADDRESS/amd64/debian:buster
```

Then when we do the buildx bake in the ci we should override as follows
```bash
docker buildx bake --set *.args.SOURCE_CONTAINER_REGISTRY_ADDRESS=${{ secrets.CI_ACR_NAME }}.azurecr.io
```

danigian:
It would be great to split the PRs in two different ones (so that we can bring back the Cayenne easily if needed, even though I am not expecting that)

LoRaEngine/modules/LoRaBasicsStationModule/Dockerfile.amd64
roel4ez: that's a great idea!

danigian (approved):

roel4ez:
> It would be great to split the PRs in two different ones (so that we can bring back the Cayenne easily if needed, even though I am not expecting that)

see #1768 and #1769

PR #1769, opened 2022-07-04
fix: use internal docker images

roel4ez:
# PR for issue #1764

## What is being addressed

We are using public docker hub images as a source for our images, but should use only vetted images.

## How is this addressed

Use an ARG to set the source of the docker image in the docker files.


roel4ez:
**NOTE**: #1768 should be closed first and merged into this PR before closing.

PR #1772, opened 2022-07-05
fix: e2e-ci pipeline

roel4ez:
# PR for issue #1764

Fixes the e2e CI pipeline to also include the ARG overwrite for building the LBS image


roel4ez:
@danigian The e2e tests failed, as you predicted. However, the Build Docker Images step passes, which is where this change was made.

danigian (approved):

PR #1773, opened 2022-07-05
disable docker detector for a few files.

roel4ez:
# PR for issue #1764

Ignore component governance for LBS and Universal Decoder docker files until we clarify the correct behavior with CG


roel4ez:
> Let's create an item in the backlog to remove these

see #1774 1774

Issue #1770, opened 2022-07-04
Redis cache not properly updated because of IoT Hub query results delay


danigian:
Let's imagine the following scenario. One OTAA device is joining and is in reach of two basic stations (connected each to a different network server). One of the two LNS is processing the join, while providing a DevAddr to the device. The other LNS that lost the race knows nothing about the DevAddr given to such device.

As soon as the device starts sending upstream dataframes it will use the DevAddr and the LNS that previously won the race would have the device in its cache (and process the message), the other LNS would query the Facade Function for getting the needed information (device id and primary key).

The facade function is querying IoT Hub by searching for a device which has in its reported properties the DevAddr received by the LNS. Unfortunately as per [documentation](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-query-language) "Query results can have a few minutes of delay with respect to the latest values in device twins". Therefore the query could return no results and the device is being marked as "not our device".

In order to prevent this kind of issue, the Azure Function has also a SyncDevAddrCache function that is running every 5 minutes and reloading all the IoT Hub devices which had a update in any desired property or in the DevAddr reported property in the last 5 minutes.

Again, the problem is that the results can have few minutes delay and, in some tests, we noticed more than 5 minutes of delay.

## Proposed solution

There are two fixes needed to make sure that everything is reliable.

- [ ] The SyncDevAddrCache should not just get updates for the last 5 minutes, but should query IoT Hub for updates since the most recent "LastUpdated" value of the previous iteration (i.e.: Iteration 1 running at time t+0 is reporting the most recent update for all devices being at t-3, Iteration 2 running at time t+15 should get all updates starting from t-3)
- [ ] As soon as a join request is processed, the JoinRequestMessageHandler should proactively send the DevAddrCacheInfo to Azure Function for persisting it in Redis. Doing this, we should be sure that if a device successfully joined and the processing LNS rebooted immediately after, no data messages are lost if another LNS is in reach as it could get device information from the Azure Function immediately.

PR #1776, opened 2022-07-05
fix: Redis cache not properly updated because of IoT Hub query results delay

danigian:
# PR for issue #1770

## What is being addressed

More insights on reasoning of this PR are in #1770 

## How is this addressed

- [x] The SyncDevAddrCache queries IoT Hub for updates since the most recent "LastUpdated" value of the previous iteration (i.e.: Iteration 1 running at time t+0 is reporting the most recent update for all devices being at t-3, Iteration 2 running at time t+15 should get all updates starting from t-3). Tests have been adjusted to reflect this reality.
- [x] As soon as a join request is processed, the JoinRequestMessageHandler proactively sends (fire and forget) the DevAddrCacheInfo to the Azure Function for persisting it in Redis.
- [x] Not strictly related to the original issue, but related to slow query timings, a62006d0b818322ccb559ecc0f3a393c760a572b is directly querying the device twin instead of relying on IoT Hub queries. During testing the pipeline was also impacted by slowness of iot hub query results propagation.

## Validation
- E2E run with OTAA only [here](https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/2617697654)
- FullCI [here](https://github.com/Azure/iotedge-lorawan-starterkit/runs/7229614668)

### DIFF
--- a/LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
@@ -207,16 +207,17 @@ namespace LoraKeysManagerFacade
         private async Task PerformDeltaReload(IDeviceRegistryManager registryManager)
         {
             // if the value is null (first call), we take updates from one hour before this call
-            var lastUpdate = this.cacheStore.StringGet(LastDeltaUpdateKeyValue) ?? DateTime.UtcNow.AddHours(-1).ToString(LoraKeysManagerFacadeConstants.RoundTripDateTimeStringFormat, CultureInfo.InvariantCulture);
-            var query = $"SELECT * FROM devices where properties.desired.$metadata.$lastUpdated >= '{lastUpdate}' OR properties.reported.$metadata.DevAddr.$lastUpdated >= '{lastUpdate}'";
-            var devAddrCacheInfos = await GetDeviceTwinsFromIotHub(registryManager, query, DateTime.Parse(lastUpdate, CultureInfo.InvariantCulture));
+            var lastUpdate = long.TryParse(this.cacheStore.StringGet(LastDeltaUpdateKeyValue), out var cachedTicks) ? cachedTicks : DateTime.UtcNow.AddHours(-1).Ticks;
+            var lastUpdateDateTime = new DateTime(lastUpdate, DateTimeKind.Utc).ToString(LoraKeysManagerFacadeConstants.RoundTripDateTimeStringFormat, CultureInfo.InvariantCulture);
+            var query = $"SELECT * FROM devices where properties.desired.$metadata.$lastUpdated >= '{lastUpdateDateTime}' OR properties.reported.$metadata.DevAddr.$lastUpdated >= '{lastUpdateDateTime}'";
+            var devAddrCacheInfos = await GetDeviceTwinsFromIotHub(registryManager, query, lastUpdate);
             BulkSaveDevAddrCache(devAddrCacheInfos, false);
         }
 
-        private async Task<List<DevAddrCacheInfo>> GetDeviceTwinsFromIotHub(IDeviceRegistryManager registryManager, string inputQuery, DateTimeOffset? lastDeltaUpdateFromCache)
+        private async Task<List<DevAddrCacheInfo>> GetDeviceTwinsFromIotHub(IDeviceRegistryManager registryManager, string inputQuery, long? lastDeltaUpdateFromCacheTicks)
         {
             var query = registryManager.CreateQuery(inputQuery);
-            var isFullReload = lastDeltaUpdateFromCache is null;
+            var isFullReload = lastDeltaUpdateFromCacheTicks is null;
             var devAddrCacheInfos = new List<DevAddrCacheInfo>();
             while (query.HasMoreResults)
             {
@@ -245,7 +246,7 @@ namespace LoraKeysManagerFacade
                             && twin.Properties.Desired.GetMetadata() is { LastUpdated: { } desiredUpdateTime }
                             && twin.Properties.Reported.GetMetadata() is { LastUpdated: { } reportedUpdateTime })
                         {
-                            lastDeltaUpdateFromCache = new[] { lastDeltaUpdateFromCache, desiredUpdateTime.ToUniversalTime(), reportedUpdateTime.ToUniversalTime() }.Max();
+                            lastDeltaUpdateFromCacheTicks = Math.Max(lastDeltaUpdateFromCacheTicks.Value, Math.Max(desiredUpdateTime.Ticks, reportedUpdateTime.Ticks));
                         }
                     }
                 }
@@ -253,7 +254,7 @@ namespace LoraKeysManagerFacade
 
             if (!isFullReload)
             {
-                _ = this.cacheStore.StringSet(LastDeltaUpdateKeyValue, lastDeltaUpdateFromCache.Value.ToString(LoraKeysManagerFacadeConstants.RoundTripDateTimeStringFormat, CultureInfo.InvariantCulture), TimeSpan.FromDays(1));
+                _ = this.cacheStore.StringSet(LastDeltaUpdateKeyValue, lastDeltaUpdateFromCacheTicks.Value.ToString(CultureInfo.InvariantCulture), TimeSpan.FromDays(1));
             }
 
             return devAddrCacheInfos;
--- a/LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
+++ b/LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
@@ -11,7 +11,11 @@ namespace LoraKeysManagerFacade
     using LoRaTools;
     using LoRaTools.Utils;
     using LoRaWan;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.AspNetCore.Mvc;
     using Microsoft.Azure.Devices.Common.Exceptions;
+    using Microsoft.Azure.WebJobs;
+    using Microsoft.Azure.WebJobs.Extensions.Http;
     using Microsoft.Extensions.Logging;
     using Microsoft.Extensions.Logging.Abstractions;
     using Newtonsoft.Json;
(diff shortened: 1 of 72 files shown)


LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
```diff
@@ -238,10 +240,22 @@ private async Task<List<DevAddrCacheInfo>> GetDeviceTwinsFromIotHub(IDeviceRegis
                             NwkSKey = twin.GetNwkSKey(),
                             LastUpdatedTwins = twin.Properties.Desired.GetLastUpdated()
                         });
+
+                        if (!isFullReload
+                            && twin.Properties.Desired.GetMetadata() is { LastUpdated: { } desiredUpdateTime }
+                            && twin.Properties.Reported.GetMetadata() is { LastUpdated: { } reportedUpdateTime })
+                        {
+                            lastDeltaUpdateFromCache = new[] { lastDeltaUpdateFromCache, desiredUpdateTime.ToUniversalTime(), reportedUpdateTime.ToUniversalTime() }.Max();
```
p-schuler, line 248: I assume the desiredUpdateTime and reportedUpdateTime have to be in UTC.
The problem with this approach is that you keep on allocating a new array for each iteration / device. If we have a few thousand of devices (the first load, without an entry in the lastDeltaUpdateFromCache and there was a bulk insert), this can be adding up. We could do something like:

```suggestion
lastDeltaUpdateFromCacheTicks  = Math.Max(lastDeltaUpdateFromCacheTicks, Math.Max(desiredUpdateTime.Ticks, reportedUpdateTime.Ticks))
```
You keep a new variable (last update in ticks) and in the end, you either store the ticks (preferred) or you convert it to a date: new Date(ticks)

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
```diff
@@ -0,0 +1,56 @@
...
+        private readonly ILogger<LoRaDevAddrCacheFunction> logger;
+
+        public LoRaDevAddrCacheFunction(LoRaDevAddrCache loRaDevAddrCache, ILogger<LoRaDevAddrCacheFunction> logger)
+        {
+            this.loRaDevAddrCache = loRaDevAddrCache;
+            this.logger = logger;
+        }
+
+        [FunctionName("StoreInDevAddrCache")]
+        public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Function, "post", Route = "storeindevaddrcache")] HttpRequest req)
```
p-schuler, line 31: you should be able to use the type as a parameter and let the functions framework handle the deserialization?

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
```diff
@@ -0,0 +1,56 @@
...
+                return new BadRequestObjectResult(ex.Message);
+            }
+
+            var requestBody = await req.ReadAsStringAsync();
+            if (string.IsNullOrEmpty(requestBody))
+            {
+                return new BadRequestObjectResult("missing body");
+            }
+
+            var devAddrCacheInfo = JsonConvert.DeserializeObject<DevAddrCacheInfo>(requestBody);
```
p-schuler, line 48: I'm contemplating if leaking a server specific type (cache info) through the API is a good idea. Usually keeping them separate is a good idea. We could also use the endpoint for something more generic, if we name it something like "DeviceJoinNotification", we could hook in other things into the join process from there. WDYT?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceAPIService.cs
```diff
@@ -259,5 +260,23 @@ public override async Task<HttpContent> FetchStationFirmwareAsync(StationEui eui
         }
 
         private HttpClient CreateClient() => this.httpClientFactory.CreateClient(LoRaApiHttpClient.Name);
+
+        public override async Task StoreDevAddrInCacheAsync(DevAddrCacheInfo devAddrCacheInfo, CancellationToken token)
+        {
+            using var client = CreateClient();
+            var url = BuildUri("StoreInDevAddrCache", new Dictionary<string, string>
```
p-schuler, line 267: nit: make this a const and re-use in the LogError message.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
```diff
@@ -203,18 +206,17 @@ private async Task PerformFullReload(IDeviceRegistryManager registryManager)
...
             BulkSaveDevAddrCache(devAddrCacheInfos, false);
         }
 
-        private async Task<List<DevAddrCacheInfo>> GetDeviceTwinsFromIotHub(IDeviceRegistryManager registryManager, string inputQuery)
+        private async Task<List<DevAddrCacheInfo>> GetDeviceTwinsFromIotHub(IDeviceRegistryManager registryManager, string inputQuery, DateTimeOffset? lastDeltaUpdateFromCache)
         {
             var query = registryManager.CreateQuery(inputQuery);
-            var lastQueryTs = DateTime.UtcNow.AddSeconds(-10); // account for some clock drift
-            _ = this.cacheStore.StringSet(LastDeltaUpdateKeyValue, lastQueryTs.ToString(LoraKeysManagerFacadeConstants.RoundTripDateTimeStringFormat, CultureInfo.InvariantCulture), TimeSpan.FromDays(1));
+            var isFullReload = lastDeltaUpdateFromCache is null;
```
p-schuler, line 219: general question: we know that the changes are propagated delayed. Are the timestamps synchronized when the changes propagate, or when they were made? If it is when they were made, and it could happen that they are not propagated all together, wouldn't there be a chance that we could assume we last updated > than a not yet propagated change?

p-schuler (changes requested):
I didn't see a new integration test or unit test that validates we are calling the remote function after joining. Is this still planned?

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
danigian: Using ticks with commit c502852f

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
danigian: I would not introduce this for consistency with all the other httptrigger functions in the project.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
danigian: I was thinking about making it more generic. DevAddrCacheInfo is now extending IoTHubDeviceInfo class (which is already shared, see #1777), by only adding LastUpdatedTwins.

We could ideally have a more generic DeviceJoinNotification function that is actually only taking IoTHubDeviceInfo as input (as anyways we are not needing the LastUpdatedTwins for now).

Would it be better?

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
danigian: The timestamps I could see are about update time and not propagated time.

As far as I understood some properties in the registry are eventually consistent, therefore it could happen that some changes were not propagated.

It's hard to make a call here, not sure whether this is the optimal solution, but it's definitely better than what we had before.

Any suggestion?

danigian:
@p-schuler you were right about the integration tests, I forgot extending the existing ones.
Commit 89368806 is now covering the API calls

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
p-schuler: you could start using it and create an issue to track moving all to the new way - reducing code is always a good thing.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
p-schuler: Not sure we can consolidate the 2, looks like the required information on the 2 is quite different. The one from the facade only has DevEUI, DevAddr and PrimaryKey. Whereas the one in the LNS also has the GatewayId and the NwkSKey.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
p-schuler: To mitigate, we could add an overlap of let's say 10min to the last used timestamp. That would mean, we will always get the ones that were updated last though. However, with the changes on proactively adding the entry when joining, I think the risk that we will miss an item is very low. Hence, We should document the possibility, but argue that we believe it is a managable risk.

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
danigian: Followed your suggestion with commits f93500a669b418694746f7eb3a9f5cbb5186c4f0 and 30302f608e1a16b5e3ff25600ac460639bfd9a97

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
danigian: Comment added with e9a99ee8a6b364c70bbddf9d6c47dcd2e49cdbfb

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCacheFunction.cs
danigian: Done with eb7ca046237c70daf7f77023a6a7c078469f0056 and created #1781

LoRaEngine/LoraKeysManagerFacade/LoRaDevAddrCache.cs
```diff
@@ -194,27 +197,35 @@ internal async Task PerformNeededSyncs(IDeviceRegistryManager registryManager)
...
+            var devAddrCacheInfos = await GetDeviceTwinsFromIotHub(registryManager, query, null);
             BulkSaveDevAddrCache(devAddrCacheInfos, true);
         }
 
         /// <summary>
         /// Method performing a deltaReload. Typically occur every 5 minutes.
+        ///
+        /// Please be aware that changes to twin are delayed in IoT Hub queries.
+        /// The Delta reload is keeping track of the most recent update and using that timestamp
+        /// as the start date/time for the next iteration.
```
p-schuler, line 209: Thanks.

Issue #265, opened 2020-10-12
Support for Local dev ARM64V8


Mandur:
**Is your feature request related to a problem? Please describe.**
When deploying locally, we should be able to build against ARM64V8

**Describe the solution you'd like**
Similar dev experience as AMD64 or ARM32v7

**Additional context**
At the moment, the local build and push fails with no logs, we would need to add ARM64V8 as list of possible target

PR #1780, opened 2022-07-06
feat: Introducing arm64v8 support for LoRa BasicStation container and LoRaWanNetworkServer

danigian:
# PR for issue #265

## What is being addressed

This PR is introducing Dockerfiles for arm64v8 architectures for both LoRaWan Network Server and BasicsStation

## Validation

- CI passing
- Manually run LBS and LNS containers in a Raspberry Pi 4 (running with 64bit raspbian os)

### DIFF
--- a/LoRaEngine/modules/LoRaBasicsStationModule/Dockerfile.arm64v8
+++ b/LoRaEngine/modules/LoRaBasicsStationModule/Dockerfile.arm64v8
@@ -1,0 +1,47 @@
+#This docker builds a container for the LoRa Basics station on arm32 architecture
+
+# NOTE: Use either docker.io or your own registry address to build the image
+ARG SOURCE_CONTAINER_REGISTRY_ADDRESS=your-registry-address.azurecr.io
+FROM $SOURCE_CONTAINER_REGISTRY_ADDRESS/amd64/debian:buster as build
+RUN apt-get update \
+    && apt-get install -y git apt-utils build-essential gcc-aarch64-linux-gnu
+RUN git clone --branch v2.0.6 --single-branch --depth 1 https://github.com/lorabasics/basicstation.git
+
+# creating links for allowing cross-compilation of arm-linux-gnueabihf binaries
+RUN mkdir -p ~/toolchain-rpi/bin/ ~/toolchain-corecell/bin/ \
+    && ln -s /usr/bin/aarch64-linux-gnu-gcc ~/toolchain-rpi/bin/aarch64-linux-gnu-gcc \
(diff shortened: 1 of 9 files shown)


LoRaEngine/modules/LoRaBasicsStationModule/Dockerfile.arm64v8
```diff
@@ -0,0 +1,47 @@
+#This docker builds a container for the LoRa Basics station on arm32 architecture
```
roel4ez, line 1:
```suggestion
#This docker builds a container for the LoRa Basics station on arm64 architecture
```

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
```diff
@@ -266,8 +266,8 @@ internal async Task ProcessJoinRequestAsync(LoRaRequest request)
 
                 var joinAcceptBytes = loRaPayloadJoinAccept.Serialize(appKey);
 
+                // For join accept messages the RX1DROffsetvalue is ignored as the join accept message carry the settings to the device.
```
roel4ez, line 269: nit: is this relevant?

roel4ez:
Do we need to add something to the docs to explain why there are different architectures and where/how each one should be used?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/JoinRequestMessageHandler.cs
danigian: Without this change I was not able to build locally, therefore I moved the comment before

LoRaEngine/modules/LoRaBasicsStationModule/Dockerfile.arm64v8
danigian: Done with a5d0efce

danigian:
> Do we need to add something to the docs to explain why there are different architectures and where/how each one should be used?

Honestly I am not sure. I think a user should be aware of his hw architecture and what to use. Additionally, we have a section in the documentation covering [this](https://azure.github.io/iotedge-lorawan-starterkit/2.1.0/user-guide/devguide/?h=arm32#setup-concentrator-with-azure-iot-edge)

We might want to expand it with arm64v8 option from the dropdown. Wdyt?

roel4ez (approved):
LGTM

PR #1795, opened 2022-07-21
Remove references to Microsoft.Azure.Devices from IDeviceRegistryManager interface

kbeaugrand:
# PR for remove references to Microsoft.Azure.Devices from IDeviceRegistryManager interface

## What is being addressed

The objective of the PR was to remove all references from IoT Hub in the DeviceRegistryManager interface.

## How is this addressed

- Move Create devices methods to RegistryManager interface and instance
- Remove GetDevice method in profit of GetTwin
- Add the GetDevicePrimaryKey method to the DeviceRegistryManager
- Adapt unit test
- Add unit tests to Registry Manager class


LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
```diff
@@ -194,22 +76,16 @@ async Task<ConfigurationContent> GetConfigurationContentAsync(Uri configLocation
...
-                var edgeGateway = await this.registryManager.GetDeviceAsync(deviceName);
+                var edgeGateway = await this.registryManager.GetTwinAsync(deviceName);
 
                 if (edgeGateway == null)
                 {
                     return PrepareResponse(HttpStatusCode.Conflict);
                 }
 
-                if (deployEndDevice)
+                if (deployEndDevice && !await this.registryManager.DeployEndDevices())
```
Mandur, line 86: I am not sure how idempotent is the operation, anyway you are already doing this operation on line 69. What about setting a local variable defaulting to false to capture the result of the operation?

LoRaEngine/LoraKeysManagerFacade/SearchDeviceByDevEUI.cs
```diff
@@ -52,14 +52,14 @@ private async Task<IActionResult> RunGetDeviceByDevEUI(HttpRequest req)
 
             using var deviceScope = this.logger.BeginDeviceScope(parsedDevEui);
 
-            var device = await this.registryManager.GetDeviceAsync(parsedDevEui.ToString());
-            if (device != null)
+            var primaryKey = await this.registryManager.GetDevicePrimaryKeyAsync(parsedDevEui.ToString());
+            if (primaryKey != null)
```
Mandur, line 56: If the device was not found in the Iot hub wouldn't this throw`?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubRegistryManager.cs
```diff
@@ -4,53 +4,63 @@
...
         {
             this.instance = registryManagerProvider() ?? throw new InvalidOperationException("RegistryManager provider provided a null RegistryManager.");
+            this.httpClientFactory = httpClientFactory;
+            this.logger = logger;
         }
 
-        public Task<Configuration> AddConfigurationAsync(Configuration configuration) => this.instance.AddConfigurationAsync(configuration);
-
-        public Task<Device> AddDeviceAsync(Device edgeGatewayDevice) => this.instance.AddDeviceAsync(edgeGatewayDevice);
+        public async Task<bool> AddDevice(IDeviceTwin twin)
```
Mandur, line 44: As this is an async method could we add the async suffix to the method name?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubRegistryManager.cs
```diff
@@ -4,53 +4,63 @@
...
-
-        public async Task<IDeviceTwin> UpdateTwinAsync(string deviceId, string moduleId, IDeviceTwin deviceTwin, string eTag)
-            => new IoTHubDeviceTwin(await this.instance.UpdateTwinAsync(deviceId, moduleId, deviceTwin.ToIoTHubDeviceTwin(), eTag));
+        public async Task<string> GetDevicePrimaryKeyAsync(string deviceId)
+        {
+            var device = await this.instance.GetDeviceAsync(deviceId);
 
-        public async Task<IDeviceTwin> UpdateTwinAsync(string deviceId, IDeviceTwin twin, string eTag, CancellationToken cancellationToken)
-            => new IoTHubDeviceTwin(await this.instance.UpdateTwinAsync(deviceId, twin.ToIoTHubDeviceTwin(), eTag, cancellationToken));
+            return device.Authentication.SymmetricKey.PrimaryKey;
```
Mandur, line 62: As said in a previous comment, if for some reason the device is null (not found), this might throw

LoRaEngine/LoraKeysManagerFacade/CreateEdgeDevice.cs
kbeaugrand: I made the DeployEndDevices idempotent internally.

LoRaEngine/LoraKeysManagerFacade/SearchDeviceByDevEUI.cs
kbeaugrand: As far I know, the SDK doesn't throw an exception when the device doesn't exist. 
To be sure, I just tested what I say, and I can confirm that the GetDevice method returns a null object if not present in the registry.

image

Maybe the occasion to create a unit test for that case to clarify this behavior in the wrapper implementation ?

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubRegistryManager.cs
kbeaugrand: Same answer, the SDK returns an empty object...

kbeaugrand: But I added a null conditional operator for this case

LoRaEngine/LoraKeysManagerFacade/SearchDeviceByDevEUI.cs
Mandur: As discussed orally, I was not expect the iot hub sdk to throw, rather accessing attributes of a null object. The new changes are fine with me, we could add unit test covering this case to the getdeviceprimarykeyasync that would be neat.

Mandur (approved):

PR #1798, opened 2022-07-23
Add default code analysis rules from vs2022-preview

kbeaugrand:
# PR update editor config file to match VS2022 Preview rules

## What is being addressed

Some new rules are added in VS 2022 Preview.
To still be able to build the solution in this IDE version, we need to update some rules.

## How is this addressed
 
Let defaults settings from VS 2022

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,6 +6,13 @@ root = true
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


danigian:
As the solution still builds fine on VS 17.3 (released a couple days ago) I will close this PR without merging. I would like to avoid adding rules from preview versions of VS that might not be brought into stable versions (like happened with VS 17.3)

Thanks anyways for your work 👍

PR #1808, opened 2022-08-04
Fix local development

Mandur:
# PR to fix local development

## What is being addressed

With the addition of the cloud deployment, the local development experience always starts with the Edge mode. The proposed fix explicitly disable edge when running in debug mode.


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWanNetworkSrvModule/Program.cs
```diff
@@ -32,4 +32,6 @@
 };
 
 var configuration = NetworkServerConfiguration.CreateFromEnvironmentVariables();
+// when running in local debug mode we explicitely disable IoT Edge
+configuration.RunningAsIoTEdgeModule = false;
```
danigian, line 36: I don't get the fix. This is not only applied for local development, but also to any docker image being built no?

I would rather leave the code as is, but add something in the docs for the local development

Mandur: Woops. you are totally right, I was remembering this project was only for local dev and I was obviously wrong. Thanks for the catch.

What do you think to execute this line if we have a debugger attached? This typically should not occur in normal scenarios?

danigian: I would personally prefer to have cleaner code, therefore I would go for completely removing it and just adding a documentation note. As an alternative we might make use of launchsettings.json and appsettings.development.json for any dev related parameter like this

Mandur: I added a new parameters 'LOCAL_DEVELOPMENT' that will disable edge when set to true, If agreed I will also edit the [docs](https://azure.github.io/iotedge-lorawan-starterkit/2.1.0/user-guide/devguide/#debugging-in-visual-studio-outside-of-iot-edge-and-docker)

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -154,7 +156,7 @@ public static NetworkServerConfiguration CreateFromEnvironmentVariables()
             var envVars = new CaseInsensitiveEnvironmentVariables(Environment.GetEnvironmentVariables());
             config.ProcessingDelayInMilliseconds = envVars.GetEnvVar("PROCESSING_DELAY_IN_MS", config.ProcessingDelayInMilliseconds);
             config.RunningAsIoTEdgeModule = !envVars.GetEnvVar("CLOUD_DEPLOYMENT", false);
```
danigian, line 158: We can make this row check both CLOUD_DEPLOYMENT and LOCAL_DEVELOPMENT (if any of them is true, the RunningAsIoTEdgeModule will be false). In this way we don't need to alter the Program.cs

Let's also update the docs (please create an issue) with the steps to eventually start a Redis instance for testing those functionalities that require it

danigian (changes requested):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWanNetworkSrvModule/Program.cs
danigian: Replied in another comment. Let's move the discussion there. It's fine to have local_development but we can simplify code

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWanNetworkSrvModule/Program.cs
```diff
@@ -32,4 +32,5 @@
 };
 
 var configuration = NetworkServerConfiguration.CreateFromEnvironmentVariables();
+
```
danigian, line 35:
```suggestion
```
Let's remove the added whiteline here 👍

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
```diff
@@ -145,6 +145,8 @@ public class NetworkServerConfiguration
         /// </summary>
         public int ProcessingDelayInMilliseconds { get; set; } = Constants.DefaultProcessingDelayInMilliseconds;
 
+        public bool IsLocalDevelopment { get; set; }
```
danigian, line 148: I would say that if we don't use it, we should remove it and eventually add later on.
But it's also fine if you want to keep here and use it in the future

danigian (approved):

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
Mandur: Done

LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/NetworkServerConfiguration.cs
Mandur: Done

PR #1811, opened 2022-08-05
Fix CI conditions

Mandur:
# PR to avoid CI triggering unneeded events

## What is being addressed

Currently any trigger of the E2E CI is going to reset the redis and turn on EFLOW VM. We want to avoid this therefore I add conditions to avoid such operation if they are not deemed necessary


roel4ez (approved):
LGTM!

PR #1822, opened 2022-08-11
Implement IoT Hub Station twin

kbeaugrand:
# PR for implementing Station twin in IoT Hub implementation

## What is being addressed

Implementing the station twin, helps me to decouple LoRa management facade from the IoT Hub device twin.

## How is this addressed

- Create the IStationTwin interface
- Implement new method in Registry manager to retrieve station by its StationEui
- Adapt unit test

PR #1841, opened 2022-08-30
Refactor - Wrapping TwinCollection to ITwinProperties interface

kbeaugrand:
# PR for wrapping TwinCollection to ITwinProperties interface

## What is being addressed

The objective is to decouple the IDevice interface from the ``Microsoft.Azure.Devices`` namespace. This introduce the ITwinProperties that gives access to properties from the twin as usual.

This should be merged after #1822.

## How is this addressed

- [ ] Introduce the ITwinProperties interface
- [ ] Add TwinProperties extensions methods
- [ ] Adapt unit tests

### DIFF
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubDeviceTwin.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubDeviceTwin.cs
@@ -10,13 +10,12 @@ namespace LoRaTools.IoTHubImpl
     {
         internal Twin TwinInstance { get; }
 
-        public TwinProperties Properties => this.TwinInstance.Properties;
+        public ITwinPropertiesContainer Properties { get; }
 
         public IoTHubDeviceTwin(Twin twin)
         {
-            ArgumentNullException.ThrowIfNull(twin, nameof(twin));
-
             this.TwinInstance = twin;
+            this.Properties = new IoTHubTwinPropertiesContainer(twin);
         }
 
         public string ETag => this.TwinInstance.ETag;
--- a/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubDeviceTwin.cs
+++ b/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubDeviceTwin.cs
@@ -14,6 +14,8 @@ namespace LoRaTools.IoTHubImpl
 
         public IoTHubDeviceTwin(Twin twin)
         {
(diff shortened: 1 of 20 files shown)


LoRaEngine/modules/LoRaWanNetworkSrvModule/LoraTools/IoTHubImpl/IoTHubDeviceTwin.cs
```diff
@@ -10,13 +10,12 @@ public class IoTHubDeviceTwin : IDeviceTwin
     {
         internal Twin TwinInstance { get; }
 
-        public TwinProperties Properties => this.TwinInstance.Properties;
+        public ITwinPropertiesContainer Properties { get; }
 
         public IoTHubDeviceTwin(Twin twin)
         {
-            ArgumentNullException.ThrowIfNull(twin, nameof(twin));
```
Mandur, line 17: Is there a reason to remove the null check ? Is it performed somewhere else?

kbeaugrand: Probably not, I re-added this check.

kbeaugrand:
@mandur, as discussed I created the #2071 issue to take care of the dynamic binding that was introduced in this PR.

Tests/Common/IntegrationTestFixtureBase.cs
```diff
@@ -335,7 +335,7 @@ public async Task UpdateExistingConcentratorCrcValues(StationEui stationEui, uin
             if (getDeviceResult == null)
                 throw new InvalidOperationException("Concentrator should exist in IoT Hub");
             var deviceTwin = await registryManager.GetTwinAsync(stationDeviceId);
-            var cupsJson = ((object)deviceTwin.Properties.Desired[BasicsStationConfigurationService.CupsPropertyName]).ToString();
+            var cupsJson = (string)(deviceTwin.Properties.Desired[BasicsStationConfigurationService.CupsPropertyName]).ToString();
```
Mandur: Is this cast needed?

Tests/Common/IntegrationTestFixtureBase.cs
```diff
@@ -354,7 +354,7 @@ public async Task UpdateExistingFirmwareUpgradeValues(StationEui stationEui, uin
...
```
Mandur: Is this cast needed?

kbeaugrand: Cast is needed because at build time, this is a dynamic object.

Tests/Common/IntegrationTestFixtureBase.cs
kbeaugrand: Cast is needed because at build time, this is a dynamic object.

Mandur (approved):

PR #1849, opened 2022-09-01
Fix the local development again

Mandur:
The local development experience was only partially fixed by #1808. There is still one condition missing that wants the local development to connect to the redis. This PR aims to finally fix the issue


danigian (approved):
As a quick fix is fine. Maybe we should add an item in the backlog for setting up a redis instance for local development, rather than running in "iot edge" mode

PR #1908, opened 2022-10-21
Device provisioning in Bicep

fbeltrao:
# PR for issue #1893 

## What is being addressed

Creating iot devices using the Bicep infrastructure. 

Devices:
- Gateway
- Concentrator
- Leaf devices (if enabled)

## How is this addressed

Implemented in Bicep templates. Still missing:
- Remove functionality in Function that deploys devices

### DIFF
--- a/TemplateBicep/main.bicep
+++ b/TemplateBicep/main.bicep
@@ -103,3 +103,20 @@ module discoveryService './discoveryService.bicep' = if (useDiscoveryService) {
   }
 }
 
+module createDevices 'devices.bicep' = {
+  name: 'createDevices'
+  params: {
+    location: location
+    deployDevice: deployDevice
+    logAnalyticsName: observability.outputs.logAnalyticsName
+    functionAppName: function.outputs.functionName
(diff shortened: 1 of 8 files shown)


TechPreacher (approved):

Mandur (approved):

TemplateBicep/main.bicep
```diff
@@ -103,3 +103,20 @@ module discoveryService './discoveryService.bicep' = if (useDiscoveryService) {
   }
 }
 
+module createDevices 'devices.bicep' = {
```
Mandur, line 106: I think this shoudl be conditional

fbeltrao: I thought so too :)
This is what I found out:
- The function is called nevertheless (no condition on the function launcher)
image

- The launcher also does not have a condition:
image

- The condition is made available as environment variable `DEPLOY_DEVICE` in the function. It only controls if leaf devices are created:
image

fbeltrao: As discussed, we will keep the current behavior

Issue #1927, opened 2022-10-26
[Docs] Document fix for rasperry pi with reset pin = 7


ouphi:
Story: #XX

**Description**
We experienced that the concentrator does not start properly when the reset pin is 7. This is related to an update of the Rasperry Pis kernel. The issue and the fix is documented here: https://github.com/Lora-net/lora_gateway/issues/162 .

We want to document the instructions to avoid this issue in the Developer guidance section.

(It is not clear to us if this is happening only when the pin is 7 or when it is smaller than 25. Part of the task would be to figure it out before documenting it.)

ouphi:
https://github.com/Azure/iotedge-lorawan-starterkit/pull/1933

PR #1933, opened 2022-10-27
Add troubleshooting doc about reset pin with spi enabled.

ouphi:
# PR for issue #1927 

## What is being addressed

Document issue encountered:
- Reset pin was already in used because it was the `SPI0 CE1` pin by default.

## How is this addressed
Documented how to fix the issue.

### DIFF
--- a/docs/user-guide/troubleshooting.md
+++ b/docs/user-guide/troubleshooting.md
@@ -12,13 +12,11 @@ ral_config failed with status 0x08
 Closing connection to muxs - error in s2e_onMsg
 ```
 
-It could be because SPI is enabled on the reset pin, then the reset pin cannot be used.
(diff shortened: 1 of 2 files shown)


docs/user-guide/troubleshooting.md
```diff
@@ -0,0 +1,23 @@
+# Troubleshooting
```
fbeltrao, line 1: Nice addition. How will people find this page? Do we need to link it from the main page?

fbeltrao (approved):

docs/user-guide/troubleshooting.md
ouphi: Good catch, it was actually missing.
In the `mkdocs.yaml` I added `- user-guide/troubleshooting.md` to the `User guide` section to make it work.

danigian:
Good addition @ouphi ! Thanks for this!

Issue #1885, opened 2022-10-11
Migrate the CI to the new Azure environment


Mandur:
# Description

As per guidance, the E2E CI needs to be moved to the new Azure subscription (fdpo) before the end of the month otherwise it will be deleted.

# Acceptance Criteria

- [x] The infrastructure is deployed on the new tenant
- [x] An [Azure Eflow VM](https://learn.microsoft.com/windows/iot/iot-enterprise/azure-iot-edge-for-linux-on-windows) is provisioned and connection to IoT Hub
  - Eflow is provisioned with a Static IP exposed by a Nat to the private network
  - Eflow should be RDP accessible, either by Bastion or by using JIT controls (prefer the least expensive)
- [x] An [Azure VPN gateway](https://learn.microsoft.com/azure/vpn-gateway/vpn-gateway-about-vpngateways) (VpnGw1) is provisioned, and a certificate chain built
  - The local environment (ItestUp device) can connect to the Azure VPN automatically
- [x] #1898
- [x] #1899
- [x] #1900
- [x] #1947
- [x] #1901
- [REMOVED] #1902

# Resources

Current CI description https://devblogs.microsoft.com/cse/2022/03/15/e2e-tests-for-lorawan-starter-kit-with-real-hardware/

PR #1934, opened 2022-10-27
Feature/migrate ci to new tenant

noraabiakar:
# PR for issue #1885 

## What is being addressed

Migrate the CI to the new Azure subscription (fdpo). In the process, switch to using Azure KeyVault for storing secrets instead of 
GitHub secrets. 

## How is this addressed

- Add a new GitHub action that, given a list of keys, will read the secrets from a KeyVault (in parallel) and expose the values as masked GitHub environment variables.
- Make the necessary changes to the e2e CI to work with the new way of reading secrets. 
- Address other small issues encountered while making the above changes.


.github/workflows/e2e-ci.yaml
```diff
@@ -9,24 +9,27 @@ on: # rebuild any PRs and main branch changes
...
+        required: false
       TestsToRun:
         default: '[SensorDecodingTest,OTAAJoinTest,ABPTest,OTAATest,MacTest,ClassCTest,C2DMessageTest,MultiGatewayTest,MultiConcentratorTest,CupsTest,LnsDiscoveryTest,CloudDeploymentTest]'
         description: 'tests to run'
+        required: false
       TxPower:
         description: 'TXPower value to use in E2E tests'
         default: '6'
+        required: false
+permissions:
```
Mandur, line 26: Why is this needed?

noraabiakar: It is needed for OIDC authentication https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers#adding-permissions-settings

Mandur (approved):

Issue #1950, opened 2022-10-31
[Story] Create pull requests from release workflow


ouphi:
## Description

Add the following functionality to the prerelease workflow:
- Create a PR updating the URL button version in the doc (quickstart.md#49), on the target branch `docs/main`.
- Create a PR updating the default bicep version (templateBicep/main.bicep#40), on the target branch `dev`

This is a follow up of this [story](https://github.com/Azure/iotedge-lorawan-starterkit/issues/1928).

## Acceptance criteria

- [x] When a release is triggered, 2 pull requests are created:
  - [x] One pull request to update the version in bicep
  - [x] One pull request to update the version of the button URL in the doc 

## Tasks
- [x] Create a custom action to create pull requests
- [x] Update the bicep file & use the custom action to create a PR in workflow
- [x] Update the button URL & use the custom action to create a PR in workflow

PR #1953, opened 2022-10-31
Release workflow creates pull requests to update versions in files.

ouphi:
# PR for issue #1950 

## What is being addressed

- Automated the following release steps:
  - Create a PR updating the URL button version in the doc, on the target branch docs/main.
  - Create a PR updating the default starter kit version in bicep, on the target branch dev

- Make sure the ARM release artefact has the correct default started kit version.

I considered creating a custom action to reduce the code duplication. I didn't implement this solution because of the following overhead:
- State: The custom action would depend on a state: the current checked out branch. This can be confusing IMO.
- A custom action can only be used it the action file is present in the branch. The custom action is not present in the `docs/main` branch. We could consider committing it to this branch, but it would mean maintaining this action twice.

## How it was tested

It was tested on a fork to avoid creating PR on this repository.
You can find [here a successful run](https://github.com/ouphi/iotedge-lorawan-starterkit/actions/runs/3363155234).
You can find [here the release created by the run](https://github.com/ouphi/iotedge-lorawan-starterkit/releases/tag/v3.3.3).

You can see the following 2 PRs have been created by the workflow:
- [Update bicep button PR](https://github.com/ouphi/iotedge-lorawan-starterkit/pull/14)
- [Update default version in bicep](https://github.com/ouphi/iotedge-lorawan-starterkit/pull/15)

### DIFF
--- a/.github/workflows/create_prerelease.yaml
+++ b/.github/workflows/create_prerelease.yaml
@@ -125,6 +125,8 @@ jobs:
           zip -r $DISCOVERY_SERVICE_ZIP $PUBLISH_FOLDER
           # Define DISCOVERY_SERVICE_ZIP env var for next steps.
           echo "DISCOVERY_SERVICE_ZIP=$DISCOVERY_SERVICE_ZIP" >> $GITHUB_ENV
+      - name: Update the default LoRaWAN Starter Kit version to use in Bicep. 
+          run: sed -i "s/param version string = '[0-9/.]*'/param version string = '${VERSION}'/g" TemplateBicep/main.bicep
       - name: Generate ARM file
         working-directory: TemplateBicep
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/create_prerelease.yaml
```diff
@@ -41,7 +41,65 @@ jobs:
...
+      - name: Update doc URL with new version
+        run: sed -r -i "s/iotedge-lorawan-starterkit%2Fv[0-9\.]+/iotedge-lorawan-starterkit%2Fv${VERSION}/g" docs/quickstart.md
+      - name: Set up git user using the built-in token
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+      - name: Create branch
+        run: |
+          BRANCH_NAME="docs/release-${VERSION}-${{ github.run_id }}"
+          DESCRIPTION="Update button URL with version $VERSION."
```
Mandur, line 61: Why do you need to put those in the github environment? Isn't that enough to have them as bash var?

Mandur (approved):

.github/workflows/create_prerelease.yaml
ouphi: Because I reuse the var in the next step.

PR #1957, opened 2022-11-01
Fix the Redis test bugs in integration tests

### LINKED ISSUE #1936: Failed integration test, possible race condition
[This failing CI](https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/3338737316/jobs/5526780625) run indicates a race condition as caught by @Mandur

Mandur:
We saw multiple cases of integration tests failing because the Redis reference was missing e.g., https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/3338737316/jobs/5526780625.

This is coming from the fact that we have a code ensuring the Redis container started properly but we are not actively retrying to check if the check fails the first time. This PR aims at fix that and retry on 10x each seconds to verify whether Redis started.

You can see the retry happening in the tests run that occured on this branch https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/3368478002/jobs/5587065674#step:12:143

Additionally update the Redis tested version to 6, to be in line with the Version deployed on Azure.


Tests/Integration/RedisFixture.cs
```diff
@@ -44,7 +45,7 @@ private async Task StartRedisContainer()
 
             try
             {
-                
+
```
danigian, line 48: Remove newline

.github/workflows/ci.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
 
     # Pull Redis Cache docker image
     - name: Pulling Redis Cache image
-      run: docker pull redis:5.0.4-alpine
+      run: docker pull redis:6.0-alpine
```
danigian, line 73: Let's also cleanup the unused image from the CI machine to save space

Tests/Integration/RedisFixture.cs
```diff
@@ -124,17 +125,32 @@ public async Task InitializeAsync()
...
-                Database = this.Redis.GetDatabase();
+                try
+                {
+                    Redis = await ConnectionMultiplexer.ConnectAsync(redisConnectionString);
+                    Database = this.Redis.GetDatabase();
+                    isRedisReady = true;
+                }
+                catch (Exception ex)
+                {
+                    await Task.Delay(1000);
```
danigian, line 141: nit: let's increase the delay and reduce the amount of retries

Tests/Integration/RedisFixture.cs
```diff
@@ -124,17 +125,32 @@ public async Task InitializeAsync()
...
+                    Console.WriteLine($"Failed to connect to redis at '{redisConnectionString}'. If running locally with docker: run 'docker run -d -p 6379:6379 redis'. If running in Azure DevOps: run redis in docker.", ex);
+                }
+
+                tryRun++;
```
danigian, line 145: nit: we can put the tryRun++ inside the catch block, to not have it executed when try is succeeding

Tests/Integration/RedisFixture.cs
```diff
@@ -124,17 +125,32 @@ public async Task InitializeAsync()
...
             {
-                this.Redis = await ConnectionMultiplexer.ConnectAsync(redisConnectionString);
-                Database = this.Redis.GetDatabase();
+                try
+                {
+                    Redis = await ConnectionMultiplexer.ConnectAsync(redisConnectionString);
+                    Database = this.Redis.GetDatabase();
+                    isRedisReady = true;
+                }
+                catch (Exception ex)
```
danigian, line 139: nit: is there a way to catch a more specific exception? ideally we should retry only if we could not connect, but not if we could not get the database, no?

.github/workflows/ci.yaml
Mandur: This is happening on the Github Hosted agents afaik. Are you suggesting we should clean those?

Tests/Integration/RedisFixture.cs
Mandur: Good catch thanks

.github/workflows/ci.yaml
danigian: You are right, I thought the pipeline was running on our local hardware. Sorry, ignore 👍

Tests/Integration/RedisFixture.cs
Mandur: Yes I was lazy here, let me try to make this better

danigian (approved):

Mandur:
The new code seems to still fix the issue correctly https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/3370350839/jobs/5591149563#step:12:149

PR #1958, opened 2022-11-01
Clean up EFLOW VM scripts and add readme

noraabiakar:
# PR for issue #1900


Tools/Eflow/install-hyperv.ps1
```diff
@@ -0,0 +1,15 @@
...
+.NOTES
+     Author     :
+        Mikhail Chatillon       - chmikhai@microsoft.com
+        Daniele Antonio Maggio  - daniele.maggio@microsoft.com
+#>
+
+# Install Hyper-V
+Install-WindowsFeature -Name Hyper-V
+Install-WindowsFeature -Name Hyper-V-PowerShell
+Enable-WindowsOptionalFeature -Online -FeatureName VirtualMachinePlatform -All -NoRestart
```
danigian, line 15: Please add newline at the end of the file

Tools/Eflow/provision-eflow.ps1
```diff
@@ -1,109 +1,85 @@
 <# .SYNOPSIS
     EFLOW Windows Server provisioning Script
 .DESCRIPTION
-    THIS SCRIPT IS NOT RECOMMENDED TO BE USED AS IS, BUT SHOULD BE SEEN AS A COLLECTION OF STEPS TO BE RUN MANUALLY
-    TO GET EFLOW UP AND RUNNING.
-    This script automatically install Edge For Linux On Windows (EFLOW) on a new Windows Server VM
-    The script might trigger a restart and might need to be rerun after a restart
+    This script automatically installs Edge For Linux On Windows (EFLOW) on a new Windows Server VM.
+    This script assumes that the Windows VM has Hyper-V already installed.
+    The script accepts the following parameters:
```
danigian, line 6: Let's double check about internet connection sharing with the EFLOW team. If this is needed, we should add an item in the backlog to eventually improve this script for automating such connection sharing.

Tools/Eflow/provision-eflow.ps1
```diff
@@ -1,109 +1,85 @@
...
-
-if(Get-NetNat -Name "$switchName" -ErrorAction SilentlyContinue)
-{
-    throw "Net Nat with name $switchName already existing";
-}
+$startIp=$ipAddressFamily+$startEflowIpRange
 
 New-NetNat -Name "$switchName" -InternalIPInterfaceAddressPrefix "$natIp/24"
 
-#Install DHCP
```
danigian, line 86: We don't need DHCP anymore? Are we relying on the DHCP from the network to which the VM is connected?

Tools/Eflow/provision-eflow.ps1
noraabiakar: I tried to automate this, and it _kind of_ worked. 

The powershell script I used (See [here](https://superuser.com/questions/470319/how-to-enable-internet-connection-sharing-using-command-line))
```
regsvr32 hnetcfg.dll
$m = New-Object -ComObject HNetCfg.HNetShare
$c = $m.EnumEveryConnection |? { $m.NetConnectionProps.Invoke($_).Name -eq "Ethernet" }
$config = $m.INetSharingConfigurationForINetConnection.Invoke($c)
$config.EnableSharing(0)
```

After running this, if you check the network connections/properties/sharing check box, you will see that it is checked. However, the desired effect (a change in the IP address of the VM switch) doesn't happen. I could only make it happen when checking the box manually.  

I have to stop working on this for now, that's why I documented the manual approach. But I can open an issue describing what I tried for whoever tries to implement this next.

Tools/Eflow/provision-eflow.ps1
noraabiakar: I believe we never needed DHCP because we are assigning a static IP address to the EFLOW VM. @Mandur can you verify?

Tools/Eflow/install-hyperv.ps1
```diff
@@ -0,0 +1,15 @@
+<# .SYNOPSIS
+    Hyper-V installation Script
+.DESCRIPTION
+    This script automatically installs Hyper-V on a Windows Server VM to allow eflow provisioning.
+    The VM requires a restart after this script is run before eflow can be provisioned.
+.NOTES
+     Author     :
```
Mandur, line 7: You could add yourself here ;)

Tools/Eflow/provision-eflow.ps1
Mandur: By reading [this docs](https://learn.microsoft.com/en-us/azure/iot-edge/iot-edge-for-linux-on-windows-networking?view=iotedge-1.4#ip-address-allocations) I was under the impression it was not needed as we need to use static IP to have a static nat mapping. Do you have another opinion @danigian ?

Tools/Eflow/provision-eflow.ps1
Mandur: Here is the issue where I found the idea : https://github.com/Azure/iotedge-eflow/issues/59#issuecomment-893620909

danigian: I think the issue is related to the fact that we are creating an "Internal" virtual switch.
Can we try using an ["External" virtual switch](https://learn.microsoft.com/en-us/azure/iot-edge/iot-edge-for-linux-on-windows-networking?view=iotedge-1.4#virtual-switch-choices) instead? I think we did not use it at the beginning as it was not supported for EFLOW, but now seems to be supported for both windows server and client environments.

Tools/Eflow/provision-eflow.ps1
danigian: All good, makes sense to not use DHCP for this scenario

Tools/Eflow/install-hyperv.ps1
danigian, line 15: We should either restart the machine or ask the user to restart it here

noraabiakar: It's a step in the readme

Tools/Eflow/provision-eflow.ps1
Mandur: The [doc](https://learn.microsoft.com/en-us/azure/iot-edge/how-to-create-virtual-switch?view=iotedge-1.4#create-virtual-switch) seems to indicate you can't. But we will try anyway

noraabiakar: I've also tried it
image

Mandur (approved):
Very cool

Tools/Eflow/provision-eflow.ps1
danigian: See docs [here](https://learn.microsoft.com/en-us/powershell/module/hyper-v/new-vmswitch?view=windowsserver2022-ps#-switchtype) on how to create an external switch.

@Mandur, I just saw the note, but we could give it a try anyways

noraabiakar: I tried again. The switch is assigned an internal IP, and EFLOW does not have access to the internet. 
image

image

Tools/Eflow/install-hyperv.ps1
danigian: Yep, can we also print a log saying "Please reboot"? from the script? 👍

noraabiakar: The script prints out a log that indicates that a restart is needed. But I will add an extra message :)

Tools/Eflow/provision-eflow.ps1
danigian: Ok, let's keep as it is. Will maybe investigate in the future 👍

danigian (approved):

PR #1989, opened 2022-11-19
Fix ARM URL to point to release

Mandur:
The Docs QuickStart page mistakenly still pointed to the arm version and not to the generated bicep file. This pr aims to fix this.


Mandur:
Thank you for finding this out, it indeed doesn't work, I superseeded it by #2003

Issue #1984, opened 2022-11-18
Release pipeline: Azure Functions zip improperly created


danigian:
## Expected Behavior
The Azure Functions zip should have, in its root, the content of the folder specified [in this line](https://github.com/Azure/iotedge-lorawan-starterkit/blob/f82dcbd3d015ecd40b238e7dabc000257f495f2e/.github/workflows/create_release.yaml#L117) and not the full folder structure.

See comparison below
![image](https://user-images.githubusercontent.com/1955514/202679418-bbba5b25-f89b-4ad7-9f0c-df929a8ec994.png)

Azure Functions zip for release 2.2.1 is now manually fixed by me. Pipeline should be fixed before next release

Mandur:
Fixed by https://github.com/Azure/iotedge-lorawan-starterkit/pull/1992

PR #1992, opened 2022-11-19
Fix function ZIP in release pipeline

Mandur:
# PR for issue #1984

## What is being addressed

The function ZIP is now properly created with no nested file structure. Download https://github.com/Azure/iotedge-lorawan-starterkit/releases/download/untagged-7d59fa008be8b2864239/function-0.0.0.zip to check for what does the pipeline generate with this changes


.github/workflows/create_release.yaml
```diff
@@ -108,13 +108,13 @@ jobs:
...
       - uses: actions/setup-dotnet@v3.0.3
-        with: 
+        with:
           dotnet-version: ${{ env.DOTNET_VERSION }}
       - name: Package Azure Function
         run: |
           dotnet publish LoRaEngine/LoraKeysManagerFacade -c Release
           FUNCTION_ZIP=function-${VERSION}.zip
-          zip -r $FUNCTION_ZIP LoRaEngine/LoraKeysManagerFacade/bin/Release/net${DOTNET_VERSION%??}
+          cd LoRaEngine/LoraKeysManagerFacade/bin/Release/net${DOTNET_VERSION%??} && zip -r $FUNCTION_ZIP . && mv $FUNCTION_ZIP ../../../../..
```
danigian, line 117: Rather than moving the function zip, couldn't we just properly set the environment variable on row 119 instead? Something like: FUNCTION_ZIP=$(pwd)/$FUNCTION_ZIP could work?

danigian (approved):
Approving, with one suggestion

.github/workflows/create_release.yaml
Mandur: sure

Mandur: fixed in https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/3558870073

PR #2037, opened 2023-01-22
Update station-module-configuration.md

ABHITHLALC:
I would recommend a detailed documents about how to access lorawan cli and configure environment variables

# PR for issue #XXXX


Mandur:
Hello, 

Thank you for the suggestion, if you want to leave a suggestion, please open an issue rather than a Pull Request. 

What exactly are the section missing from the doc section  [here](https://azure.github.io/iotedge-lorawan-starterkit/2.2.1/tools/device-provisioning/#supported-commands) you would recommenD?

PR #2074, opened 2023-04-11
Fix the GetStationTwinAsync to be handle non existing twins

### LINKED ISSUE #1866: LNS Discovery Tests are broken
#1795 broke LNS Discovery Test by introducing a dependency on environment variables that are only available in the deployed web app for Facade Function.

# Acceptance criteria
- [ ] E2E should be independent from those function app related environment variables
- [ ] Tests should work again

Mandur:
Since the PR #1822, the result of the call to GetStationTwinAsync was improperly wrapped and the null case was improperly handled which caused the discovery test to fail ([example](https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/4596429155/jobs/8119000342#step:19:101)). This fix now resolves the issue ([GetStationTwinAsync](https://github.com/Azure/iotedge-lorawan-starterkit/actions/runs/4659558596/jobs/8246693837#step:19:2))

close #1866


Mandur:
I think that will also be closing #1866

danigian (approved):

PR #2102, opened 2023-10-17
Change provisioning to random

Mandur:
# Change provisioning to random Keys

The demo provisioning was using fixed keys for the demo devices, we want now to use random keys to avoid triggering scanning tools.

Documentation changes will be addressed in a subsequent PR


Template/azuredeploy.json
Mandur, line 1: I think we still need it

Mandur: I will readd in subsequent release pipeline change

noraabiakar (approved):

PR #2106, opened 2023-10-29
Replace secrets with random

Mandur:
## What is being addressed

Randomize unit tests to avoid triggering security scanners

### DIFF
--- a/Tools/Cli-LoRa-Device-Provisioning/Tests/LoRaWan.Tools.CLI.Tests.Unit/DeviceProvisioningTest.cs
+++ b/Tools/Cli-LoRa-Device-Provisioning/Tests/LoRaWan.Tools.CLI.Tests.Unit/DeviceProvisioningTest.cs
@@ -17,6 +17,8 @@ namespace LoRaWan.Tools.CLI.Tests.Unit
         private readonly Mock<RegistryManager> registryManager;
 
         private const string NetworkName = "quickstartnetwork";
+        private static readonly Random Random = new Random();
+
         private static readonly string DevEUI = GetRandomHexNumber(16);
         private const string Decoder = "DecoderValueSensor";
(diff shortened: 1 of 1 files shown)


Tools/Cli-LoRa-Device-Provisioning/Tests/LoRaWan.Tools.CLI.Tests.Unit/DeviceProvisioningTest.cs
```diff
@@ -17,20 +17,24 @@ public class DeviceProvisioningTest
         private readonly Mock<RegistryManager> registryManager;
 
         private const string NetworkName = "quickstartnetwork";
-        private const string DevEUI = "46AAC86800430028";
+        private readonly string devEUI = GetRandomHexNumber(16);
```
fbeltrao: Do we want different values per test, or same value for all tests?
If we want same value for all tests we could make this static (as `private static readonly`)

Tools/Cli-LoRa-Device-Provisioning/Tests/LoRaWan.Tools.CLI.Tests.Unit/DeviceProvisioningTest.cs
```diff
@@ -104,7 +108,7 @@ public async Task AddABPDevice(bool deviceExistsInRegistry)
             if (deviceExistsInRegistry)
             {
                 this.registryManager.Verify(c => c.UpdateTwinAsync(
-                        DevEUI.ToString(),
+                        devEUI.ToString(),
```
fbeltrao: nit: we don't need `.ToString()`, as it is already one.
There are other occurences with `devEUI` (check `appKey` and `appEui` for completeness).

fbeltrao (approved):

PR #2129, opened 2024-05-30
Fixes #2126 - Adds configuration of upstream protocol in cloud based mode

### LINKED ISSUE #2126: LNS only supports Amqp_Tcp_Only in Cloud-based mode
## Expected Behavior
I would like to be able to choose the communication protocol with the IoT Hub (e.g. Amqp_Tcp_Only, Amqp_Websocket_Only, Mqtt, etc.) in Cloud-Based mode, which was possible when using the Edge Hub.

## Current Behavior
The LNS currently only support Amqp_Tcp_Only (see [LoRaDeviceFactory.cs#L155](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs#L155)), which prevents us from using the same protocol as when using the Edge Hub. Amqp_Tcp_Only uses port 5671 (and 8883 ?) which is not allowed in our firewall.

Our current implementation uses Amqp_Websocket_Only (port 443), and due to strict firewall policies we can't use this mode in our infrastructure. It should be possible to choose the protocol of our choice, [as supported in the Edge Hub](https://github.com/Azure/iotedge/blob/main/doc/EnvironmentVariables.md#edgehub) module with the `UpstreamProtocol` environment variable.

## Steps to Reproduce

1. Close outbound ports 5671 and 8883 in firewall
2. Deploy LNS
3. Connect a gateway to the LNS

-> Connection will get stuck when trying to update the Device Twin

### Device (Host) Operating System

Ubuntu 20.04

### Architecture

amd64

### Container Operating System

Linux containers

### LoRaWAN Module Version
v2.2.2

#### Docker

26.1.1-1

## Additional Information

When ports are blocked, the communication gets stuck indefinitely in [LoRaDeviceClient.cs#L106](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceClient.cs#L106).

I forked the project and tried adding Amqp_Websocket_Only as a second protocol in transportSettings ([LoRaDeviceFactory.cs#L155](https://github.com/Azure/iotedge-lorawan-starterkit/blob/dev/LoRaEngine/modules/LoRaWanNetworkSrvModule/LoRaWan.NetworkServer/LoRaDeviceFactory.cs#L155) but it won't connect. It only works when only Amqp_Websocket_Only is present. But I may be missing something :)

Metal-Mighty:
# PR for issue #2126 

## What is being addressed

Adds the ability to configure the Upstream Protocol for AMQP communication between Azure Device Clients and Azure IoT Hub in cloud-based mode.

This setting was present in the edgeHub module when not in cloud-based mode and the LNS forced communication in AMQP TCP, which uses different port than AMQP WS and can break in restricted network environments.

## How is this addressed

- Adds an `UPSTREAM_PROTOCOL` environment variable and an `UpstreamProtocol` property in the LNS configuration
- Adds an extension method and enum with supported options (`Amqp` or `AmqpWs`)
- Initializes the LoRaDeviceClient with the selected protocol, with AMQP TCP (`Amqp`) the default value (same as before)


Metal-Mighty:
@microsoft-github-policy-service agree company="Michelin"

Mandur (approved):

Mandur:
We cannot run the CI on this branch, therefore it is not possible to merge it at the moment. I am currently in travel I will have a look when I come back

Tests/Unit/NetworkServer/DisposableExtensionsTests.cs
```diff
@@ -10,7 +10,7 @@ namespace LoRaWan.Tests.Unit.NetworkServer
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
-    using LoRaWan.NetworkServer;
+    using LoRaWan.NetworkServer.Extensions;
```
Mandur, line 13: I had to revert this change to have it compiled

Tests/Unit/NetworkServer/CancellationTokenExtensionsTest.cs
```diff
@@ -5,7 +5,7 @@ namespace LoRaWan.Tests.Unit.NetworkServer
 {
     using System;
     using System.Threading;
-    using LoRaWan.NetworkServer;
+    using LoRaWan.NetworkServer.Extensions;
```
Mandur, line 8: I had to revert this change to have it compiled

Mandur:
There is still a problem with our CI to accept the external change - however when running the PR locally in addition to the code changes I see some tests failing:

image

Metal-Mighty:
Hi @Mandur Thanks for the feedback. Weird about the revert, we have a working build on our environment. It's true I didn't check the unit tests though ! 
I'll work on it asap and keep you updated

